PR #3, opened 2020-12-14
Lightning redux

Jasonvdb:
- Lightning redux store
- Custom react hook for starting LND and keeping state/info up to date
- Basic wallet list item for LND wallet (will change when to just list all available wallets that inherit from `IDefaultWalletShape` when that's ready)


coreyphillips (approved):
LGTM

PR #4, opened 2020-12-15
Setup & Create Wallet

coreyphillips:
This sets up, creates and stores a wallet on the device when the app is first launched.


src/App.tsx
```diff
@@ -1,16 +1,39 @@
...
 import useLightning from "./utils/hooks/lightning";
 import themes from "./styles/themes";
+import { getStore } from "./store/helpers";
+import { createWallet } from "./store/actions/wallet";
+import { start as startElectrum } from "rn-electrum-client/helpers";
+
+const startApp = async (): Promise<void> => {
+	try {
+		InteractionManager.runAfterInteractions(async () => {
+			//Create wallet if none exists.
```
Jasonvdb, line 18: Should I move the `startLnd` to here as well?

src/store/reducers/wallet.ts
```diff
@@ -4,7 +4,7 @@ import { IWallet } from "../types/wallet";
 const wallet = (state: IWallet = {
 	loading: false,
 	error: false,
-	selectedNetwork: "bitcoin",
+	selectedNetwork: __DEV__ ? "bitcoinTestnet" : "bitcoin",
```
Jasonvdb, line 7: If you build and upload an APK release is `__DEV__` still true? We may want to use an env var or something for when we push to testflight or google play alpha and we still want the app to use testnet. 

I also need to configure LND to use this network as well.

Jasonvdb (approved):

PR #9, opened 2020-12-22
Invoice state change subscriptions, toast notifications

Jasonvdb:
- New `Result.ts` type utility. I changed all the code I wrote to use it but didn't want to end conflicts your side so let me know if it's your stuff is all committed for now and I'll update the rest.
- In app toast notifications. Used currently when invoices are paid, channels opening, etc. Expecting to customise or remove later for our own designs.
- Invoice list added to redux store and refreshed when there's any changes to an invoice. We can optimise later for using pagination if required, just figured that's a bit premature for now.

Simulator Screen Shot - iPhone 11 - 2020-12-22 at 12 18 37

Simulator Screen Shot - iPhone 11 - 2020-12-22 at 12 18 53


coreyphillips (approved):
Nice! LGTM

I'll begin migrating my old code to the new `Result.ts` type utility as needed.

PR #16, opened 2020-12-30
Paste from clipboard button on QR scanner

Jasonvdb:
- Button on the QR scanner that allows the user to paste from clipboard if they have any addresses or payment requests copied. Useful for simulator testing. 
- Showing some toast notifications at the bottom of the screen else you can't see them with the iOS 14 clipboard warnings.

 I don't think we should ever automatically check their clipboard on iOS as those iOS 14 clipboard warnings really freak people out when they didn't initiate the check. We could probably work in a different experience on Android though.

### DIFF
--- a/src/screens/Scanner/index.tsx
+++ b/src/screens/Scanner/index.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 import { View } from '../../styles/components';
 import { Alert, StyleSheet } from 'react-native';
+import Clipboard from '@react-native-community/clipboard';
 import Camera from '../../components/Camera';
 import {
 	showErrorNotification,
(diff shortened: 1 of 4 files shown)


src/screens/Scanner/index.tsx
```diff
@@ -11,27 +12,34 @@ import { decodeQRData, EQRDataType, QRData } from '../../utils/scanner';
...
+	const handleData = async (data: QRData): Promise<void> => {
 		if (data.network !== selectedNetwork) {
-			return showErrorNotification({
-				title: 'Unsupported network',
-				message: `App is currently set to ${selectedNetwork} but QR is for ${data.network}.`,
-			});
+			return showErrorNotification(
+				{
+					title: 'Unsupported network',
+					message: `App is currently set to ${selectedNetwork} but QR is for ${data.network}.`,
```
coreyphillips, line 27: For onchain payments, if we scan a testnet address, but it's currently set to mainnet, we'll just switch to testnet and do our best to handle the payment from our onchain testnet wallet. For lightning payments/invoices, we may need to store an additional data point in `shapes/lightning.ts` to determine which network it's currently using since we won't be running both a mainnet and testnet instance of lnd. It will be one or the other.

coreyphillips (approved):

PR #20, opened 2021-01-07
On chain transaction activity list

Jasonvdb:
Updates:
- Wallet transactions added to activity list store
- Updated tests
- Pull to refresh on activity screen

Questions: 
- Updating the activity transactions from `src/store/actions/wallet.ts`. Is that the most appropriate place?
- For the activity item I'm setting `confirmed` as `true` only when block height isn't zero. That correct? Or should we add number of confs to this as well and use that rather?
- We don't have a description field for on chain transactions. Will we later just allow users to label addresses and then just use that lookup for the description in the activity list?
- How do I get the timestamp of a tx?


coreyphillips:
- We should probably keep all activity logic separate from lightning/wallet. There are a few things we can do instead. One is to listen for `UPDATE_TRANSACTIONS` in the activity reducer and filter/update the activity list accordingly from there. Another is to create a generic `updateActivityList` method in `actions/activity` that will pull data from the lightning/wallet store and format/update the activity list as needed. The `updateActivityList` method could be run at the tail end of any wallet refresh method.
- Anything <= 0 indicates an unconfirmed transaction. The height is coming from get_history [here](https://electrum.readthedocs.io/en/latest/protocol.html?highlight=methods#blockchain-address-get-history).
- It hasn't been implemented yet, but an onchain transaction may have both a description and a message. The description can be added by the user to any given transaction as a type of memo and may also be provided as part of an invoice when paying a transaction and added automatically. The second is a message that could be any data embedded in an OP_RETURN.
- The timestamp is coming from blocktime in ITransaction. Just made a commit [here](https://github.com/synonymdev/spectrum/pull/20/commits/b5eacdb9fac4094bf7c8bfc4b4f0f03f035dfbfb) to add that to the stored transaction.

Jasonvdb:
Ok good to go I think. Originally I was trying to avoid storing transactions in the activity store as well as in the individual wallet stores. Was also trying to just update individual items instead of whole lists. But that's all probably premature optimisation. It should be simpler this way now.

Not sure how I could get it to listen for updates but if you want to add that please go ahead. I've just created a `updateLightingActivityList` and `updateOnChainActivityList` that update the activity items from the wallet stores after  the transactions are refreshed.


Simulator Screen Shot - iPhone 11 - 2021-01-08 at 18 58 58

coreyphillips (approved):

PR #23, opened 2021-01-11
Added transaction to wallet.wallets.

coreyphillips:
- Added UPDATE_ON_CHAIN_TRANSACTION & RESET_ON_CHAIN_TRANSACTION actions.
 - Added onChainTransaction to wallet shapes.
 - Added IOnChainTransactionData, EOnChainTransactionData & IOnChainTransaction to wallet types.
 - Updated children type in AssetCard.
 - Wrapped BitcoinCard in Pressable and added logic to display/hide buttons when tapped.


Jasonvdb (approved):
Not now but maybe later we should consider refactoring vars that have units to include he unit in the name. Like `feeSats` or `transactionSizeBytes`. Would probably make things simpler for any future devs reading the code.

PR #25, opened 2021-01-15
On-Chain Send Functionality

coreyphillips:
This PR adds send functionality for on-chain transactions to the app. 
Added functionality includes:
 - Scan/Verify addresses via QRCode.
 - Send form now appears when pressing "Send" from the on-chain wallet.
 - Able to include OP_RETURN in message field when sending a transaction.
 - Able to successfully create and broadcast a transaction to a specific address, with a specific amount and fee.
 - The app should successfully update the balance once a transaction is complete.


Jasonvdb (approved):
nice!

PR #29, opened 2021-01-20
GitHub actions linting and tests setup

Jasonvdb:
- GH action lint check (add inline lint warnings to PRs)
- GH action testing

`build-rn-lightning.sh` is a temporary solution for including react-native-lightning from a private repo and getting it to work. Will remove once the package can be published to npm.


coreyphillips:
It doesn't appear to create the `dist` directory when installing. Resulting in the following error on my end:

Screen Shot 2021-01-20 at 1 34 20 PM

Are you able to reproduce this as well or am I doing something wrong?

coreyphillips:
Ahh, I see. Please ignore the previous comment.
How do you feel about adding and calling the following function in `postinstall.js`?

```
function buildRnLightning() {
	exec(
		'echo "***Start of temporary hack for building react-native-lightning. TODO: remove this once published to npm.***"\n' +
		'cd node_modules/react-native-lightning/\n' +
		'rm -rf dist\n' +
		'mkdir dist\n' +
		'yarn protobuf\n' +
		'cd ../../\n' +
		'cp node_modules/react-native-lightning/dist/rpc.js node_modules/react-native-lightning/src/lightning\n' +
		'sed -i -e "s/dist\\/index.js/src\\/lightning\\/index.ts/g" node_modules/react-native-lightning/package.json\n' +
		'echo "***End of temporary hack.***"'
	)
}
```

Then we could remove  `build-rn-lightning.sh` and not worry about calling the script for new installs.

Jasonvdb:
Reason I put it in its own bash script was because I thought maybe it only needed to be called from the github action. But maybe I can tweak it so it fixes things for local dev as well. Then I'll put it in the postinstall rather. Will try update quick.

Jasonvdb:
Cool updated and added to `postinstall.js`. Fixes local development env as well now.

coreyphillips (approved):

PR #32, opened 2021-01-22
Unit test fixes

Jasonvdb:
- Test coverage command
- Tests running on MacOS for consistency. Ubuntu was causing some issues with mock dependencies.

Once the tests were actually running properly it picked up some issues I also fixed in this PR. 

The first one was parsing a on chain payment request. 

```
Expected: 50000
Received: 0.0005
```

I made `QRData` use sats but `parseOnChainPaymentRequest` returned the bitcoin amount. I changed it to use sats as the base unit. I figured because LND is using sats consistently we should probably do the same so there's never any confusion on the unit being used under hood. And then right before it's displayed in the UI we convert it to whatever. 

The second failed test was to do with the message/label not getting the expected result from `parseOnChainPaymentRequest`. Originally I added both `message` and `label` to `QRData` because of BIP21. But seems like all apps only use the message field so I guess we can just stick with `message` everywhere for now?

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,3 +6,4 @@ android/*
 metro.config.js
 postinstall.js
 __mocks__/**
+coverage
--- a/.github/workflows/jest.yml
+++ b/.github/workflows/jest.yml
@@ -5,7 +5,7 @@ on: pull_request
 jobs:
   tests:
     name: Run jest unit tests
-    runs-on: ubuntu-latest
+    runs-on: macos-latest
 
     steps:
       - name: Check out Git repository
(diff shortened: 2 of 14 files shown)


coreyphillips:
LGTM. I updated the sats value to return no decimals and passed the message data to label when updating info for the transaction. It's not implemented yet, but labels were intended to be locally saved descriptors of a transaction and `messages` are OP_RETURN data found in a given transaction. I may need to eventually change `messages` to `opReturn` or something similar for clarity.

coreyphillips (approved):

coreyphillips (approved):

Jasonvdb:
Ok nice catch thanks! Yeah probably a new field for `opReturn` would be clearer then. I tested a few apps and they all consistently using the `message` field in a payment request.

PR #41, opened 2021-03-04
Neutrino cache download

Jasonvdb:
- Downloads and unzips a cache of the Neutrino DB and headers.
- Will only attempt a download if the `lnd` directory does not already exist.
- LND will be started to sync manually as a fallback if the download were to fail.

Working on a way to automate the hosted cache so it's updated often enough but that will come in another PR.


src/utils/lightning/cachedHeaders.ts
```diff
@@ -0,0 +1,126 @@
...
+import RNFS from 'react-native-fs';
+import { unzip, subscribe } from 'react-native-zip-archive';
+import { err, ok, Result } from '../result';
+import { updateCachedNeutrinoDownloadState } from '../../store/actions/lightning';
+import { getStore } from '../../store/helpers';
+
+export const downloadNeutrinoCache = async (
+	network: LndNetworks,
+): Promise<Result<boolean>> => {
+	const url = 'https://github.com/Jasonvdb/lnd-ios/releases/download/1/';
```
coreyphillips, line 12: We'll eventually want to move this to https://github.com/synonymdev/spectrum

coreyphillips (approved):

src/utils/lightning/cachedHeaders.ts
Jasonvdb: Yeah just easier to test this with a public repo. Also once the cache updates are automated this link will get updated each time I think.

PR #44, opened 2021-03-19
Derivation Path Upgrade

coreyphillips:
This PR makes the key derivation path much more granular. Allowing us to tweak each path as needed moving forward. The new format is as follows:

```
interface IKeyDerivationPath {
	purpose: '84' | '49' | '44'; //"bech32" | "segwit" | "legacy";
	coinType: '0' | '1'; //"mainnet" | "testnet";
	account: '0' | '2' | '3'; //"On-Chain Wallet" | "RGB" | "Omnibolt";
	change: '0' | '1'; //"Receiving Address" | "Change Address";
	addressIndex: string; //'0' | '1' | '2' | '3', etc.
}
```

### DIFF
--- a/src/utils/backup/protos/scheme.proto
+++ b/src/utils/backup/protos/scheme.proto
@@ -14,7 +14,17 @@ message Wallet {
   string key = 1; //e.g. wallet0, wallet1
   string mnemonic = 2; //24 word seed phrase
   string passphrase = 3; //seed passphrase
-  string keyDerivationPath = 4; //Usually 84
+
+  message KeyDerivationPath {
+    string purpose = 1;
+    string coinType = 2;
+    string account = 3;
+    string change = 4;
+    string addressIndex = 5;
+  }
+
+  KeyDerivationPath keyDerivationPath = 4;
+
 
   enum AddressType {
     bech32 = 0;
--- a/src/store/actions/wallet.ts
+++ b/src/store/actions/wallet.ts
@@ -1,6 +1,5 @@
 import actions from './actions';
 import {
-	EWallet,
 	ICreateWallet,
 	IFormattedTransaction,
 	IOnChainTransactionData,
@@ -10,7 +9,6 @@ import {
 import {
 	createDefaultWallet,
 	formatTransactions,
-	generateAddresses,
 	getAddressHistory,
 	getCurrentWallet,
 	getExchangeRate,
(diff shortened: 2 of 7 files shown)


src/utils/backup/protos/scheme.proto
```diff
@@ -14,7 +14,17 @@ message Wallet {
...
   string mnemonic = 2; //24 word seed phrase
   string passphrase = 3; //seed passphrase
-  string keyDerivationPath = 4; //Usually 84
+
+  message KeyDerivationPath {
+    string purpose = 1;
+    string coinType = 2;
+    string account = 3;
+    string change = 4;
+    string addressIndex = 5;
```
Jasonvdb, line 23: Is this index for receiving or a change address? When importing a wallet would spectrum not just keep checking addresses until it found an unused one?

coreyphillips: > Is this index for receiving or a change address?

It's a receiving-address when `change` is set to 0 and a change-address when `change` is set to 1. 

So using this format:
`/ m / purpose' / coin_type' / account' / change / address_index`

This represents a bech32  onchain receiving-address at index 0:
`/ m / 84' / 0' / 0' / 0 / 0`

This represents a bech32  onchain change-address at index 0:
`/ m / 84' / 0' / 0' / 1 / 0`

Each new address/change-address that we need to generate will bump the respective `address_index` up by one.
So for an onchain bech32 receiving-address at index 1 we would use `/ m / 84' / 0' / 0' / 0 / 1`

<br />

> When importing a wallet would spectrum not just keep checking addresses until it found an unused one?

It will keep checking addresses up to a specified address gap limit. The standard gap limit is 20.
[Reference here](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit). [Article here](https://blog.lopp.net/mind-the-bitcoin-address-gap/#:~:text=Address%20gap%20limit%20is%20currently,stops%20searching%20the%20address%20chain.).
So if there are 20 unused/tx-free addresses the app will stop searching and assume there are no Bitcoin to be had at higher indexes. Currently the gap limit in Spectrum is set to 1 for development purposes and will need to be updated later to accommodate the standard 20 or even allow users to customize it in the settings.

Jasonvdb (approved):

PR #63, opened 2021-04-23
Activity sliding sheet

Jasonvdb:
- Sliding sheet with activity list
- Search filter
- Updated activity list style although didn't get too stuck in the details yet
- Moved the scanner button to the tab for now

Still working on the TX details view and value formatting. Will push that in a separate PR.

Simulator Screen Shot - iPhone 11 - 2021-04-23 at 16 17 37

Simulator Screen Shot - iPhone 11 - 2021-04-23 at 16 17 24


coreyphillips:
react-native-reanimated v2+ is still fairly new, has its quirks and is not fully compatible with other libraries yet. I wasn't able to get the swipe behavior to work as expected with v2, but downgrading to v1 has it working as intended on both iOS & Android. Please feel free to look over my changes and ensure everything checks out. Otherwise, LGTM!

Jasonvdb:
Nice, thanks! Just did a quick layout fix on that search input for android.

PR #65, opened 2021-05-07
PSBT support for LND channel funding

Jasonvdb:
- Unit tests for PSBT steps
- Unit tests for on chain transaction creation
- Refactored `createTransaction` into reusable chunks that can be used for the LND funding steps
- Skip shuffling targets when executed from jest so tests are reproducible

Originally my attempt at this I was trying to keep all PSBT creation separate from the transaction creation but it became almost the same which is why I opted for the refactoring instead. Before refactoring I added a basic test for tx creation to make sure I wasn't breaking anything but we should probably make a more elaborate test for all the different tx creation scenarios in the future.

I'm working on some UI that allows the user to move funds on-chain to off-chain but will submit that in another PR as it's not 100% yet.


src/utils/wallet/transactions.ts
```diff
@@ -306,9 +306,263 @@ export interface ICreateTransaction {
...
+		outputs,
+	});
+
+	const { currentWallet } = getCurrentWallet({
+		selectedNetwork,
+		selectedWallet,
+	});
+
+	const network = networks[selectedNetwork];
+	//TODO: Get address type by inspecting the address or path.
```
coreyphillips, line 397: We may want to consider adding a library such as [bitcoin-address-validation](https://github.com/ruigomeseu/bitcoin-address-validation) for this purpose.

coreyphillips (approved):

PR #68, opened 2021-05-18
Lightning PSBT Funding Fee Update

coreyphillips:
This PR adds a way to calculate and adjust the fee when funding a lightning channel via psbt.
More specifically, `fundingLightning` was added as a param to the `getTotalFee` method. Setting `fundingLightning` to true will add an additional bech32 output into its calculation to account for the future output address lnd will return.

psbt-fee-update

### DIFF
--- a/src/screens/Wallets/SendOnChainTransaction/FeeSummary.tsx
+++ b/src/screens/Wallets/SendOnChainTransaction/FeeSummary.tsx
@@ -1,0 +1,119 @@
+import { defaultOnChainTransactionData } from '../../../store/types/wallet';
+import React, { memo, ReactElement, useCallback } from 'react';
+import Summary from './Summary';
+import { LayoutAnimation, StyleSheet } from 'react-native';
+import { getTransactionOutputValue } from '../../../utils/wallet/transactions';
+import { useSelector } from 'react-redux';
+import Store from '../../../store/types';
+import { getFiatBalance } from '../../../utils/helpers';
+import { View } from '../../../styles/components';
+
+const FeeSummary = ({
+	amount: _amount = '0',
+	lightning = false,
+}: {
+	amount?: string | number;
+	lightning?: boolean;
+}): ReactElement => {
+	const selectedWallet = useSelector(
+		(store: Store) => store.wallet.selectedWallet,
+	);
+	const selectedNetwork = useSelector(
+		(store: Store) => store.wallet.selectedNetwork,
+	);
+
+	const transaction = useSelector(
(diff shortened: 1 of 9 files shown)


src/screens/Wallets/SendOnChainTransaction/FeeSummary.tsx
```diff
@@ -0,0 +1,119 @@
...
+	lightning?: boolean;
+}): ReactElement => {
+	const selectedWallet = useSelector(
+		(store: Store) => store.wallet.selectedWallet,
+	);
+	const selectedNetwork = useSelector(
+		(store: Store) => store.wallet.selectedNetwork,
+	);
+
+	const transaction = useSelector(
```
Jasonvdb: I made[ these hooks](https://github.com/synonymdev/spectrum/blob/490b9344d6addfae28d3d73df16094e4731ef35d/src/screens/Wallets/SendOnChainTransaction/TransactionHook.ts#L11) for this type of repeated logic if you want to use it

src/utils/wallet/transactions.ts
```diff
@@ -271,30 +307,37 @@ export const getTotalFee = ({
...
+			outputAddresses.push(changeAddress);
+		}
+
+		//Determine the address type of each address and construct the object for fee calculation
+		const inputParam = constructByteCountParam(utxoAddresses);
+		// @ts-ignore
+		const outputParam = constructByteCountParam(outputAddresses);
+		//Increase P2WPKH output address by one for lightning funding calculation.
+		if (fundingLightning) {
+			outputParam.P2WPKH = (outputParam?.P2WPKH || 0) + 1;
```
Jasonvdb, line 334: For lightning isn't is `MULTISIG-P2SH:2-2`? Or doesn't it make a difference to the byte count?

src/screens/Wallets/SendOnChainTransaction/FeeSummary.tsx
coreyphillips: Nice. Added that [here](4be6412554a829fc2a89c9f75b6ae7d25879bf54).

src/utils/wallet/transactions.ts
coreyphillips: `MULTISIG` is only applied when used as an input. When funding a lightning channel we treat it as a normal bech32 output.

Jasonvdb: Ah right I see 👍

Jasonvdb (approved):

PR #79, opened 2021-06-22
omnibolt-js update

coreyphillips:
This PR updates Backpack to the latest instance of omnibolt-js [here](https://github.com/synonymdev/omnibolt-js/pull/11), allowing Backpack to offload a majority of the Omnibolt signing and checkpoint logic. This PR also adds Mempool as an option in the `getBlockExplorerLink` method.


Jasonvdb (approved):
nice, much cleaner

PR #87, opened 2021-07-14
Implement Auto Coinselect

coreyphillips:
This PR implements auto-coinselect functionality into the SendForm component for on-chain transactions.


Jasonvdb (approved):
`runCoinSelect` should probably be something used when opening a channel as well. Not sure if we want to expose that option there.

PR #96, opened 2021-08-23
Electrum Config

coreyphillips:
This PR adds an "Electrum Config" option to the Settings menu allowing users to connect to an Electrum server of their choosing. This page also allows users to select a random peer acquired from their currently connected peer or the hardcoded list of peers if they're not currently connected.

Simulator Screen Shot - iPhone 12 - 2021-08-22 at 21 19 18


src/screens/Settings/ElectrumConfig/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+];
+
+const validateInput = ({
+	host = '',
+	port = '',
+}: {
+	host: string;
+	port: string | number;
+}): Result<string> => {
+	try {
```
Jasonvdb: Need this try/catch? Don't think anything in throws an error

src/screens/Settings/ElectrumConfig/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+				host: peerData.host.trim(),
+				protocol: peerData.protocol.trim(),
+				[protocol]: peerData.port.toString().trim(),
+			};
+			const connectResponse = await connectToElectrum({
+				selectedNetwork,
+				customPeers: [connectData],
+			});
+			setLoading(false);
+			if (connectResponse.isOk()) {
```
Jasonvdb, line 183: If the connection fails should we not show an error?

Jasonvdb (approved):

src/screens/Settings/ElectrumConfig/index.tsx
coreyphillips: An error notification was previously called in the connectToElectrum method when unable to connect. I pulled that out and now present a notification/message as needed given its response.

src/screens/Settings/ElectrumConfig/index.tsx
coreyphillips: Agreed, removed.

PR #97, opened 2021-08-24
UTXO List

coreyphillips:
This PR adds a utxo list to the send form when creating on-chain transactions for manual coin control. This effectively allows users to add and remove available utxo's for any transaction. 

Tapping on the left column will forward the user to their block explorer of choice to display the relevant transaction. Tapping the switch/right column will add it to or remove it from the current transaction. 

Simulator Screen Shot - iPhone 12 - 2021-08-23 at 21 50 15


coreyphillips:
This is a pretty rough implementation of coin control and will need some design guidance down the road. Calling it "UTXO List" or "Coin Control" may not be ideal and it's difficult to decide what relevant information should be displayed in the list itself (ex: hash, timestamp, value, etc.).

One thing to note is that the current implementation still performs the coin select operation even if the user has selected specific utxo's for a given transaction. We may want to discuss and consider changing this behavior to only perform the coin select operation when the user has not removed utxo's from their transaction.

Jasonvdb (approved):

PR #101, opened 2021-08-31
Balance Header

coreyphillips:
This PR adds the fiat balance header to the main view.
fiat-balance-header

### DIFF
--- a/src/screens/Wallets/SendOnChainTransaction/WalletHook.ts
+++ b/src/screens/Wallets/SendOnChainTransaction/WalletHook.ts
@@ -1,0 +1,19 @@
+import { useSelector } from 'react-redux';
+import Store from '../../../store/types';
+
+/**
+ * Retrieves the total wallet balance for the currently selected wallet and network.
+ */
+export function useBalance(): number {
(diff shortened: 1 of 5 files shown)


src/screens/Wallets/SendOnChainTransaction/WalletHook.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export function useBalance(): number {
+	const selectedWallet = useSelector(
+		(store: Store) => store.wallet.selectedWallet,
+	);
+	const selectedNetwork = useSelector(
+		(store: Store) => store.wallet.selectedNetwork,
+	);
+
+	return useSelector(
```
Jasonvdb, line 15: Should this not include LND's balance also?

Jasonvdb (approved):

src/screens/Wallets/SendOnChainTransaction/WalletHook.ts
Jasonvdb: Will add this in so I can use reuse for the asset detail view 👍

PR #124, opened 2021-10-27
Single Seed Derivation

coreyphillips:
This PR sets the app up to derive all seeds used for lightning, on-chain & omnibolt wallets from a single seed.

The current derivation is as follows:
Lightning Seed -> On-Chain & Omnibolt Seeds -> Omnibolt Login

The lightning seed is initially derived via `CipherSeed.random()` from the `aezeed` library when creating the wallet.

The on-chain/omnibolt seeds are derived by getting the sha256 hash of the lightning seed and passing it to `bip39.entropyToMnemonic(hash);`.

The omnibolt login phrase is derived by getting the sha256 hash of the on-chain seed and passing it to `bip39.entropyToMnemonic(hash);`.

### DIFF
--- a/src/store/actions/lightning.ts
+++ b/src/store/actions/lightning.ts
@@ -7,14 +7,18 @@ import lnd, {
 	ICachedNeutrinoDBDownloadState,
 	ss_lnrpc,
 } from '@synonymdev/react-native-lightning';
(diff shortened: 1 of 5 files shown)


Jasonvdb (approved):
Looks good to me

PR #138, opened 2022-03-15
Add react-native-mmkv

coreyphillips:
This PR:
 - Replaces AsyncStorage with MMKV for redux-persist storage.


src/store/mmkv-storage.ts
```diff
@@ -0,0 +1,23 @@
+import { MMKV } from 'react-native-mmkv';
+import { Storage } from 'redux-persist';
+
+const storage = process?.env?.JEST_WORKER_ID
+	? require('../../__mocks__/react-native-mmkv/react-native-mmkv')
```
Jasonvdb: Is this needed? Jest automatically knows to use the mocks directory when running tests as long as the mock is in the same format as the actual dependency

coreyphillips: I was unable to get tests to run, otherwise. Might need additional tweaking.

Jasonvdb: https://github.com/synonymdev/bitkit/pull/138/commits/0ce5b95d85f604fd4831426e40875893f6e57d66

Jasonvdb (approved):

PR #146, opened 2022-04-23
react-native-lightning Removal

coreyphillips:
This PR:
 - Removes all instances of react-native-lightning from the app.
 - Separates types from exchange-rate/index.ts.
 - Deletes LightningCard.tsx, lnd.conf, LndLogs.tsx.
 - Temporarily removes activity, fiat & backup tests.
 - Deletes react-native-lightning mock.
 - Adds protobufjs dependency to package.json.
 - Updates yarn.lock & Podfile.lock accordingly.


coreyphillips:
I tried my best to be surgical in this PR.
I left or commented out a few of the old Lightning components and methods in the event we can re-use some of the previous logic and flow when implementing LDK. High chance they too will be removed, but it makes for a good reference when implementing nonetheless.
Once some additional omnibolt logic and components are removed and LDK is in place we'll want to restore the tests removed in this PR.

package.json
```diff
@@ -48,10 +47,11 @@
     "omni-tracker-api": "git+ssh://git@github.com/synonymdev/omni-tracker-api#7ede407d825a339451a88ad0ae489a91a4a6507b",
     "omnibolt-js": "git+ssh://git@github.com/synonymdev/omnibolt-js#e9232089bdfb1394bf7f6c51f1831eb68813e008",
     "process": "^0.11.10",
+    "protobufjs": "^6.11.2",
     "radio-buttons-react-native": "^1.0.4",
     "react": "^17.0.2",
     "react-i18next": "^11.14.3",
-    "react-native": "^0.67.3",
+    "react-native": "0.66.4",
```
limpbrains, line 55: It's already 0.67.3 in master

limpbrains (changes requested):

__tests__/transactions.ts
```diff
@@ -60,61 +55,4 @@ describe('On chain transactions', () => {
...
-			expect(fundedPsbtRes.error.message).toBeUndefined();
-			return;
-		}
-
-		expect(fundedPsbtRes.value.toBase64()).toEqual(
-			'cHNidP8BAKYCAAAAAgwOqzFJuj7Xq9j0yY6r4suyt8NZBAS2bKDwGt32HsZxAAAAAAAAAAAAUb2EiFHK23G/NObg5GsMQhTC0GzMHVyg9brv3PhiaSAAAAAAAAAAAAACQOIBAAAAAAAiACBUjq4HEBbKBcpE8A9aXYBCI9DvdjT6HQN4m8ZDom+yGI4RAAAAAAAAFgAUZpqTI0GGk7gdRMGdp7H+eRGyFCkAAAAAAAEBH0DiAQAAAAAAFgAUm5iywlEvl4nZkQqRtDFLIUs1mDIAAQEfjhIAAAAAAAAWABQcXF2QWgJ2ysNHP1M+3dw0nAWaQwAAAA==',
-		);
-
-		//Sign the PSBT for LND finalization
-		const signedPsbtRes = await signPsbt({
```
Jasonvdb, line 105: Can we not keep the above code? Up until LND signs. Because we'll use this test with LDK psbt channel funding. If it breaks things then also fine, we can add it back when it's time.

Jasonvdb (approved):

__tests__/transactions.ts
coreyphillips: `lnrpc.ReadyForPsbtFunding.create` was a removed dependency breaking the test so I pulled it. We'll have to reincorporate the psbt funding logic once ldk is added.

Jasonvdb: Ah right ok

package.json
coreyphillips: - This version number was improperly upgraded to 0.67 from 0.66 and needs a proper upgrade in the future. 
- v0.67 also causes the ios build to fail after recent xcode/Mac OS updates after the introduction of skia in the project.

limpbrains (approved):

PR #147, opened 2022-04-29
new styles for Activity, Button, SeedInput, Slideshow

limpbrains:
- new button components. It now have new properties: size=small/large variant=primary/secondary. Also I removed margins from it, it could make some forms look ugly
- new ActivityDetail screen. For now I removed all tx bumping logic.
- new seed recovery with error reporting
- new icons, new styles, new colors
- new `useColors()` hook

I added `color: undefined` to all icon components, cause without it I can't use `<LightningIcon color="brand" />`, it throws that such color doesn't exist.


src/navigation/tabs/TabNavigator.tsx
```diff
@@ -75,39 +76,73 @@ const WalletsStack = (): ReactElement => {
...
+					snapPoint: 1,
+				},
+			});
+		}
+	}, [screen, params]);
+
 	const openScanner = useCallback(
-		() => navigation.navigate('Scanner'),
+		// () => navigation.navigate('Scanner'),
+		() => navigation.navigate('BitcoinToLightning'),
```
coreyphillips: Was `() => navigation.navigate('Scanner')` accidentally left commented out or is it meant to navigate to `BitcoinLightning`?

limpbrains: Yes, fixed

coreyphillips (approved):

Jasonvdb (approved):

PR #150, opened 2022-05-05
Slashtags integration

Jasonvdb:
- Added @synonymdev/react-native-slashtags 
- Updated to @react-native-clipboard/clipboard
- Add/update slashtags profiles
- Slashtags redux store
- React hook for keeping SDK in sync with redux store

https://user-images.githubusercontent.com/5300488/166933967-22884c09-7310-4f8d-a99a-bb42a2dfbe74.mov


src/store/actions/slashtags.ts
```diff
@@ -0,0 +1,50 @@
...
+	});
+};
+
+export const resetSlashtagsStore = (): Result<string> => {
+	dispatch({
+		type: actions.RESET_SLASHTAGS_STORE,
+	});
+
+	return ok('');
+};
```
limpbrains, line 50: `resetSlashtagsStore` is not an async function, but you call it later with `await`

src/hooks/slashtags.tsx
```diff
@@ -0,0 +1,120 @@
...
+			if (!state.sdkSetup) {
+				return;
+			}
+
+			if (state.profiles === Object.keys(profiles).length) {
+				return;
+			}
+
+			const profileNames = Object.keys(profiles);
+			for (let index = 0; index < profileNames.length; index++) {
```
limpbrains: You can use `for ... in` to iterate over Object keys

src/hooks/slashtags.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+		try {
+			const state = await slashtags.current.state();
+			if (!state.sdkSetup) {
+				return;
+			}
+
+			if (state.profiles === Object.keys(profiles).length) {
+				return;
+			}
```
limpbrains, line 35: Is it ok, that you can alter profile with `updateProfile`, but it will not be synced, because number of profiles has not been changed?

limpbrains (changes requested):

limpbrains:
Just a couple of questions:
- maybe it would be nicer, if `SlashtagsContext` would provide `slashtags` without `.current` prefix?
- as I understand, app waits while slashtags webview is ready. Maybe, while it is not, you can delay app render to prevent user from interacting with slashtags? Then you won't need in all those `api not ready` checks

src/hooks/slashtags.tsx
Jasonvdb: Yes this is fine, it does not need to be triggered again. The profile just needs to be updated with it's new slashtag after it's added to the SDK.

Jasonvdb:
> * maybe it would be nicer, if `SlashtagsContext` would provide `slashtags` without `.current` prefix?

I think I did try this but it caused the context to not automatically be updated. That `slashtags` value is actually a `useRef()` and it seemed to only work properly when the whole ref is passed to the provider [here](https://github.com/synonymdev/react-native-slashtags/blob/1bf00fb837960ca77b5de4726545b46eee52309d/lib/src/index.tsx#L43). 

>* as I understand, app waits while slashtags webview is ready. Maybe, while it is not, you can delay app render to prevent user from interacting with slashtags? Then you won't need in all those api not ready checks

I don't think it's necessary to add in an artificial delay every time the user opens the app. Nothing is remotely loaded, all slashtags web code is bundled with the app so it shouldn't take long to be ready. Especially when app is not in debug mode. The chances of someone scanning a QR or adding a new profile before it loads is unlikely but if it does happen we can look at this.

src/hooks/slashtags.tsx
Jasonvdb: Updated

src/store/actions/slashtags.ts
Jasonvdb: Fixed. Same thing with all the reset functions, not sure why that happened.

src/hooks/slashtags.tsx
limpbrains: You can do even better `for (const i in profiles) {`

limpbrains (approved):

src/hooks/slashtags.tsx
Jasonvdb: I still need the `name` (key) inside the loop though?

limpbrains: Sorry, missed that. It is fine

coreyphillips (approved):

PR #151, opened 2022-05-07
ActivityFiltered screen, ActivityDetail styles, Activity dates grouping, Send screen styles

limpbrains:
- new ActivityFiltered screen
- on ActivityDetail screen you now can switch to tx detail view
- ActivilyList on main screen
- items in ActivilyList are grouped by date

https://user-images.githubusercontent.com/155891/167251136-0657de81-fbd3-4d84-8340-20f4dee65375.mp4


src/styles/components.ts
```diff
@@ -35,6 +35,10 @@ import {
 	userIcon,
 	gitBranchIcon,
 	noteIcon,
+	checkCircleIcon,
+	clockIcon,
```
coreyphillips, line 39: `clockIcon` is missing from `../assets/icons/wallet`

limpbrains:
Fixed. 
We need a better way to work with Icons. It is hard to find which one you already have in components. Maybe importing them from svg files directly would be better, in this case you can just them in a picture viewer

coreyphillips (approved):

Jasonvdb (approved):

PR #153, opened 2022-05-20
new send onchain flow

limpbrains:
Here is new Send flow. It is still not fully implemented, a few things are still missing, like custom tx fee selection. But I think it is ready to be merged.

Right now I don't delete old screens and components and keeping them as a reference, I will do it when everything will be ready.


src/store/types/user.ts
```diff
@@ -7,7 +7,9 @@ export type TViewController =
 	| 'receiveAssetPicker'
 	| 'coinSelection'
 	| 'feePicker'
-	| 'numberPad';
+	| 'SendNavigation'
```
coreyphillips: 1. For consistency we may want to rename `SendNavigation` to `sendNavigation` unless you would prefer switching the other views to Pascal case.
2. `sendNavigation` will need to be included in the `defaultUserShape` `viewController` object [here](https://github.com/synonymdev/bitkit/blob/0fe5f5834c8f4d74a3e6a57daf4b59a55b6b5537/src/store/shapes/user.ts#L18) like so: `sendNavigation: { ...defaultViewController }`

src/store/types/user.ts
```diff
@@ -7,7 +7,9 @@ export type TViewController =
...
+	| 'numberPad'
+	| 'test1';
```
coreyphillips: `test1` can be removed.

src/store/types/user.ts
limpbrains: fixed

src/store/types/user.ts
limpbrains: fixed

coreyphillips:
The SendNavigation component requires a name change [here](https://github.com/limpbrains/bitkit/blob/8cd0ea518cf5dae02bda745e5d8898a332cb224d/src/navigation/bottom-sheet/SendNavigation.tsx#L32) as well.

coreyphillips (approved):

Jasonvdb (approved):

PR #155, opened 2022-06-07
use changeAddress in CPFP boost transaction

limpbrains:
Maybe I'm missing something, but CPFP doesn't work for me. After calling `setupCpfp` new transaction doesn't have outputs.
So I think it is better to set address by passing it to `sendMax`


coreyphillips (approved):

coreyphillips:
The current code will consolidate all UTXO's for a CPFP. I added a TODO to utilize only the required UTXO's according to the user's coin selection preferences.

PR #160, opened 2022-06-13
Slashtags SDK's integration

Nuhvi:
Read commit messages.

If you aren't running a DHT relay, it will show an error message, but I didn't want to add the script for running a relay in dev environment to this repo just yet.

You can run a DHT relay as follows:

```js
const { setupRelay } = require('dht-universal/setup-relay.js')

const main = async () => {
  const { port } = await setupRelay({
    wsServerOptions: { port: 8888 }
  })
  const relay = 'ws://localhost:' + port

  console.log('DHT mainnet Relay: ' + relay)
}

main()
```

And make sure it is in a directory with its own `package.json` not the same as Bitkit `package.json`.

```js
{
  "devDependencies": {
    "dht-universal": "^0.4.1"
  }
}
```
Screen Shot 2022-06-20 at 11 11 12


src/App.tsx
```diff
@@ -51,16 +55,30 @@ const App = (): ReactElement => {
 		return walletExists ? <RootNavigator /> : <OnboardingNavigator />;
 	}, [walletExists]);
 
+	const currentPrimaryKey = useMemo(async () => {
+		const { error, data } = await getSlashtagsPrimaryKey(
+			wallets[selectedWallet]['seedHash'],
+		);
+		return error ? null : Buffer.from(data);
+	}, [wallets[selectedWallet]['seedHash']]);
```
limpbrains: useMemo doen't support async functions, you need to use useEffect to call `getSlashtagsPrimaryKey` and then useState to store it in state component

src/components/SlashtagsProvider.tsx
```diff
@@ -0,0 +1,56 @@
...
+			}
+
+			const sdk: ISDK = await SDK.init({
+				primaryKey: await primaryKey,
+				persist: false,
+				// TODO: replace hardcoded relays with configurable relays
+				swarmOpts: { relays: ['ws://localhost:8888'] },
+			}).catch((err) => {
+				onError(err);
+			});
```
limpbrains: Is it ok, that SDK.init will exception will be eaten by `.catch` and it will try to setState with broken sdk futher?

limpbrains (changes requested):

src/components/SlashtagsProvider.tsx
Nuhvi: Fair enough I will change that.

package.json
```diff
@@ -73,6 +74,7 @@
     "react-native-toast-message": "^2.1.5",
     "react-native-touch-id": "^4.4.1",
     "react-native-typography": "^1.4.1",
+    "react-native-url-polyfill": "^1.3.0",
     "react-native-vector-icons": "^9.1.0",
     "react-native-webview": "^11.18.2",
```
Jasonvdb: Remove react-native-webview while you're at it. It's only used for the previous hack.

Jasonvdb (approved):

coreyphillips (approved):

limpbrains (approved):

PR #164, opened 2022-06-17
fix fractual satoshi

limpbrains:
There is a bug in bitkit - you can't create transaction with fiat amount. It will convert fiat to satoshi as float number, not an integer. 
To fix that bug, I'm adding rounding in `fiatToBitcoinUnit` func. 
I think this might lead to some future bugs with rounding, if you will try to switch bitcoin unit to fiat back and forth, but they should be rare.

### DIFF
--- a/src/utils/exchange-rate/index.ts
+++ b/src/utils/exchange-rate/index.ts
@@ -69,7 +69,7 @@ export const fiatToBitcoinUnit = ({
 	exchangeRate?: number;
 	currency?: string;
 	bitcoinUnit?: TBitcoinUnit;
-}): number => {
+}): string => {
 	if (!currency) {
 		currency = getStore().settings.selectedCurrency;
 	}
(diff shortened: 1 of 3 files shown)


src/utils/exchange-rate/index.ts
```diff
@@ -69,7 +69,7 @@ export const fiatToBitcoinUnit = ({
 	exchangeRate?: number;
 	currency?: string;
 	bitcoinUnit?: TBitcoinUnit;
-}): number => {
+}): string => {
```
Jasonvdb, line 72: This being a string now won't cause any calculation issues elsewhere? Should be fine but I haven't checked the rest of the app

coreyphillips (approved):

src/utils/exchange-rate/index.ts
limpbrains: I haven't changed the return type, it was string before that and it is now, I just fixed error in types

Jasonvdb (approved):

PR #169, opened 2022-06-30
Sync various styles with Figma mockups

limpbrains:
Lots of fixes in:
- margins / paddings
- text
- styles / colors
- icons

Also I removed separate Haas components, now Display and Title use it by default


MiguelMedeiros:
@limpbrains I looked at the code, but as the modification affects 50+ files, I confess that I'm not sure which screens to analyze and which specific details you changed to match the layout in Figma. I've browsed through several screens of the app and haven't found any errors that break the code, so I'm going to approve the PR. :)

Suggestion: create smalls PRs to fix one thing at a time, even if it gets a little more bureaucratic, we will be able to be sure of what we are approving and we can also revert the changes more easily if necessary.

MiguelMedeiros (approved):

coreyphillips (approved):

Jasonvdb (approved):

PR #172, opened 2022-07-05
Contacts

Nuhvi:


src/components/LabeledInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+		paddingBottom: 32,
+		borderBottomWidth: 2,
+		borderBottomColor: 'rgba(255, 255, 255, 0.1)',
+	},
+	label: {
+		fontWeight: '500',
+		fontSize: 13,
+		lineHeight: 18,
+		textTransform: 'uppercase',
+		color: '#8E8E93',
```
limpbrains: Here and bellow: 
When you insert <Text, please use components from styles/components.
Colors can be obtained using `useColors` hook, or by providing color to our components 
Here it would be <Caption13Up color="gray1"

src/components/SlashtagURL.tsx
```diff
@@ -0,0 +1,51 @@
...
+	color: string;
+}): JSX.Element => {
+	return (
+		<TouchableOpacity
+			onLongPress={async (): Promise<void> => {
+				if (url) {
+					Clipboard.setString(url);
+					console.debug('Copied slashtag url:', url);
+				}
+			}}
```
limpbrains: Looks like this function doesn't needs to be async

src/screens/Contacts/Contacts.tsx
```diff
@@ -0,0 +1,113 @@
+import React, { useState } from 'react';
+import { PlusIcon, Text, View } from '../../styles/components';
+import NavigationHeader from '../../components/NavigationHeader';
+import { StyleSheet } from 'react-native';
+import Store from '../../store/types';
+import { useSelector } from 'react-redux';
+import { useSlashtags } from '../../hooks/slashtags';
+import SafeAreaInsets from '../../components/SafeAreaInsets';
+import ContactsOnboarding from './ContactsOnboarding';
+import { TouchableOpacity } from 'react-native-gesture-handler';
```
limpbrains: why do you need TouchableOpacity from react-native-gesture-handler and not from 'react-native'?

src/screens/Contacts/ContactsOnboarding.tsx
```diff
@@ -0,0 +1,77 @@
...
+	},
+	introText: {
+		marginTop: 8,
+		fontSize: 17,
+		lineHeight: 22,
+		flex: 1,
+	},
+	button: {
+		fontWeight: '800',
+	},
```
limpbrains: I think this will not work, because this styles will be applied to <TouchableOpacity in Button component, not to the text
If you really need to overwrite text style for button, you can do:
```
<Button text={<Text01B>bla bla bla</Text01B>}
```

src/screens/Profile/ProfileEdit.tsx
```diff
@@ -0,0 +1,227 @@
...
+				<View style={styles.editDataModal}>
+					<Subtitle style={styles.addDataTitle}>Add data</Subtitle>
+					<View style={styles.editDataContent}>
+						<Input
+							label="Label"
+							value={addDataForm.key}
+							onChange={(key: string): void => {
+								setAddDataForm({ ...addDataForm, key });
+							}}
+						/>
```
limpbrains: Inside BottomSheetWrapper it is better to use BottomSheetTextInput, it will work nice with on screen keyboard.  
Please test this screen with on screen keyboard

limpbrains (changes requested):

limpbrains:
FYI: I renamed some components:
TitleHaas -> Title
DisplayHaas -> Display

src/screens/Profile/ProfileEdit.tsx
Nuhvi: Used `BottomSheetTextInput` in `LabeledInput` not sure if this is enough, and didn't get a chance to test with an on screen keyboard yet. honestly, I don't know how even.

Nuhvi:
closing to open again from `synonymdev:contacts`

PR #178, opened 2022-07-12
About screen: fix svg import pattern

MiguelMedeiros:
## Description

Fix svgs pattern at `about screen`.

## How to review
Click on the `Cog Icon` > `Settings` > `About Bitkit`.
Check if icons are loading properly.


limpbrains (approved):

limpbrains:
I would prefer if we use same default white color for Icons everywhere. Because if Brand color will be changed, we will need to update All the icons. If color would be defined only in one place, that would be much simpler

Jasonvdb (approved):

limpbrains (approved):

coreyphillips (approved):

PR #181, opened 2022-07-13
Tether coming soon

limpbrains:
Lightning is still shown on send/receive popup, but that is ok for now


src/components/AssetCard.tsx
```diff
@@ -9,15 +15,22 @@ const AssetCard = ({
...
+	disabled?: boolean;
 }): ReactElement => {
+	const container = useMemo(
+		() => [styles.container, disabled && { opacity: 0.3 }],
+		[disabled],
+	);
+
 	return (
-		<Pressable style={styles.container} onPress={onPress} color="transparent">
+		<Pressable style={container} onPress={onPress} color="transparent">
```
coreyphillips: If `disabled` is true we should also disable `onPress`.

MiguelMedeiros: I agree with corey, it would be nice to disable the click or show a popup with some information about what's to come.

https://user-images.githubusercontent.com/5798170/178824977-88b737b4-a91c-478f-a7fb-ea24ee96472a.mp4

BitcoinErrorLog:
The longer we go without progress on token support, the less comfortable I am with teasing it as a feature at all, tbh. I don't want a Tether ad in our app for 9 months...

Jasonvdb:
@BitcoinErrorLog want to leave it out the code base entirely? If we leave it in and make it open source people will find it

src/components/AssetCard.tsx
limpbrains: fixed

BitcoinErrorLog:
I don't mind people finding it in the code, but I am having Aldert update the main prototype to remove Tether/token concepts for now.

PR #204, opened 2022-07-22
Fix jest mocks config

limpbrains:
Also I removed jest config section from package.json to prevent multiple configurations warning
```

● Multiple configurations found:
    * /Users/limpbrains/s/bitkit/jest.config.js
    * `jest` key in /Users/limpbrains/s/bitkit/package.json

  Implicit config resolution does not allow multiple configuration files.
  Either remove unused config files or select one explicitly with `--config`.

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html
```


Jasonvdb (approved):
Nice thanks!

PR #206, opened 2022-07-23
Add `Swipe to hide` switch feature to settings menu

MiguelMedeiros:
## Description
Add `Swipe to hide` switch feature to settings menu

## Preview

https://user-images.githubusercontent.com/5798170/180585919-9499aac6-66c9-4c49-8fbc-5a4e3863de2d.mp4

### DIFF
--- a/src/store/types/settings.ts
+++ b/src/store/types/settings.ts
@@ -40,7 +40,7 @@ export interface ISettings {
 	unitPreference: 'asset' | 'fiat';
 	showSuggestions: boolean;
 	transactionSpeed: TTransactionSpeed;
-	showSuggestions: boolean;
+	swipeBalanceToHide: boolean;
 	[key: string]: any;
 }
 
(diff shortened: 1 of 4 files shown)


limpbrains (changes requested):
please add `swipeBalanceToHide: true` to `store/shapes/settings.ts`, I think it should be turned on by default.

src/store/types/settings.ts
```diff
@@ -40,7 +40,7 @@ export interface ISettings {
 	unitPreference: 'asset' | 'fiat';
 	showSuggestions: boolean;
 	transactionSpeed: TTransactionSpeed;
-	showSuggestions: boolean;
```
limpbrains, line 43: I'm not sure if showSuggestions should be removed

MiguelMedeiros: I can make this change in another PR if you want, but we need to remove it, because It is duplicated: line 41 and 43.

limpbrains (approved):

Jasonvdb (approved):

PR #210, opened 2022-07-26
Restore rn_edit_text_material.xml file

limpbrains:
It was removed in #198. But without it `npx react-native run-android` fails with `AAPT: error: resource drawable/rn_edit_text_material (aka com.bitkit:drawable/rn_edit_text_material) not found.` error

PS we need test build for PRs


MiguelMedeiros (approved):
- [x] Tested the build on android and it worked. 
- [x] Tested sending bitcoins and the confetti animation worked. 

## Preview
image
image

Jasonvdb:
Thanks! We have a GH action for building android on every push to master, will update it to check PRs as well

Jasonvdb (approved):

PR #211, opened 2022-07-26
Add @synonymdev/react-native-ldk

coreyphillips:
This PR:
 - Adds `@synonymdev/react-native-ldk` dependency to `package.json`.
- Updates `yarn.lock` & `Podfile.lock` accordingly.
- Updates `README.md`.
- Adds `onReceive` param to `subscribeToHeader` method in `electrum.ts`.
- Adds "Lightning Node Info" button to Networks in Settings to display the LDK node ID for testing.
- Adds nodeId to lightning redux store.
- Updates `LightningNodeInfo.tsx` component to only display LDK node ID at this time.

The LDK Node ID can be found when in Regtest by navigating to `Settings->Networks->Lightning Node Info`.


src/utils/lightning/index.ts
```diff
@@ -28,35 +46,223 @@ export const wipeLndDir = async ({
...
+			seed,
+		};
+		await Keychain.setGenericPassword(name, JSON.stringify(account), {
+			service: name,
+		});
+		return true;
+	} catch {
+		return false;
+	}
+};
```
limpbrains: What happens if keychain save fails?Looks like result of `setAccount()` is not checked anywhere, is it safe to ignore it?

limpbrains (changes requested):

src/utils/lightning/index.ts
coreyphillips: Fixed [here](https://github.com/synonymdev/bitkit/pull/211/commits/78e5e75224959c5d83007e088f5f365e87fddd9f).

limpbrains (approved):

src/utils/lightning/index.ts
```diff
@@ -28,35 +46,234 @@ export const wipeLndDir = async ({
...
+
+/**
+ * Returns the current LDK node id.
+ * @returns {Promise<Result<string>>}
+ */
+export const getNodeId = async (): Promise<Result<string>> => {
+	try {
+		const nodeIdResponse = await ldk.nodeId();
+		if (nodeIdResponse.isErr()) {
+			return err(nodeIdResponse.error.message);
```
Jasonvdb: I think if we push Result to it's own lib and share it everywhere then you won't have to double wrap these calls just to get the same result

Jasonvdb:
I'm not coming right, in regtest but LDK won't start because of this:
Simulator Screen Shot - iPhone 13 - 2022-07-28 at 14 16 24

src/utils/lightning/index.ts
coreyphillips: Agreed. We have multiple projects using `Result` that could benefit from pulling from the same source.

The code referenced above was unnecessary in the `getNodeId` method and updated [here](https://github.com/synonymdev/bitkit/commit/906154d5402a3e641f8b21e8645bcf9a7238a149#diff-3898c176b9ed4ed723d2e4c5e7ba04e4d07ff33d39ddc16386e6f4cd41748b1cL271)

coreyphillips:
> I'm not coming right, in regtest but LDK won't start because of this: Simulator Screen Shot - iPhone 13 - 2022-07-28 at 14 16 24

Fixed [here](https://github.com/synonymdev/bitkit/commit/906154d5402a3e641f8b21e8645bcf9a7238a149#diff-3898c176b9ed4ed723d2e4c5e7ba04e4d07ff33d39ddc16386e6f4cd41748b1cL178).

Jasonvdb (approved):

PR #219, opened 2022-08-01
Fix: Settings glow effect

pwltr:
Changes:
- update List component styles to fix glow effect

Notes:
Removed `paddingBottom` from generic List component which is also used in `LightningChannels` component, which seems to currently be unused. Not sure if that's breaking the styles there.

| Before         | After     | 
|--------------|-----------|
| Simulator Screen Shot - iPhone 13 - 2022-08-01 at 17 06 09 | Simulator Screen Shot - iPhone 13 - 2022-08-01 at 16 36 47      |

### DIFF
--- a/src/components/List.tsx
+++ b/src/components/List.tsx
@@ -214,15 +214,11 @@ const List = ({
 				// eslint-disable-next-line react-hooks/exhaustive-deps
 			}, [])}
 			stickySectionHeadersEnabled={false}
-			contentContainerStyle={styles.contentContainerStyle}
 		/>
 	);
 };
 
 const styles = StyleSheet.create({
-	contentContainerStyle: {
-		paddingBottom: 55,
-	},
 	row: {
 		height: 55,
 	},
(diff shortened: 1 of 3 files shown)


MiguelMedeiros:
I don't think it is a problem of padding or height. 
Maybe you should change the settings menu to transparent background instead of black?

## Preview
I'm still getting the glowing orange shadow below the menu.
image

Nuhvi:
Removing the `zIndex: -1` in [SettingsMenu](https://github.com/synonymdev/bitkit/blob/67ad4c865bdc5876b7e18041745328c4bdc4b9e5/src/screens/Settings/index.tsx#L80) solves this.

pwltr:
Thank you both, fixed now.

EDIT: sorry, getting ahead of myself here, changing z-index causes the last item to not be clickable.

Nuhvi:
Does React Native have `pointer-events: none`?

Edit: seems so https://stackoverflow.com/questions/53326871/using-image-component-without-it-blocking-touch

pwltr:
yes, interestingly enough it has to be passed as a prop, it is not implemented on `style`.

pwltr:
@MiguelMedeiros @coreyphillips Should be good now

MiguelMedeiros (approved):

coreyphillips (approved):

PR #226, opened 2022-08-03
feat: reduce exchange rate services to Blocktank

pwltr:
Blocked by https://github.com/synonymdev/blocktank-fx-service/pull/3

### Changes:
- use [Blocktank FX exchange rate service API](https://github.com/synonymdev/blocktank-fx-service)
- remove exchange rate service selection
- add SettingsView search filter, enable for Currencies screen
- navigate the user back on choice

### Notes:
Currency symbols should appear once linked PR is merged

### DIFF
--- a/src/components/SearchInput.tsx
+++ b/src/components/SearchInput.tsx
@@ -1,12 +1,18 @@
 import React, { ReactElement, memo } from 'react';
-import { StyleSheet, View } from 'react-native';
+import { StyleProp, StyleSheet, View, ViewStyle } from 'react-native';
 import { TextInput, MagnifyingGlassIcon } from '../styles/components';
 
-const SearchInput = ({ style, children, ...props }): ReactElement => {
+type SearchInputProps = {
+	style?: StyleProp<ViewStyle>;
+	children?: ReactElement;
+	[x: string]: any;
+};
+
+const SearchInput = ({ style, children, ...props }: SearchInputProps) => {
 	const rootStyle = StyleSheet.compose(style, styles.root);
 
 	return (
-		<View style={rootStyle}>
+		<View style={style ? rootStyle : null}>
 			<MagnifyingGlassIcon style={styles.icon} />
 			<TextInput style={styles.input} placeholder="Search" {...props} />
 			{children && <View style={styles.tags}>{children}</View>}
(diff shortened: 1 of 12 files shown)


MiguelMedeiros (approved):

MiguelMedeiros:
Tested ACK.
Also tested https://github.com/synonymdev/blocktank-fx-service/pull/3.

## Review Preview
https://user-images.githubusercontent.com/5798170/182730050-56bbdb63-e67f-4814-8170-bdd449fa4327.mp4

src/components/SearchInput.tsx
```diff
@@ -1,12 +1,18 @@
...
+	children?: ReactElement;
+	[x: string]: any;
+};
+
+const SearchInput = ({ style, children, ...props }: SearchInputProps) => {
 	const rootStyle = StyleSheet.compose(style, styles.root);
 
 	return (
-		<View style={rootStyle}>
+		<View style={style ? rootStyle : null}>
```
limpbrains: If no style provided to SearchInput, it should use `styles.root`, not null
Also I suggest we swap `StyleSheet.compose(style, styles.root)` to `StyleSheet.compose(styles.root, style)` so that base styles can be overwritten.

src/utils/exchange-rate/index.ts
```diff
@@ -2,63 +2,27 @@ import { default as bitcoinUnits } from 'bitcoin-units';
...
-	IExchangeRates,
-	supportedExchangeTickers,
-} from './types';
+import { defaultDisplayValues, IDisplayValues, IExchangeRates } from './types';
 
 export const getExchangeRates = async (): Promise<Result<IExchangeRates>> => {
-	let { exchangeRateService } = getStore().settings;
+	// TODO: pull this out into .env
+	const response = await fetch('http://35.233.47.252:443/fx/rates/btc');
+	const { tickers } = await response.json();
```
limpbrains: this func should return err(e) in case of the error

limpbrains (changes requested):

src/components/SearchInput.tsx
pwltr: What do you think of this syntax: `<View style={[styles.root, style]}>`? It's less pure but I have noticed it being used in the codebase in a lot of places.

limpbrains: I'm ok with it. Technically, style property also could be an array, but we are not writing component library, so that is fine

limpbrains (approved):

PR #229, opened 2022-08-04
fix: add pull-to-refresh on wallets/index screen

limpbrains:
Adding pull-to-refresh on main screen. For now it only does the same as in wallet details, refresh active wallet. We can add something else later, probably slashtags refresh?


coreyphillips:
We can probably change `await updateOnChainActivityList();` in [src/utils/wallet/index.ts](https://github.com/synonymdev/bitkit/blob/e642287fbca2dc13d1eec943165bb754fedfac5b/src/utils/wallet/index.ts#L113) to `await updateActivityList();`. This would allow us to remove `await updateActivityList();` from the `onRefresh` method in [src/screens/Wallets/index.ts](https://github.com/synonymdev/bitkit/blob/e642287fbca2dc13d1eec943165bb754fedfac5b/src/screens/Wallets/index.tsx#L48) and prevent us from calling the method twice.

Otherwise, LGTM!

coreyphillips (approved):

PR #236, opened 2022-08-07
feat: Scanner UI

pwltr:
### Changes:
- add Scanner UI
- add read QR from file
- fix Electrum Server port TextField value
- fix some typos

### Notes:
~The UI currently doesn't have the blurred overlay~ Blur only on iOS and missing rounded corners as per design. Additional Annotations below.


src/utils/wallet/transactions.ts
```diff
@@ -68,7 +69,12 @@ export const parseOnChainPaymentRequest = (
 			return err(data);
 		}
 
-		let validateAddressResult = validateAddress({ address: data });
+		let validateAddressResult = validateAddress({
+			address: data,
+			selectedNetwork: selectedNetwork
+				? EAvailableNetworks[selectedNetwork]
+				: undefined,
+		});
```
pwltr: This was returning the wrong network in regtest mode since Regtest and Testnet p2pkh/p2sh addresses are valid for both networks (see `networks.ts`).  So now we're passing the network we actually want to validate against.

src/utils/scanner.ts
```diff
@@ -265,7 +272,7 @@ export const handleData = async ({
 					outputs: [{ address, value: amount }],
 				},
 			}).then();
-			refreshWallet().then();
+			await refreshWallet({ showNotification: false });
```
pwltr: This was logging an error, the argument is not optional

src/screens/Scanner/index.tsx
```diff
@@ -19,13 +34,13 @@ const ScannerScreen = ({ navigation }): ReactElement => {
...
 	const onRead = async (data): Promise<void> => {
-		const res = await decodeQRData(data);
+		const res = await decodeQRData(data, selectedNetwork);
 
 		if (res.isErr() || (res.isOk() && res.value.length === 0)) {
 			showErrorNotification(
 				{
 					title: 'Error',
-					message: 'Failed to detect any readable data',
+					message: 'Failed to detect any address data',
```
pwltr: Changed this to be more specific to the error

src/components/Camera.tsx
```diff
@@ -29,34 +33,40 @@ const _Camera = ({
...
+					style={styles.container}
+					onBarCodeRead={({ data }): void => {
+						if (_data !== data) {
+							setData(data);
+							onBarCodeRead(data);
+						}
+					}}
+					onMountError={(): void => {
+						console.log(
+							'An error was encountered when loading the camera. Please ensure BitKit has permission to use this feature in your phone settings.',
```
limpbrains: User will not see this error, perhaps we need to show Notification here

limpbrains (changes requested):

src/components/Camera.tsx
pwltr: added a toast for this error with a shorter text so it's not cut off.

limpbrains (approved):

src/screens/Scanner/index.tsx
Jasonvdb: Maybe confirm with Alder/John for the copy here. We will support scanning much more than just addresses.

src/utils/scanner.ts
Jasonvdb: Refreshing can take a few seconds so I think that's why this wasn't using it this way. This way the scanning needs to wait for the whole wallet to be refreshed. Rather `refreshWallet({ showNotification: false }).catch(e => {handleMe(e)})`

Jasonvdb (changes requested):

src/utils/wallet/transactions.ts
coreyphillips: Good catch. The only time we may want to distinguish between testnet and regtest, however, is when we're specifically using regtest. Otherwise, we may lose the ability to distinguish and switch between mainnet and testing states. Even then, in most situations we'll be using bech32 address where it won't be necessary. So I would suggest the following condition for regtest in the `parseOnChainPaymentRequest` method:
```
if (!selectedNetwork) {
  selectedNetwork = getSelectedNetwork();
}
let validateAddressResult = validateAddress({
	address: data,
	selectedNetwork: selectedNetwork === 'bitcoinRegtest' ? EAvailableNetworks[selectedNetwork] : undefined,
});
```

pwltr: sounds good [6456f54](https://github.com/synonymdev/bitkit/pull/236/commits/6456f54e478773d89fcf1de362985d6b7708ee37)

src/screens/Scanner/index.tsx
pwltr: I see, address is definitely not generic enough here. I reverted it for now.

src/utils/scanner.ts
pwltr: fixed [85eacce](https://github.com/synonymdev/bitkit/pull/236/commits/85eacce8146fea47f27a2569776d1a0c20580cee)

Jasonvdb (approved):

PR #239, opened 2022-08-09
fix: exchange rate undefined property causing crash

pwltr:
### Changes
- use fallback value of 0 everywhere

### DIFF
--- a/src/hooks/displayValues.ts
+++ b/src/hooks/displayValues.ts
@@ -24,7 +24,7 @@ export default function useDisplayValues(
 
 	const displayValues: IDisplayValues = useMemo(() => {
 		//Exchange rates haven't loaded yet
-		if (!exchangeRates) {
+		if (Object.entries(exchangeRates).length === 0) {
 			return defaultDisplayValues;
 		}
 
--- a/src/hooks/displayValues.ts
+++ b/src/hooks/displayValues.ts
@@ -41,3 +41,17 @@ export default function useDisplayValues(
 
 	return displayValues;
 }
+
+/**
+ * Returns 0 if no exchange rate for currency found or something goes wrong
+ */
+export const useExchangeRate = (currency = 'EUR'): number => {
+	try {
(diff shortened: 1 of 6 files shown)


pwltr:
Sorry about that, app shouldn't crash anymore with this.

coreyphillips:
It may make sense to create a `useExchangeRate` hook for components and a `getExchangeRate` method for methods and apply them each where needed.

Suggested Hook (src/hooks/displayValues):
```
export const useExchangeRate = (currency = 'EUR'): number => {
	try {
		const exchangeRates = useSelector(
			(state: Store) => state.wallet.exchangeRates,
		);
		return exchangeRates[currency]?.rate ?? 0;
	} catch {
		return 0;
	}
};
```

Suggested Method (src/utils/exchange-rate):
```
export const getExchangeRate = (currency = 'EUR'): number => {
	try {
		const exchangeRates = getStore().wallet.exchangeRates;
		return exchangeRates[currency]?.rate ?? 0;
	} catch {
		return 0;
	}
};
```

pwltr:
Is the `try...catch` needed here? There is nothing here that can really throw. What am I missing?

I guess we could put everything in the selector, and export that:
```
export const selectExchangeRate = state => state.wallet.exchangeRates[state.settings.selectedCurrency]?.rate ?? 0
```

```
const Component = () => {
  const exchangeRate = useSelector(selectExchangeRate);

  return <></>;
};
```

```
const Function = () => {
  const store = getStore()
  const exchangeRate = selectExchangeRate(store)

  return;
};
```

What do you think?

coreyphillips:
In the event an unknown currency is passed through or an unforeseen error occurs it doesn't hurt to have a try/catch to fall back on.
It may not be the best to pass the entire store into selectExchangeRate when we can access the exchange rate by passing only the desired currency into a hook or method similar to what was mentioned above, otherwise.

Jasonvdb (approved):

PR #244, opened 2022-08-10
feat: detox

limpbrains:


limpbrains:
Current status:
- Android: doesn't build because of skia compilation error https://github.com/Shopify/react-native-skia/issues/728
- iOS: doesn't run because of Detos error https://github.com/wix/Detox/issues/3462

PR #254, opened 2022-08-16
perf(wallet): Reduced Address Generation On-Start

coreyphillips:
This PR reduces the number of addresses generated on-start. Instead of generating 30+ addresses that include all address types, we're only generating 4 `p2wpkh` addresses. As a result, `p2sh` and `p2pkh` addresses will no longer appear in the receive view's address carousel. However, we will be phasing this out in a future PR in favor of a unified qr-code (on-chain+lightning).

Ultimately, this should speed up tests for testers on live devices, Android emulators and simulators until we are ready to implement the [nodejs-mobile](https://github.com/synonymdev/bitkit/tree/nodejs-mobile) update.

This PR:
- Adds `updateAllAddressTypes` param to `refreshWallet` method in `utils/wallet/index.ts`.
- Creates `constants.ts` and set `GENERATE_ADDRESS_AMOUNT` to 2 in `utils/wallet/constants.ts`.
- Passes `addressType` to `updateAddressIndexes` in `BitcoinNetworkSelection.tsx`.

### DIFF
--- a/src/screens/Settings/Bitcoin/BitcoinNetworkSelection.tsx
+++ b/src/screens/Settings/Bitcoin/BitcoinNetworkSelection.tsx
@@ -10,7 +10,10 @@ import {
 } from '../../../store/actions/wallet';
 import { getNetworkData } from '../../../utils/helpers';
 import { startWalletServices } from '../../../utils/startup';
-import { getCurrentWallet } from '../../../utils/wallet';
+import {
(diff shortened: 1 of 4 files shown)


limpbrains (approved):

Jasonvdb (approved):

limpbrains:
Does this mean that gap limit now will be 4?

PR #256, opened 2022-08-17
fix: change slashtag url to "Your name" on main screen

limpbrains:
image


Nuhvi:
For the record: I didn't like "Your name" choice and deliberately chose the URL as a fallback.

Nuhvi (approved):

PR #259, opened 2022-08-17
fix: Typescript errors

pwltr:
### Changes
- update related dev dependencies
- update tsconfig to remove most common error (temp)
- fix a lot of minor type errors
- add some navigation types
- update redux store config with `@reduxjs/toolkit`
- fix a state mutation in activity reducer
- add GitHub workflow for type-checking on PR

### Notes
The `configureStore` from `@reduxjs/toolkit` replaces `createStore` which was flagged as deprecated by TS. It contains some common middleware by default (see https://redux-toolkit.js.org/api/getdefaultmiddleware/#included-default-middleware). It now logs a warning if a mutation in reducers is detected and also expects serializable data in dispatches. That's why we need to add an [exception](https://redux-toolkit.js.org/usage/usage-guide#use-with-redux-persist) for `redux-persist`

Only necessary type fixes for now. Some `@ts-ignore`s to be removed and types to be added. But we can run the type-checker on PRs and maybe on `pre-commit` from this point on.


pwltr:
Had to remove the 2nd ESLint check in the action. I think it isn't compatible with latest `eslint` version. Let me know if I should revert the version update.

MiguelMedeiros (approved):

MiguelMedeiros:
Tested ACK.

## Preview
image

coreyphillips (approved):

Jasonvdb (approved):

PR #261, opened 2022-08-18
fix(wallet): Address Types Fixes

coreyphillips:
This PR:
- Added checks where needed to ensure addresses of a given type have been generated before proceeding.
- Moved `subscribeToHeader` further up in `startWalletServices` method in `src/utils/startup/index.ts` to ensure LDK can acquire header information.
- Updated `UPDATE_TRANSACTIONS` & `DELETE_ON_CHAIN_TRANSACTION` in `src/store/reducers/wallet.ts`.

### DIFF
--- a/src/utils/startup/index.ts
+++ b/src/utils/startup/index.ts
@@ -113,6 +113,13 @@ export const startWalletServices = async ({
 							'Unable to connect to Electrum Server',
 					});
 				}
+				let onReceive = (): void => {};
+				// Ensure LDK syncs when a new block is detected.
+				if (lightning) {
+					onReceive = lm.syncLdk;
+				}
+				// Ensure we are subscribed to and save new header information.
+				await subscribeToHeader({ selectedNetwork, onReceive });
 			}
 
 			const walletExists = await checkWalletExists();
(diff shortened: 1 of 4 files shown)


src/utils/wallet/electrum.ts
```diff
@@ -62,6 +62,13 @@ export const getUtxos = async ({
 				if (!selectedWallet) {
 					selectedWallet = getSelectedWallet();
 				}
+				// Check if addresses of this type have been generated. If not, skip.
+				if (
+					Object.keys(currentWallet.addresses[selectedNetwork][addressTypeKey])
+						?.length <= 0
```
limpbrains, line 68: `Object.keys()` will always have `.length`, no need to check that

src/utils/startup/index.ts
```diff
@@ -113,6 +113,13 @@ export const startWalletServices = async ({
 							'Unable to connect to Electrum Server',
 					});
 				}
+				let onReceive = (): void => {};
+				// Ensure LDK syncs when a new block is detected.
+				if (lightning) {
+					onReceive = lm.syncLdk;
```
limpbrains: maybe we should refresh whole wallet when new block arrives, not just ldk?
Currently we are listening only receiving addresses. But it is possible that there is outgoing tx created from another devices has been broadcasted, or incoming tx to change address.

limpbrains (approved):

src/utils/startup/index.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/261/commits/657474524c6fc5b0a2d912260578354e49fdb6ce).

Previously, `refreshWallet` was always called when a new block was found. The `onReceive` param for `subscribeToHeaders` was used to pass additional items we wanted to update as well (lightning in this case). This can be confusing, so updated it by removing the default `refreshWallet` call in `subscribeToHeaders` and now pass it through the `onReceive` param instead.

limpbrains: Thanks!

src/utils/wallet/electrum.ts
coreyphillips: In the event an address type was not generated for a given wallet `Object.keys` will return a length of `0` on `currentWallet.addresses[selectedNetwork][addressTypeKey]`. Currently, this will be the case for all `p2pkh` & `p2sh` address types since the address generation patch. Once we're able to generate addresses without slowing the UI we'll be able to generate at least one address per type and this will no longer be an issue.

Jasonvdb (approved):

PR #275, opened 2022-08-22
feat: unified receive flow

pwltr:
### Changes
- update receive flow UI
- add "receive" reducer
- update "metadata" reducer
- trigger tag matching logic on wallet sync
- add some helpers (reset metadata & a hook)
- fixed `wallet` type and resulting errors & related bugs

### Notes
- outstanding task wrt. onchain address reuse
- tags only work for onchain txs right now, will need to revisit when Lightning UI is implemented more
- some code duplication that could be improved (NumberPad, Tags screen)
- there is a weird issue with NumberPad popping up after reopening the app but I believe it's a general problem that I'll look at seperately


src/store/actions/wallet.ts
```diff
@@ -740,11 +740,15 @@ export const updateOnChainTransaction = async ({
...
 							(o) => o.address === output.address,
 						);
-						if (foundOutput?.length) {
-							outputs[foundOutput.index] = output;
+						if (foundOutputs?.length) {
+							foundOutputs.forEach((foundOutput) => {
+								if (foundOutput.index) {
+									outputs[foundOutput.index] = output;
+								}
+							});
```
pwltr: @coreyphillips There was a bug here: `foundOutput` is an array and not a single `IOutput`. Can you confirm that this fix is what this was intending to do?

src/hooks/helpers.ts
```diff
@@ -7,3 +7,18 @@ export const usePrevious = (value): any => {
 	}, [value]);
 	return ref.current;
 };
+
+export const useEffectOnlyOnce = (callback, dependencies, condition) => {
```
limpbrains, line 11: Looks like `useEffectOnlyOnce` not used anywhere

src/navigation/bottom-sheet/ReceiveNavigation.tsx
```diff
@@ -21,7 +23,7 @@ const ReceiveNavigation = (): ReactElement => {
 		useSelector(
 			(store: Store) => store.user.viewController?.receiveNavigation,
 		) ?? {};
-	const snapPoints = useMemo(() => [600], []);
+	const snapPoints = useMemo(() => ['87%'], []);
```
limpbrains: Why switch to percentage?

src/screens/Wallets/Receive/index.tsx
```diff
@@ -3,134 +3,122 @@ import React, {
...
+	useEffect(() => {
+		if (tags.length !== 0 && receiveAddress) {
+			updateMetaIncTxTags(receiveAddress, lightningInvoice, tags);
+		}
+	}, [tags]);
+
+	// Getting receive address shouldn't take long, so we don't show a spinner
+	if (!receiveAddress) {
+		return <></>;
+	}
```
limpbrains: receiveAddress generated in useMemo, it runs synchronously, so you don't need to wait for it.
But `receiveAddress` can be undefined if `getReceiveAddress` or `getReceiveNewAddress` fails. 
Maybe show error in this case? I'm not sure

limpbrains (changes requested):

src/store/actions/wallet.ts
coreyphillips: Good catch! However, we should either remove the` if (foundOutput.index) {` condition on line 748 or if we want to be really safe, replace it with `if (foundOutput?.address) {`

pwltr: We need to make sure `foundOutput.index` exists before using it as an index for the array as it is an optional field on `IOutput`. What should we do if the output doesn't have the index?

src/hooks/helpers.ts
pwltr: This was in anticipation of the `getNewAddress` method which is not yet implemented. I'm assuming it will be asynchronous as we might have to generate new addresses on demand. Will remove this if mistaken.

src/screens/Wallets/Receive/index.tsx
pwltr: This will also change with the `getNewAddress` function. Depending on if it's async or not I will change this.

src/navigation/bottom-sheet/ReceiveNavigation.tsx
pwltr: No reason, but the sheet is larger now. Should we switch back to pixels?

src/store/actions/wallet.ts
coreyphillips: The index may be `0` so we'll just need to ensure it's not failing that condition. So something like, `if (foundOutput?.index >= 0)`

src/hooks/helpers.ts
coreyphillips: Leave it for now. If I end up not using it when implementing `getNewAddress` we can consider tossing it.

src/navigation/bottom-sheet/ReceiveNavigation.tsx
limpbrains: I think yes, because of small screen devices. We don't test them well.

pwltr: update here: https://github.com/synonymdev/bitkit/pull/275/commits/7eeeabdb0f210e97dd79478fff0d4cc6dfeeb839

Using screen height minus top SafeAreaInset + height of HeaderBar. I think this way it should look pretty much the same on all devices (untested). Seems ok?

src/store/actions/wallet.ts
pwltr: fixed here https://github.com/synonymdev/bitkit/pull/275/commits/f3bf95b745974f374f95eccf50ceee165a9302bc in a way that makes TS happy. Still seems off to me, if multiple `foundOutputs` are missing the index they are going to overwrite each other on `outputs[0]`. Not really sure what this code does tbh otherwise I would test.

coreyphillips: This code mostly applies when we add or specify more than one output beyond the changeAddress when sending a transaction. Because we currently only allow for sending to one address, it is unlikely that this change will impact the current transaction creation process. However, this change does affect the transaction creation flow.

With that said, this change is beyond the scope of this PR. Lets revert this change back to the original, create an issue for the bug you discovered here and I'll create a separate PR to test and review the needed changes.

src/navigation/bottom-sheet/ReceiveNavigation.tsx
limpbrains: Still needs some adjustments for Android. probably limit qrcode size, when screen is wide 
image

this is how 87% looks like:
image

PS also eslint error

pwltr: Added a `maxHeight` for the QR https://github.com/synonymdev/bitkit/pull/275/commits/ae5eb424a1f432c755a4ed74cabc2e30bd47c111

Screenshot_1661416386

src/store/actions/wallet.ts
pwltr: reverted https://github.com/synonymdev/bitkit/pull/275/commits/f73cde5e82ee79d87bfd29ed69a4f70812d118ac

limpbrains (approved):

coreyphillips (approved):

PR #280, opened 2022-08-24
fix(settings): bitcoin address type menu

MiguelMedeiros:
## Description

This PR changes the UX when you select the default `bitcoin address type` on the `settings menu`.

**There is a delay when you change the address but it will be fixed by another PR which optimizes the performance to create a new bitcoin address.**

## Preview

https://user-images.githubusercontent.com/5798170/186425399-892444e9-bf6a-46de-af05-3e17928ab1c2.mp4

### DIFF
--- a/src/screens/Settings/AddressTypePreference/index.tsx
+++ b/src/screens/Settings/AddressTypePreference/index.tsx
@@ -1,4 +1,5 @@
 import React, { memo, ReactElement, useMemo, useState } from 'react';
+import { useNavigation } from '@react-navigation/native';
 
 import { IListData } from '../../../components/List';
 import SettingsView from '../SettingsView';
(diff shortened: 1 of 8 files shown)


src/screens/Settings/AddressTypePreference/index.tsx
```diff
@@ -81,7 +93,7 @@ const AddressTypeSettings = (): ReactElement => {
 
 	return (
 		<SettingsView
-			title={'Address types preference'}
+			title={'Bitcoin Address Types'}
```
pwltr: nit: should be "Bitcoin Address Type"

src/store/types/settings.ts
```diff
@@ -38,6 +43,7 @@ export interface ISettings {
 	unitPreference: 'asset' | 'fiat';
 	showSuggestions: boolean;
 	transactionSpeed: TTransactionSpeed;
+	addressType: TAddressType;
 	[key: string]: any;
```
pwltr: Could remove this `[key: string]: any;` while we're here and fix the ensuing type errors

pwltr (approved):

src/screens/Settings/AddressTypePreference/index.tsx
MiguelMedeiros: Done! [efd3043](https://github.com/synonymdev/bitkit/pull/280/commits/efd304323541383f551dc5c1f4160bda52c2d812)

src/store/types/settings.ts
MiguelMedeiros: Done! [5a70860](https://github.com/synonymdev/bitkit/pull/280/commits/5a7086028175b789ae903ae9a5009a70f75d8efa)

src/screens/Settings/AddressTypePreference/index.tsx
```diff
@@ -8,14 +9,19 @@ import { getSelectedAddressType } from '../../../utils/wallet';
 import { capitalize } from '../../../utils/helpers';
 import { updateSelectedAddressType } from '../../../store/actions/wallet';
 import { TAddressType } from '../../../store/types/wallet';
+import { updateSettings } from '../../../store/actions/settings';
 
 const AddressTypeSettings = (): ReactElement => {
+	const navigation = useNavigation();
```
limpbrains: navigation object is already in props, no need to use the hook
`const AddressTypeSettings = ({ navigation }): ReactElement => {`

src/store/types/settings.ts
pwltr: Do you want to fix the 4 type errors that are now there as well? I know it's unrelated but we can clean up as we go. Up to you, I can also do it in another PR.

src/screens/Settings/AddressTypePreference/index.tsx
MiguelMedeiros: Fixed here: [16901ae](https://github.com/synonymdev/bitkit/pull/280/commits/16901ae2fdad18a23f008d073f8488b4c1acb1d6)

Also changed the other files that used the same object.

limpbrains (approved):
Still a few typescript errors

PR #281, opened 2022-08-24
feat(ui): add ConnectivityIndicator

pwltr:
### Changes
- add `NetInfo` listener in `App.tsx`
- add Electrum server connectivity listener
- add `ConnectivityIndicator` on home screen
- add a boolean for `isElectrumConnected` to `user` store
- build custom toast (awaiting final design)

### Notes
- "Retry" button has been removed as discussed with Aldert
- using `user.isOnline` from store instead of `useNetInfo` hook as we might want to use this for Electrum connection
- added a minimal pubSub around `electrum.pingServer()` in an attempt to have this be clean
- we're now trying to reconnect to the Electrum server every 10 seconds

### Preview
Toggling WiFi:

https://user-images.githubusercontent.com/8538369/186459190-74ff4497-9028-4cfb-b7e1-811f1e89fa2a.mp4


MiguelMedeiros (approved):

limpbrains:
I think it is better to listen for Electrum connection, is it hard?

pwltr:
> I think it is better to listen for Electrum connection, is it hard?

I don't know I haven't looked at it yet. It would be good to have but I suggest we do that in another PR. Or was it a requirement (can't remember)? The UI would be kind of confusing imo, if Electrum connection fails the user will think his internet connection is the problem.

Jasonvdb:
Think we do need to also show when electrum is down. Maybe Aldert can help with some copy differences for when our internet is connected but it's just electrum that is down.

limpbrains:
> > I think it is better to listen for Electrum connection, is it hard?
> 
> I don't know I haven't looked at it yet. It would be good to have but I suggest we do that in another PR. Or was it a requirement (can't remember)? The UI would be kind of confusing imo, if Electrum connection fails the user will think his internet connection is the problem.

It would be more confusing if it will not show Electrum connection status.
It should be like connection indicator in Telegram, which shows connection to telegram network, not my wifi status.

Jasonvdb:
@pwltr if you just fix those new lint warnings added here I'll merge 👍

Jasonvdb (approved):

PR #282, opened 2022-08-24
fix(wallet): Create Wallet Fix

coreyphillips:
The create wallet method was receiving an error when attempting to create the wallet from the onboarding screen. It would never dispatch the action that sets `walletExists` to `true` and would get hung up on the creating wallet rocket animation until restarting the app where everything would load as expected. This PR fixes the issue by no longer pre-defining the default wallet (`wallet0`) in the wallet reducer.

This PR:
- Added check to `getSelectedAddressType` method in `src/utils/wallet/index.ts`.
- No longer pre-define the `defaultWalletShape` for `wallet0` in `defaultWalletStoreShape`'s `wallets` object.


limpbrains (approved):

pwltr (changes requested):
Getting an error on fresh app startup on iOS at:
`hooks/wallet.ts` line 51: "Cannot convert undefined value to object"
`transactions` seems to be undefined now on first startup, causing this error.

coreyphillips:
> Getting an error on fresh app startup on iOS at: `hooks/wallet.ts` line 51: "Cannot convert undefined value to object" `transactions` seems to be undefined now on first startup, causing this error.

@pwltr Fixed [here](https://github.com/synonymdev/bitkit/pull/282/commits/55605699318067338541096e7237537a21bcc261).

pwltr:
Sorry, there are a couple other related errors as well which I missed before.

|1|2|3|
|--------------|-----------|------|
|Simulator Screen Shot - iPhone 13 - 2022-08-25 at 15 57 14|Simulator Screen Shot - iPhone 13 - 2022-08-25 at 15 57 20|Simulator Screen Shot - iPhone 13 - 2022-08-25 at 15 57 23

coreyphillips:
> Sorry, there are a couple other related errors as well which I missed before.
> 
> 1	2	3
> Simulator Screen Shot - iPhone 13 - 2022-08-25 at 15 57 14	Simulator Screen Shot - iPhone 13 - 2022-08-25 at 15 57 20	Simulator Screen Shot - iPhone 13 - 2022-08-25 at 15 57 23

I'm not able to reproduce these errors on my end. To ensure you're testing with a clean store can you reset your simulator by navigating to "Device->Erase All Content and Settings"? And start the app fresh from there once it's reset?

pwltr:
Okay so when I do that it runs fine. But what I did was uninstall the app on the simulator and then rebuild the app with `yarn ios`. That gives me the errors again. Not sure if that will be a problem in production. (would be still be quite annoying in development though)

coreyphillips:
> Okay so when I do that it runs fine. But what I did was uninstall the app on the simulator and then rebuild the app with `yarn ios`. That gives me the errors again. Not sure if that will be a problem in production. (would be still be quite annoying in development though)

Thanks, I'm able to reproduce it now. Fixed it [here](https://github.com/synonymdev/bitkit/pull/282/commits/c80208f5abd72db1b40332a4e904356773a4c555).
In the future a check on-launch will be added to ensure that the store structure hasn't changed and if it has missing a variable or object, we'll update the store to add it so the app isn't trying to access a place in storage that doesn't exist.

pwltr (approved):

PR #283, opened 2022-08-25
fix(send): remove old screens, optimise components with memo()

limpbrains:
Thanks to React DevTools Profiler I found why onchain numberpad was sluggish. It was mostly because of old components from Send flow was mounted and re-rendered as redux state was changed.
So I removed all unnecessarily components. kept a few related to `screens/Blocktanks`, we will eventually remove them as well.
Also I wrapped some navigational components in `memo()`, that reduces renders significantly. 

On my dev setup lag dropped from 300ms to 50, I hope on production it would be much better.


pwltr (approved):
Nice, no more input lag here. Validation and rest of the app seems to work as normal.

https://user-images.githubusercontent.com/8538369/186661722-c5850a77-949d-47e7-a650-5f4a4638a318.mp4

coreyphillips (approved):

limpbrains:
ready to be merged

MiguelMedeiros (approved):

PR #285, opened 2022-08-26
feat(lightning): Blocktank Purchase Testing

coreyphillips:
This PR:
- Adds a basic auto-buy button to the Channels component in order to test opening regtest channels with Blocktank.
  - Requires at least `2000` regtest sats for testing at the moment.
  - Zero-conf channels do not seem to work on regtest at the moment. Once the app pays for the channel and begins to check the order status, ensure that a block is mined to continue.
- Updates `@synonymdev/blocktank-client` to `0.0.47`.
- Separated blocktank methods between `actions/blocktank` & `utils/blocktank`.

### DIFF
--- a/src/store/actions/blocktank.ts
+++ b/src/store/actions/blocktank.ts
@@ -1,70 +1,229 @@
 import actions from './actions';
 import { err, ok, Result } from '@synonymdev/result';
-import { getDispatch } from '../helpers';
-import bt, {
+import { getDispatch, getStore } from '../helpers';
+import {
 	IBuyChannelRequest,
 	IBuyChannelResponse,
+	IFinalizeChannelResponse,
 } from '@synonymdev/blocktank-client';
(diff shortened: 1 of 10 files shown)


src/store/actions/blocktank.ts
```diff
@@ -1,70 +1,229 @@
...
 
-export const refreshServiceList = (): Promise<Result<string>> => {
-	return new Promise(async (resolve) => {
-		try {
-			const res = await bt.getInfo();
+/**
+ * Refreshes available services from BLocktank.
+ * @returns {Promise<Result<string>>}
+ */
+export const refreshServiceList = async (): Promise<Result<string>> => {
```
Jasonvdb, line 31: I think going forward we don't need to keep the service list in memory all the time after channel is purchased. It only needs to be up to date right before purchasing and then again in a few months when they buy a new channel.

Jasonvdb (approved):

PR #286, opened 2022-08-26
feat(wallet): Generate New Addresses

coreyphillips:
This PR:
- Adds and implements the `generateNewReceiveAddress` method when creating Bitcoin invoices.
- Adds `lastUsedAddressIndex` & `lastUsedChangeAddressIndex` to wallet reducer to track the gap limit as new addresses are generated.
  - The gap limit is only enforced when attempting to generate new addresses and is not yet implemented when scanning addresses for balances.

### DIFF
--- a/src/screens/Wallets/Receive/index.tsx
+++ b/src/screens/Wallets/Receive/index.tsx
@@ -132,7 +132,6 @@ const Receive = ({ navigation }): ReactElement => {
 
 	useEffect(() => {
 		setInvoiceDetails().then();
-		getLightningInvoice().then();
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [amount, message, selectedNetwork, selectedWallet]);
 
--- a/src/screens/Wallets/Receive/index.tsx
+++ b/src/screens/Wallets/Receive/index.tsx
@@ -6,9 +6,15 @@ import React, {
 	useEffect,
 	useRef,
 	MutableRefObject,
+	useCallback,
 } from 'react';
 import { useSelector } from 'react-redux';
-import { StyleSheet, useWindowDimensions, View } from 'react-native';
+import {
+	ActivityIndicator,
+	StyleSheet,
+	useWindowDimensions,
+	View,
+} from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import QRCode from 'react-native-qrcode-svg';
 import { FadeIn, FadeOut } from 'react-native-reanimated';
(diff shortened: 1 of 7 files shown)


src/screens/Wallets/Receive/index.tsx
```diff
@@ -40,76 +48,109 @@ const Receive = ({ navigation }): ReactElement => {
...
-	}, [amount, message]);
+	useEffect(() => {
+		resetInvoice();
+	}, []);
+
+	useEffect(() => {
+		setInvoiceDetails().then();
+		getLightningInvoice().then();
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [amount, message, selectedNetwork, selectedWallet]);
```
pwltr: `getLightningInvoice()` is called twice here on mount, once in `setInvoiceDetails()`. Also the `.then()` doesn't do anything here and should be omitted.

src/utils/wallet/index.ts
```diff
@@ -2539,3 +2540,56 @@ export const getBalance = ({
...
+			selectedWallet = getSelectedWallet();
+		}
+		if (!addressType) {
+			addressType = getSelectedAddressType({ selectedNetwork, selectedWallet });
+		}
+		const { currentWallet } = getCurrentWallet({
+			selectedWallet,
+			selectedNetwork,
+		});
+		const addressIndex = currentWallet.addressIndex[addressType].index;
```
pwltr: This throws an error for me, leading to no new address being generated: `currentWallet.addressIndex[addressType]` is undefined.

pwltr (changes requested):

src/screens/Wallets/Receive/index.tsx
coreyphillips: Fixed [here](https://github.com/synonymdev/bitkit/pull/286/commits/d0471e287bf139c0d9502c90790aa4f835fa6234).

The `.then()` is to acknowledge that a Promise is being returned and to remove the warning from certain IDE's.
Screen Shot 2022-08-29 at 11 59 39 AM

src/utils/wallet/index.ts
coreyphillips: Nice. Fixed [here](https://github.com/synonymdev/bitkit/pull/286/commits/5d4fa5fafcd1f82689e8696f82a11466ae8a1a48).

src/screens/Wallets/Receive/index.tsx
pwltr: Interesting, never seen this but seems useful. Shame that ESLint doesn't have a rule for this.

pwltr (approved):

PR #291, opened 2022-08-29
fix: swipe-to-go-back gesture

### LINKED ISSUE #289: App hangs on back gesture on iOS
**Describe the bug**
Go to wallet details page, try go back by pulling the screen corner, app hangs

**To Reproduce**
1. Go to wallet details page
2. try go back by pulling the screen corner
3. app hangs

**Expected behavior**
app works

 - OS: iOS

limpbrains:
I’ve found a reason why app hangs on iOS when you swipe-to-go-back.
It’s because of using @react-navigation/stack instead of @react-navigation/stack-native  for RootNavigation.
There is a [workaround](https://github.com/react-navigation/react-navigation/issues/9015#issuecomment-828700138) to fix that. The downside is it may lead to memory issues (unused screens will not be detached from view hierarchy)
I think we should try it to keep nice animations for Profile screen.

closes #289

### DIFF
--- a/src/navigation/root/RootNavigator.tsx
+++ b/src/navigation/root/RootNavigator.tsx
@@ -108,6 +108,10 @@ const RootNavigator = (): ReactElement => {
 		<NavigationContainer ref={navigationRef}>
 			<Stack.Navigator
 				screenOptions={navOptions}
+				// adding this because we are using @react-navigation/stack instead of
+				// @react-navigation/native-stack header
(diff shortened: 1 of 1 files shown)


pwltr (approved):

Jasonvdb (approved):

Jasonvdb:
Thanks! Made an issue [here](https://github.com/synonymdev/bitkit/issues/293) for later

PR #294, opened 2022-08-29
fix: upgrade Blur to v4 and enable it for Android

limpbrains:
BlurView v4 now works fine with Bitkit on Android, except Scanner screen.
But it for some reason it can't be use as a container, layout breaks, that is why I use it with position: absolute, just to create mask effect


limpbrains:
Doesn't work on `WalletsDetail` either, reverting

pwltr (approved):

PR #300, opened 2022-08-30
Slashtags SDK Path Fix

coreyphillips:
This PR:
- Appends full path to the Slashtag SDK's purpose portion of the derivation path.
- Updates expected result in Slashtags test.

### DIFF
--- a/src/utils/wallet/index.ts
+++ b/src/utils/wallet/index.ts
@@ -84,6 +84,7 @@ import {
 	GENERATE_ADDRESS_AMOUNT,
 } from './constants';
 import { moveMetaIncTxTags } from '../../store/actions/metadata';
+import { showErrorNotification } from '../notifications';
 
 export const refreshWallet = async ({
 	onchain = true,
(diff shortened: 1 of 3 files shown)


src/utils/wallet/index.ts
```diff
@@ -371,18 +372,25 @@ export const slashtagsPrimaryKey = async (seed: Buffer): Promise<string> => {
 	const network = networks.bitcoin;
 	const root = bip32.fromSeed(seed, network);
 
-	const path = SDK.DERIVATION_PATH;
+	const path = `${SDK.DERIVATION_PATH}/0'/0'/0/0`;
```
Nuhvi, line 375: Why is this needed? there isn't really any need for multiple different paths for the same purpose. 

If you have a good reason why it should be done this way, then might as well change the `SDK.DERIVATION_PATH` and update the [spec](https://github.com/synonymdev/slashtags/blob/master/specs/slashtags-key-derivation.md#from-seed-to-primarykey) for that.

Nuhvi: @coreyphillips You can argue of course that bip32 is overkill in that case and we might as well just hash the seed with a prefix like `hash(Buffer.from('slashtags'), seed)` to get a `primary key` ... and I wouldn't argue with that if you think that is the best option for wallet developers.

My goals are:
1- Neither the SDK nor any utility function I provide requires (as an input) the `Seed` or any private keys that might contain bitcoin or will ever derive a bitcoin holding private keys from it. So no need to ever audit my code to know I can't steal funds.
2- The primary key is easy and natural to derive from a seed for any wallet developer without adding any more dependencies than they already using (excluding the SDK)

Nuhvi: @coreyphillips here

PR #301, opened 2022-08-31
feat: scan qr-codes on send screen

limpbrains:
I moved scanner UI to separate component, so that it can be easily reused in different screens


src/screens/Wallets/SendOnChainTransaction/Scanner.tsx
```diff
@@ -0,0 +1,44 @@
...
+		if (res.isErr() || (res.isOk() && res.value.length === 0)) {
+			showErrorNotification({
+				title: 'QR code',
+				message: 'Sorry. Bitkit can’t read this QR code.',
+			});
+			return;
+		}
+
+		navigation.pop();
+		onScan(res.value[0]);
```
coreyphillips: We'll want to return the entire array of scanned items. I would suggest updating this to `onScan(res.value);`

limpbrains: This is for onchan send, currently we can't send-to-many, so we need to pick one address from qrcode.
I've improved this logic by filtering for `EQRDataType.bitcoinAddress` [here](https://github.com/synonymdev/bitkit/pull/301/commits/e20bb899ec69834e6c04d4841c7b2137f51a3f08)

coreyphillips (approved):

MiguelMedeiros (approved):

PR #303, opened 2022-08-31
fix(ui): close bottomSheet on hardwareBackPress

### LINKED ISSUE #292: Bottom Sheet does not close on back navigation
**Describe the bug**
When the BottomSheet is open, swiping back from the edge of the screen or pressing software/physical back button of the phone does not close the BottomSheet but the app itself.

**To Reproduce**
1. Go to 'Send'
2. Navigate back like described
3. App closes

**Expected behavior**
BottomSheet should close not app.

 - OS: iOS & Android

pwltr:
closes https://github.com/synonymdev/bitkit/issues/292

This solution registers a hook on the first screen of the navigation stack of the respective BottomSheet, or if there is no nested stack directly on the BottomSheet. Then we close it via redux action and unsubscribe. Very manual but it should work.

### Changes
- added `hardwareBackPress` handler hook to close `BottomSheet`s

### Preview

https://user-images.githubusercontent.com/8538369/187926830-ce36c166-2833-4f0d-b820-e8279ce691aa.mov

### DIFF
--- a/src/hooks/bottomSheet.ts
+++ b/src/hooks/bottomSheet.ts
@@ -1,0 +1,34 @@
+import { useEffect, useRef } from 'react';
+import { BackHandler, NativeEventSubscription } from 'react-native';
+import { toggleView } from '../store/actions/user';
+import { TViewController } from '../store/types/user';
+import { useAppSelector } from './redux';
+
+export const useBottomSheetBackPress = (viewController: TViewController) => {
+	const isBottomSheetOpen = useAppSelector(
+		(store) => store.user.viewController[viewController].isOpen,
+	);
+
+	const backHandlerSubscriptionRef = useRef<NativeEventSubscription | null>(
+		null,
+	);
+
+	useEffect(() => {
+		if (!isBottomSheetOpen) {
+			return;
+		}
+
+		backHandlerSubscriptionRef.current = BackHandler.addEventListener(
+			'hardwareBackPress',
+			() => {
+				toggleView({ view: viewController, data: { isOpen: false } });
+				return true;
+			},
+		);
+
+		return () => {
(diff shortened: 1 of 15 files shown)


limpbrains:
I don't think it should work like this.
I think back button should trigger back action in navigation inside bottomSheet. and If back is not available, then close it.
Can we archive this?

limpbrains:
Maybe it can be archived with bottom-sheet + React Navigation integration? I haven't looked at it.
https://gorhom.github.io/react-native-bottom-sheet/react-navigation-integration

pwltr:
I agree, maybe we can do it by passing the navigator to the BottomSheet. I've seen the page you linked but there is no further documentation. The example link is broken, so no idea how to do the integration.

limpbrains:
I was able to build example from BottomSheet. It works exactly as we want it to.

[back.webm](https://user-images.githubusercontent.com/155891/187872068-393b2ec2-9d63-455d-8629-2498c9b76087.webm)

here is my fork with fixed dependencies
https://github.com/limpbrains/react-native-bottom-sheet

example located at `example/bare`

Do you want to give it a try?

pwltr:
Yes I also found the example by now and am working on it. :) The updated fork is helpful.

pwltr:
Had to go another route by adapting the previous solution. If you look closely at the video above it doesn't actually close the BottomSheet but the whole screen of the parent stack (that has the sheet in it). The integration example has nothing to resolve the actual problem.

This solution registers a hook on the first screen of the navigation stack of the respective BottomSheet, or if there is no nested stack directly on the BottomSheet. Then we close it via redux action and unsubscribe. Very manual but it should work.

limpbrains (approved):

PR #317, opened 2022-09-05
Pay Lightning Invoice

coreyphillips:
This PR:
- Updates the scanning & clipboard paste logic to handle and prioritize lightning invoices when able.
  - All scanned and most clipboard paste data can now be pushed through the `processInputData` method.
  - The new scanner components have been wired up accordingly.
  - This method will parse the information as before, but in the event it is a lightning or on-chain payment request it will ensure we have lighting-channels, enough funds to pay the lightning invoice and that the invoice is not expired. If not, it will attempt to determine if we have enough funds to pay the on-chain address. If not, it will simply forward the Bitcoin address to the send form.
- Adds `lightningInvoice` to `defaultBitcoinTransactionData`.
- Renamed several "OnChain" names and types to "Bitcoin".

### DIFF
--- a/src/utils/scanner.ts
+++ b/src/utils/scanner.ts
@@ -4,8 +4,18 @@
 
 import bip21 from 'bip21';
 import { err, ok, Result } from '@synonymdev/result';
-import { Alert } from 'react-native';
+import {
+	availableNetworks,
+	EAvailableNetworks,
+	networks,
+	TAvailableNetworks,
+} from './networks';
 import { address as bitcoinJSAddress } from 'bitcoinjs-lib';
+import {
+	getOnchainTransactionData,
+	parseOnChainPaymentRequest,
+} from './wallet/transactions';
+import { getStore } from '../store/helpers';
 import {
 	getLNURLParams,
 	LNURLAuthParams,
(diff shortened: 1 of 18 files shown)


src/utils/scanner.ts
```diff
@@ -193,63 +272,193 @@ export const decodeQRData = async (
...
+			const decodedInvoice = await decodeLightningInvoice({
+				paymentRequest: options.lightning,
+			});
+			if (decodedInvoice.isOk()) {
+				foundNetworksInQR.push({
+					qrDataType: EQRDataType.lightningPaymentRequest,
+					lightningPaymentRequest: options.lightning,
+					network: selectedNetwork,
+					sats: milliSatoshisToSatoshis(
+						decodedInvoice.value?.amount_milli_satoshis ?? 0,
```
Jasonvdb, line 284: A decoded ln invoice should return `amount_satoshis` instead of `amount_milli_satoshis` here. So Bitkit should not have to ever worry about milli conversions. Unless I only fixed that in a newer version of react-native-ldk?

coreyphillips: The currently used version of react-native-ldk (v0.0.51) in Bitkit still seems to use `amount_milli_satoshis`. Once upgraded, this will need to be switched over.

limpbrains (approved):

Jasonvdb (approved):

PR #324, opened 2022-09-05
fix(wallet): improve handling of offline exchange rate API

pwltr:
Task: https://app.asana.com/0/1189635884316057/1202838628848168/f

### Changes
- show a toast when exchange rate API is down
- split up `getDisplayValues` into Bitcoin and fiat so we can handle errors & fallbacks separately
- update placeholders for fiat value if the exchange rate is not known
- fix some lint errors

### Notes
Easiest way to test is to change the e.g. the port in `exchange-rate/index.ts`. 

### Preview
Toast:
Simulator Screen Shot - iPhone 13 - 2022-09-06 at 10 50 27

Receiving payment when API is down and no previous rate is known:
Simulator Screen Shot - iPhone 13 - 2022-09-05 at 15 27 44

Fallback UI when API is down and no previous rate is known:
Simulator Screen Shot - iPhone 13 - 2022-09-05 at 15 28 08

If sats amount is zero we don't need the placeholder (-):

https://user-images.githubusercontent.com/8538369/188476141-68543a08-b1aa-4999-810d-52807f6c0cbf.mp4

Typing here doesn't do anything now:
Simulator Screen Shot - iPhone 13 - 2022-09-05 at 16 15 48

Can still transact normally:
Simulator Screen Shot - iPhone 13 - 2022-09-05 at 16 17 41


limpbrains (changes requested):
Awesome job!
Do you keep last successful update time? It would be nice to show it in error popup. Something like: "Using rates from 12/12/2009"
Can you please write a unit test for `getDisplayValues` with unknown rate ?

limpbrains (approved):

src/utils/exchange-rate/index.ts
```diff
@@ -18,13 +31,29 @@ export const getExchangeRates = async (): Promise<Result<IExchangeRates>> => {
...
 		return ok(rates);
 	} catch (e) {
 		console.error(e);
+
+		if (lastUpdatedAt) {
+			const date = timeAgo(lastUpdatedAt);
+
+			showErrorNotification({
+				title: 'Blocktank FX API Error',
+				message: `Could not get exchange rate, using price from\n${date}`,
```
coreyphillips: Replace comma with period so it reads as follows:
`Could not get exchange rate. Using price from\n${date}`

coreyphillips (approved):

PR #355, opened 2022-09-08
feat: iOS only portrait mode

limpbrains:
Allow only Portrait mode on iOS.
Looks like for android it is already enabled.


pwltr (approved):

pwltr:
iOS:
Simulator Screen Shot - iPhone 13 - 2022-09-09 at 11 34 47

Android:
Screenshot_1662716775

coreyphillips (approved):

Issue #318, opened 2022-09-05
UI: update all user facing notifications (toasts copy)


pwltr:
**Describe the bug**
Currently technical (error) messages are shown to the user via toasts in some places.

**Expected behavior**
Show a non-technical description of the error/event and if applicable a suggestion for the user to resolve the error (retry, restart app etc.)

PR #369, opened 2022-09-13
fix(lightning): Add Scanning Notification

coreyphillips:
This PR:
- Adds notification in the event a user scans a lightning invoice without enough funds.

### DIFF
--- a/src/utils/scanner.ts
+++ b/src/utils/scanner.ts
@@ -157,6 +157,10 @@ export const processInputData = async ({
 				selectedNetwork,
 			});
 			if (processBitcoinTxResponse.isErr()) {
+				showErrorNotification({
+					title: 'Unable To Pay Invoice',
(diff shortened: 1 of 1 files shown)


src/utils/scanner.ts
```diff
@@ -157,6 +157,10 @@ export const processInputData = async ({
 				selectedNetwork,
 			});
 			if (processBitcoinTxResponse.isErr()) {
+				showErrorNotification({
+					title: 'Unable To Pay Invoice',
+					message: processBitcoinTxResponse.error.message,
```
pwltr, line 162: We shouldn't put technical error messages in the toasts as they are user facing. I suggest we list all of these with some context in https://github.com/synonymdev/bitkit/issues/318 so we can change the copy together with Aldert later.

pwltr (approved):

PR #388, opened 2022-09-16
fix(ui): only hide balance where it makes sense

### LINKED ISSUE #376: Send/Receive: Hiding Balance also makes send/receive flow unusable
**Describe the bug**
After hiding balance on Wallet screen the amounts in send & receive flows are also hidden, making them unusable.

**To Reproduce**
1. Go to Wallet/Home screen
2. Swipe to the side on the balance
3. Go to Send or Receive flow
4. Try to use UI

**Expected behavior**
Amounts are not hidden, only the user's balance and tx history. 

**Screenshots**
Simulator Screen Shot - iPhone 13 - 2022-09-16 at 11 18 30

pwltr:
Closes https://github.com/synonymdev/bitkit/issues/376

### Changes
- set all `<Money />` components to use the correct `enableHide` prop 
- add ability to hide/show balance in `WalletsDetail` screen

### DIFF
--- a/src/components/AssetCard.tsx
+++ b/src/components/AssetCard.tsx
@@ -29,13 +29,13 @@ const AssetCard = ({
 			<View color="transparent" style={styles.col2}>
 				<Money
 					sats={satoshis}
-					hide={true}
+					enableHide={true}
 					size="text01m"
 					style={styles.value}
 				/>
 				<Money
 					sats={satoshis}
-					hide={true}
+					enableHide={true}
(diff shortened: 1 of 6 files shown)


pwltr:
@limpbrains 
> also (always) hide the balance during the receive flow, especially once a QR is presented

Didn't change anything for this as the BottomSheet always overlaps the balance in the background.

limpbrains (approved):

PR #406, opened 2022-09-21
feat(lightning): Implemented Custom Channels

coreyphillips:
This PR:
 - Adds the ability to purchase Blocktank channels via the Custom Channels flow.

To Test:
1. Ensure the app has around at least 20,000-100,000 sats.
2. Select the "Get on Lightning" suggestion on the Home screen or navigate to:
   - Settings->Networks->Lightning Connections->Add New Connection
3. Select "Custom Setup".
4. Keep "0" selected as your spending balance and press "Continue".
5. Keep the pre-set receiving amount and press "Continue".
6. Swipe to pay on the confirm screen.
7. Once confirmed, mine a block or two and wait for Blocktank to see the transaction and open the channel.
    - If the channel doesn't open after some time attempt to refresh the app or mine another block.

Notes:
- Attempting to set the max amount when specifying spending or receiving channel balances will not always work yet. We will need to implement the calculation Blocktank uses to calculate channel costs to more accurately set these values. Right now it does its best to calculate what the user can and cannot afford.

### DIFF
--- a/src/screens/Wallets/SendOnChainTransaction/AddressAndAmount.tsx
+++ b/src/screens/Wallets/SendOnChainTransaction/AddressAndAmount.tsx
@@ -128,9 +128,7 @@ const AddressAndAmount = ({ index = 0, navigation }): ReactElement => {
 	 * Returns the current output by index.
 	 */
 	const getOutput = useMemo((): IOutput | undefined => {
-		try {
-			return transaction.outputs?.[index];
-		} catch {
+		return transaction.outputs?.[index] ?? { address: '', value: 0, index: 0 };
 			return { address: '', value: 0, index: 0 };
 		}
 	}, [index, transaction?.outputs]);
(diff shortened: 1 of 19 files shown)


limpbrains (approved):

src/screens/Wallets/SendOnChainTransaction/AddressAndAmount.tsx
```diff
@@ -131,7 +131,7 @@ const AddressAndAmount = ({ index = 0, navigation }): ReactElement => {
 		try {
 			return transaction.outputs?.[index];
 		} catch {
```
pwltr: This code does not what it's intending to do. If the `index` can not be found in the `outputs` it will return undefined. There is nothing here that can throw so instead of the `try...catch` should be:

```suggestion
		return transaction.outputs?.[index] ?? { address: '', value: 0, index: 0 };
```

There are also similiar occurrences in this file and others in f.e. `ReviewAndSend.tsx`

pwltr (changes requested):

coreyphillips:
@pwltr, nice catch. Updated [here](https://github.com/synonymdev/bitkit/pull/406/commits/4088ba748952d457a1671bc015df10eb257b1140).

pwltr (approved):

PR #411, opened 2022-09-23
feat(lightning): Pull Blocktank Node Info From API

coreyphillips:
This PR:
- We no longer rely on the hardcoded lightning peer list when adding peers. We now retrieve and save the current Blocktank peer info from Blocktank's `getInfo` method.
- Adds `info` to the `blocktank` reducer.
- Adds a `refreshBlocktankInfo` method that will refresh/update Blocktank's node info on app start.


Jasonvdb (approved):

src/store/actions/blocktank.ts
```diff
@@ -134,6 +139,20 @@ export const refreshOrder = async (
 	}
 };
 
+/**
+ * Refreshes Blocktank Node Info
+ * @returns {Promise<void>}
+ */
+export const refreshBlocktankInfo = async (): Promise<void> => {
+	const infoResponse = await getBlocktankInfo();
```
pwltr, line 147: Should handle errors here since this is a network request.

pwltr (changes requested):

src/store/actions/blocktank.ts
coreyphillips: This error will need to be handled by `@synonymdev/blocktank-client` and respond with `Promise<Result<IGetInfoResponse>>`. Until then, I've added a condition to check for a valid response prior to updating [here](https://github.com/synonymdev/bitkit/pull/411/commits/aa22d02f703499cb96c34d0bf3f16098afbbfec2).

pwltr (approved):

PR #412, opened 2022-09-23
fix(lightning): Channel Purchase Improvements

coreyphillips:
This PR:
- Adds on-chain transaction fees to the channel purchase cost in the `QuickConfirm` & `CustomConfirm` components.
- Updates the max spend/receive calculations in the `CustomSetup` and `QuickSetup` channel purchase flows.
- Fixes slider divide by zero bug.
- Updates lightning todo logic.
- Updated `ADD_TODO` reducer to add new todos to the front.

### DIFF
--- a/src/utils/todos/index.ts
+++ b/src/utils/todos/index.ts
@@ -5,6 +5,7 @@ import { addTodo, removeTodo } from '../../store/actions/todos';
 import { toggleView } from '../../store/actions/user';
 import { getOpenChannels } from '../lightning';
 import { TChannel } from '@synonymdev/react-native-ldk';
+import { getBalance } from '../wallet';
 
 type TTodoPresets = { [key in TTodoType]: ITodo };
 export const todoPresets: TTodoPresets = {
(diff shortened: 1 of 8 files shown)


src/utils/todos/index.ts
```diff
@@ -93,6 +94,7 @@ export const setupTodos = async (): Promise<void> => {
 	 * Check for lightning.
 	 */
 	const lightning = todos.some((todo) => todo.type === 'lightning');
+	const lightningIsDismissed = 'lightning' in dismissedTodos;
```
pwltr: I don't believe the `in` operator can be used like this on arrays.

```suggestion
	const lightningIsDismissed = dismissedTodos.includes('lightning');
```

pwltr (changes requested):

src/utils/todos/index.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/412/commits/5f0094d296b49e0489545d9b795243ece3742ae0). It appears several of the todos were using the faulty logic so I updated them accordingly. We may need to consider adding additional data to these dismissed todos in the future.

pwltr (approved):

limpbrains (approved):

PR #421, opened 2022-09-27
fix(profile): undefined is not a function when saving profile

Nuhvi:
Fix `undefined is not a function` should be working fine.
As for reconnection to WebSocket, I added a button to close the WebSocket manually to test it and it works but needs a real test in the app to see what happens when it disconnects on its own.


Nuhvi:
@Jasonvdb would be great if we can test this before friday!

Jasonvdb (approved):

Jasonvdb:
Sure I'll push later today

PR #423, opened 2022-09-28
feat: send to slashtag url

limpbrains:
Because the way we handle slashtag url now depends on the context (if you scan it from main screen, contact will be open, if you scan it from send form, we will try to get slash pay and send to it) processInputData now accepts new parameter - `source`. It also now accepts slashtagssdk, so we can resolve slashpay of remote profiles.

All other places where slashtags is used are now also use `processInputData`, this allows unified workflow

https://user-images.githubusercontent.com/155891/192774316-ee19954e-71d4-4976-8e0b-cf38514b340a.mp4

### DIFF
--- a/src/screens/Contacts/Contact.tsx
+++ b/src/screens/Contacts/Contact.tsx
@@ -1,49 +1,67 @@
-import React, { useCallback } from 'react';
-import { StyleSheet, Share, Alert } from 'react-native';
+import React, { useCallback, useMemo } from 'react';
+import { StyleSheet, Share } from 'react-native';
+import { useSelector } from 'react-redux';
 
 import {
 	CoinsIcon,
 	PencileIcon,
 	ShareIcon,
+	TouchableOpacity,
 	TrashIcon,
 	View,
 } from '../../styles/components';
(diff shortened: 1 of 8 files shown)


Nuhvi:
Interesting, probably a better solution than what I was considering (creating a navigator for Slashtags that sends you where to the apropriate screen) but this applies to slashpay too.

Will try to fork this to add Feeds and Auth as well.

limpbrains:
As I understand, Auth and Feeds with have different prefix, ike `slashauth:...` , right? These urls should be handled in `processInputData` and user should be redirected to new Screens that can handle them

pwltr (approved):

Nuhvi:
@limpbrains Paying to a contact with no valid address adds the contact in the `to` input field, but no send button obviously, but I wonder if this is the intended UX? how to tell the user that there is no way to pay that person?

limpbrains:
> @limpbrains Paying to a contact with no valid address adds the contact in the `to` input field, but no send button obviously, but I wonder if this is the intended UX? how to tell the user that there is no way to pay that person?

fixed that, now it will show an error

PR #427, opened 2022-09-29
feat(wallet): Bundle Boosted Transactions

coreyphillips:
This PR:
- Updates `boostedtransactions` shape, hides previously boosted transactions from the list view and accumulates the total/updated value from all boosts.
- Lists parent transactions of a boosted transaction in the transaction details view.
- Updates rbf/cpfp logic.
- Adds additional checks where needed.

Notes:
After every boost transaction we now store the following data:
```
[key: string]: { // The key is the txid
  parentTransactions: string[]; // An array of parent txids of the current txid.
  childTransaction: string; // Child txid of the current boosted txid.
  type: EBoost; // Whether it was a rbf or cpfp tx.
  fee: number; // How much the fee for the boost was.
};
```
This allows us to track boosted transactions however, we wish moving forward and is currently used to filter out previously boosted transactions from the activity list and calculate the fee accordingly.

The additional checks added in this PR are to ensure the method can handle a scenario where a transaction was removed from the mempool mid-search.

In the future we will want to create a PR to further optimize the activity list to no longer require entry updates/modifications once a tx is confirmed.

Simulator Screen Shot - iPhone 13 - 2022-09-29 at 12 21 18


limpbrains:
If I boost transaction from Activity detail, it shows me Already bootsted, but if I go to main screen and come back, I can boost it again. Is it expected behavior?

src/screens/Activity/ListItem.tsx
```diff
@@ -50,15 +60,22 @@ const ListItem = ({
...
 		}
 		if (activityType !== EActivityTypes.onChain) {
 			return false;
 		}
+		if (isBoosted) {
+			return false;
+		}
 		return canBoost(id).canBoost;
-	}, [confirmed, activityType, id]);
+	}, [confirmed, activityType, isBoosted, id]);
```
pwltr, line 73: Can remove the logic (and the button) for boost here as per design prototype v25

pwltr (changes requested):

src/screens/Activity/ListItem.tsx
coreyphillips: The decision was to keep the boost buttons on the ListView.

coreyphillips:
> If I boost transaction from Activity detail, it shows me Already bootsted, but if I go to main screen and come back, I can boost it again. Is it expected behavior?

This is happening since the Activity Detail is still referencing the old id from which it was boosted. We'll need to find a way to update that successfully after a boost, but that may need to get tackled in a separate PR. I'll create a separate issue for it.

limpbrains (approved):

pwltr (approved):

PR #432, opened 2022-09-30
feat(wallet): Handle OS Paste

coreyphillips:
This PR:
- Adds check to `onChangeText` in `AddressAndAmount.tsx` for invoice keywords from a potential OS paste.

### DIFF
--- a/src/screens/Wallets/SendOnChainTransaction/AddressAndAmount.tsx
+++ b/src/screens/Wallets/SendOnChainTransaction/AddressAndAmount.tsx
@@ -83,6 +83,7 @@ const AddressAndAmount = ({ index = 0, navigation }): ReactElement => {
 	const [decodedInvoice, setDecodedInvoice] = useState<undefined | TInvoice>(
 		undefined,
 	);
+	const [handledOsPaste, setHandledOsPaste] = useState(false);
 	const transaction = useTransactionDetails();
 	const sdk = useSlashtagsSDK();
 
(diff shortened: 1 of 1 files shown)


limpbrains:
What about onBlur? I think it should work the same way

coreyphillips:
> What about onBlur? I think it should work the same way

I tried to get it to mimic the paste button, but it depends on if we want to handle an OS paste immediately or require the user to tap out of the input for the app to handle it. We could forward it to Aldert to see which experience he may prefer.

aldertnl:
I would say we should avoid additional taps that are not essential user input required to complete a goal / flow. Therefore, immediate handling seems like the best solution (in most cases, see below for details). Also, if we handle pasting immediately, it will help users perceive the app as being 'snappy and smart'. A user can always hit the back button to change their input in case they pasted a wrong invoice or something.

- Pasting an invoice with amount > 0 can take the user immediately to the transaction review screen.
- If a user already entered an amount before pasting and address or choosing a contact, we can forward them to the transaction review screen as well (after pasting).
- Pasting anything else should probably keep the user in the same view, allowing them to enter an amount, before continuing.

coreyphillips:
Sounds great. The current functionality of this PR will process the pasted data immediately, but not forward them to the next screen for review. That logic can be handled/added in a separate PR. I created a feature request for it [here](https://github.com/synonymdev/bitkit/issues/441).

pwltr (approved):

PR #437, opened 2022-10-02
feat: add ForgotPIN and HighBalanceWarning BottomSheets

### LINKED ISSUE #429: Wallet: Add "High balance" warning
### Tasks
- add the "High balance" warning from design prototype

### LINKED ISSUE #430: PIN: add "Forgot PIN?" BottomSheet
### Tasks
- add the "Forgot PIN?" BottomSheet from design prototype

pwltr:
Closes https://github.com/synonymdev/bitkit/issues/430 https://github.com/synonymdev/bitkit/issues/429

### Changes
- add HighBalanceWarning bottomSheet
- add ForgotPIN bottomSheet
- add link to tx details on Send result screen
- add toggle to require PIN for payments on PIN setup result screen
- enable hiding the wallet onboarding message
- fix confetti animation for newTxPrompt
- fix bugs introduced in my last PR for BackupPrompt bottomSheet

### Notes
We now close all bottomSheets & modals when the App launches, in case the user closed the app while a bottomSheet was open (the redux state would be out of sync then).

### High Balance Warning
works as follows:
- minium amount 1000$ or 5000000 sats (if exchange rate not available)
- max. 3 times
- max once per day
- user has to be on Wallets screen for 3 seconds


limpbrains:
Currently, if you enter pin incorrectly in 10th attempt, wallet will be wiped. I think this we don't need this, if we have Forgot PIN bottomsheet

pwltr:
> Currently, if you enter pin incorrectly in 10th attempt, wallet will be wiped. I think this we don't need this, if we have Forgot PIN bottomsheet

I think this is useful to have in case someone else took your phone and tries to guess the PIN. We can bring this up in Slack and remove it in another PR.

limpbrains (approved):

PR #450, opened 2022-10-05
fix(profile): save profile data all at once

### LINKED ISSUE #449: New Profile saving UX causes race condition and corrupting the Hyperdrive
**Describe the bug**
 saving is async, which means you can end up with the `ProfileAddLink`, `deleteLink`, and the `save` function for the rest of the profile, all racing to append to the Hypercore, all trying to add Block X+1 to the Merkle tree, thus corrupting the hyperdrive!!
 
**To Reproduce**

1. Go to Profile Edit screen
2. Edit the name, image and/or Bio
3. Add a link
4. (optional) Remove a Link
5. save profile edits
6. refresh the app (it may not resolve correctly)
7. try to edit the profile again (it should show an error notification)

**Expected behavior**
It should either save the profile only once after the user decides all changes or each of the saving functions should somehow wait the others to finish!

**Screenshots**

https://user-images.githubusercontent.com/40009100/194096397-e3aeddff-c935-4145-af72-4badd18a7f6c.mov

pwltr:
Closes https://github.com/synonymdev/bitkit/issues/449

### Changes
- move all profile link data to Redux until user is ready to save to the remote
- add uuid to links to enable multiple links with the same label


src/store/reducers/slashtags.ts
```diff
@@ -27,6 +27,37 @@ const slashtags = (
...
+		case actions.ADD_LINK: {
+			// currently we don't support multiple links with the same title
+			const prevLinks = state.links.filter(
+				(link) => link.title !== action.payload.title,
+			);
+			return {
+				...state,
+				links: [...prevLinks, action.payload],
+			};
+		}
```
pwltr: How should we handle the case where the user adds a link with an existing title? Should we add uuids to allow multiple links with the same title? Or overwrite them as is currently the case?

src/components/ProfileLinks.tsx
```diff
@@ -10,22 +10,19 @@ import { Caption13Up, Text02S, TrashIcon } from '../styles/components';
...
-	setLink,
-	deleteLink,
 }: {
 	links?: BasicProfile['links'];
+	editable?: boolean;
 	style?: StyleProp<ViewStyle>;
-	setLink?: (title: string, url: string | undefined) => void;
-	deleteLink?: (title: string) => void;
 }): JSX.Element => {
 	links = links?.filter(({ url }) => url?.length > 0) ?? [];
```
pwltr: Can we remove this line? Currently when editing a link it disappears when fully clearing the text.

src/store/reducers/slashtags.ts
Nuhvi: Hmmm just checked Linktree... seems that it allows multiple links of the same type (label), also, we can't stop people from doing the same, since the underlying data structure is an array!

If uuid works for you and makes your life easier, go with it.

src/components/ProfileLinks.tsx
Nuhvi: Makes sense, maybe instead saving the link should be disabled if the url is empty.

limpbrains (approved):

PR #453, opened 2022-10-05
feat: freeze package versions in package.json

limpbrains:
To make dependencies maintaining more predictable and robust I suggest we freeze all versions in package.json 

I tried to to upgrade any dependency and transfer it's current version from yarn.lock to package.json


limpbrains:
For some reason yarn started to install packages into subdirectories and it breaks eslint.
before
`node_modules/eslint-plugin-ft-flow`
after
`node_modules/@react-native-community/eslint-config/node_modules/eslint-plugin-ft-flow`
I will leave it for now

PR #493, opened 2022-10-12
fix: Contacts

### LINKED ISSUE #378: Disallow adding your local slashtags as contacts
**Describe the bug**
It is possible to add your self as contact, but not delete that contact later.

**To Reproduce**
1. Copy your own slashtag url and add it as a contact.
2. Try to delete that account.

**Expected behavior**
Check if the url is already one of your local slashtags (easiest way is to check the profiles in the SlashtagsProvider), and if so, just navigate to that profile page instead of adding it as a contact.

**Screenshots**

Screenshot_20220916-133936_Bitkit

 - OS:  Android

### LINKED ISSUE #461: Don't Hide save contact while resolving
The latest change to ContactEdit https://github.com/synonymdev/bitkit/commit/c1b12239b5ddfe49ce6cc771dd58309077def77d makes the save contact unreachable until the profile resolved

https://github.com/synonymdev/bitkit/blob/master/src/screens/Contacts/ContactEdit.tsx#L100

Basically, I need to be able to save the contact even if it is taking too long to fetch!

pwltr:
Closes https://github.com/synonymdev/bitkit/issues/378 https://github.com/synonymdev/bitkit/issues/461

### Changes
- preventing adding own slashtag
- don't hide save button when adding contact
- update stpg "Create Contact" links

### Preview
Simulator Screen Shot - iPhone 13 - 2022-10-12 at 18 43 09

### DIFF
--- a/src/components/ProfileCard.tsx
+++ b/src/components/ProfileCard.tsx
@@ -125,6 +125,7 @@ const styles = StyleSheet.create({
 	},
 	text: {
 		flex: 1,
+		paddingRight: 16,
 	},
 	name: {
 		flexWrap: 'wrap',
--- a/src/components/ProfileLinks.tsx
+++ b/src/components/ProfileLinks.tsx
@@ -7,7 +7,7 @@ import {
 	ViewStyle,
 } from 'react-native';
 import { Caption13Up, Text02S, TrashIcon } from '../styles/components';
-import { Link } from '../store/types/slashtags';
+import { LocalLink } from '../store/types/slashtags';
 import { openURL } from '../utils/helpers';
 import LabeledInput from './LabeledInput';
 import { editLink, removeLink } from '../store/actions/slashtags';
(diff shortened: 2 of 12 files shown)


Nuhvi (approved):
The problem with `uuid` is that you can't really depend on everyone to create a unique `id` field ...  why not just use `${title}:${value}` as the identifier instead?

That being said, I am not against adding `id` field to links specs, just wondering if Bitkit still needs to cope with missing `id` field since `title` + `value` can still serve this purpose.

pwltr:
@Nazeh Removed the `id` field from the remote profile but as you can probably see in https://github.com/synonymdev/bitkit/pull/493/commits/6848337663e057d970d890902a6eaf378fee7349 it is way less ergonomic to work with. I'd expect most apps using this links spec will run into similiar problems without a unique identifier so I would question if it wouldn't be better for us to promote good practice by just demanding a uuid field in the spec.

PR #504, opened 2022-10-13
feat: cache profiles in redux to keep rendering profiles synchronus

Nuhvi:
biting the bullet!


src/store/types/slashtags.ts
```diff
@@ -34,4 +34,12 @@ export interface ISlashtags {
 	seeder?: {
 		lastSent?: number;
 	};
+	// cache seen profiles!
+	profiles?: {
```
limpbrains, line 38: I think `profiles` in redux doesn't need to be optional.

limpbrains (approved):

src/store/types/slashtags.ts
Nuhvi: It will be missing in older Bitkit releases, so this helps other parts of the code handle that case without throwing!

PR #519, opened 2022-10-15
fix: various fixes

### LINKED ISSUE #495: Send/Receive: Confetti not animated on some phones
Some users are not seeing the confetti animated when sending (Result screen) or receiving (Payment Received modal) a payment.

pwltr:
Closes #495 

Mostly fixes from Bitkit build v21 findings Google Doc

### Changes
- fix(settings): auth protected routes
- fix(ui): #495 force lottie autoplay
- fix(settings): update RadioButtonGroup
- fix(assets): small fixes for BitcoinBreakdown
- fix(settings): update LightningNodeInfo
- fix lint some warnings


src/components/ProfileLinks.tsx
```diff
@@ -28,7 +28,8 @@ const ProfileLinks = ({
 			) : (
 				links.map((link): JSX.Element => {
 					const trimmedUrl = link.url
-						.replace('https://www.', '')
+						.replace('https://', '')
+						.replace('www.', '')
```
limpbrains, line 32: technically www.domain.com and domain.com may act very different. 
I wouldn't decide for the user

limpbrains (approved):

src/components/ProfileLinks.tsx
pwltr: This is just the display text to make it a bit shorter, the link itself is not changed. Or do you think we should not shorten the text?

limpbrains: No, I don't think so.
This is fine then

PR #527, opened 2022-10-18
fix: replace native-stack navigator with simple one to fix swipe-to-back gesture in settings menu

### LINKED ISSUE #526: Swipe back from settings screens sometimes throws you to main screen. iOS
On the 2nd or 3rd layers swipe to go back gesture sometimes works incorrectly. Instead of going back, it goes all the way to the top route. Feels like a bug in react-navigation. I've tried to upgrade to the latest version, but that didn't help.

iOS

https://user-images.githubusercontent.com/155891/196403032-46823ddd-e16f-432e-8a1e-4bd2e8f44bc2.mp4

limpbrains:
closes #526


pwltr:
Do we need to add this (from `RootNavigator.tsx`)?
```
// adding this because we are using @react-navigation/stack instead of
// @react-navigation/native-stack header
// https://github.com/react-navigation/react-navigation/issues/9015#issuecomment-828700138
detachInactiveScreens={false}
```

pwltr:
On Android I'm getting flashes of the Home screen when navigating:

https://user-images.githubusercontent.com/8538369/196420317-a47bb784-1dcd-4cb2-832d-d27a46c48e25.mov

limpbrains:
Looks like it is not a good solution, then

pwltr:
I just noticed the same flickering on current master too. Maybe it is unrelated to Navigator, don't really see it on a real phone. So you may want to reopen this.

limpbrains:
Let's try it!

pwltr (approved):

pwltr:
Flickering issue on Android seems to be caused by this, we might need to revert or find a separate solution for iOS.

PR #537, opened 2022-10-19
feat(send): auto parse clipboard payment data on App open

pwltr:
### Changes
- split up `App.tsx` into `AppOnboarded.tsx` that has a wallet and the `SlashtagsContext`
- add Settings option to `enableAutoReadClipboard` on App start (and when it comes back to the foreground)
- `readClipboardInvoice` to parse clipboard payment data and navigate to Send flow

### Notes
Only on-chain for now. Couldn't test on iOS because clipboard + simulator is weird on M1. 

### Preview

https://user-images.githubusercontent.com/8538369/196679368-4b7838c0-1ccc-4877-8724-42d8c2641bf7.mov

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,27 +10,19 @@ import React, {
 } from 'react';
 import { Platform, UIManager, NativeModules } from 'react-native';
 import { useSelector } from 'react-redux';
-import NetInfo from '@react-native-community/netinfo';
 import Toast from 'react-native-toast-message';
-
 import { ThemeProvider } from 'styled-components/native';
-import { SafeAreaProvider } from './styles/components';
-import { StatusBar } from './styles/components';
-import RootNavigator from './navigation/root/RootNavigator';
+
+import { SafeAreaProvider, StatusBar } from './styles/components';
 import Store from './store/types';
-import { closeAllViews, updateUser } from './store/actions/user';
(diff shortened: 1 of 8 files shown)


limpbrains (commented):
Can't we use same code for
main qrcode scanner, clipboard parsing, deeplinks ?

pwltr:
Having trouble reusing the scanner code tbh, I think we need to refactor/split up some parts there so it's more reusable. Or do you mean something else?

limpbrains:
I'm not sure about this feature, wouldn't it be too annoying?
Good thing that it is disabled by default

cc @coreyphillips

coreyphillips:
Ideally, the if condition [here](https://github.com/synonymdev/bitkit/pull/537/files#diff-b68464d3cd0381c84114f108920e13595b2d7b7e177a8589b6b2af55d53f3eaeR31) should be handled by `processInputData`, but as you mentioned we would first need to split out logic to create a method that returns a boolean on whether or not the clipboard data is recognizable. If it is, then we forward it to `processInputData` to handle it as needed. This implementation is fine for now, but we should mark this (splitting out a method to determine whether we can or cannot handle certain data or at least implement a way to silent notifications in `processInputData`) as a future TODO item.

coreyphillips (approved):

PR #538, opened 2022-10-19
fix: address generation on address type switch

limpbrains:
- fixing a typo `getReceiveAddress` func. It still can fail if addresses is empty. Should we run updateAddressIndexes there?
- run `updateAddressIndexes` on addressType change
- change default addressType


pwltr:
I still don't get an address. Cleaned emulator, went directly to Receive flow without changing any settings.

limpbrains:
Yeah, there is another bug in `updateAddressIndexes` :(

limpbrains:
fixed issue in `updateAddressIndexes` it didn't wait for all the promises to be resolved first.
this caused `refreshWallet` to exit without addresses in redux store.

I think bug is not completely resolved. There is a chance that you will tap on receive button too fast while there will be no generated addresses available. We should render UI only if we already have them. But this should be a separate PR.

@pwltr please, test again. I don't have this issue on android anymore

pwltr:
Issue still there for me on iOS and Android.

coreyphillips (approved):

PR #552, opened 2022-10-20
feat(wallet): Implements Geo Block

coreyphillips:
This PR:
- Implements a geo block to prevent US users from purchasing blocktank channels.
- Adds `isGeoBlocked` to user reducer.
- Sets geo block prior to setting up Blocktank.
- Updates "Instant Payments" view text and removes the setup buttons if geo blocked, as seen in the attached screenshot.
- Removes geo block for regtest.
Simulator Screen Shot - iPhone 13 - 2022-10-20 at 20 02 42

### DIFF
--- a/src/screens/Lightning/Introduction.tsx
+++ b/src/screens/Lightning/Introduction.tsx
@@ -1,4 +1,4 @@
-import React, { ReactElement } from 'react';
+import React, { ReactElement, useMemo } from 'react';
 import { Image, StyleSheet } from 'react-native';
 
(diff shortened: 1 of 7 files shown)


limpbrains (changes requested):
Same should be done for Rebalance, right?

coreyphillips:
Indeed! Good catch. Implemented [here](https://github.com/synonymdev/bitkit/pull/552/commits/efcdb35bc888a56dbe7f157d3f2549a1a6aad1d8).

limpbrains (approved):

PR #560, opened 2022-10-21
fix: settings flickerting

limpbrains:
@pwltr Do you think this can fix Flickering on android? #527


pwltr:
Yes we should try this, I will test in a bit.

PR #561, opened 2022-10-21
Remote backups and restoring

Jasonvdb:
- Automated LDK state backups
- Right now it's just restoring the LDK state as is but I would like it to auto sweep all funds from channels for safety. Will update react-native-ldk with that settings.
- There's a bit of a flash between the 2 restoring screens I need to fix still. The reason for 2 screens is because the first one restores the seed and the 2nd one is inside the slashtags provider which is needed to restore remote data. I'd like to circle back and still rethink this later as it seems a little too messy.
- Fixes ldk wipe directory and resets ldk before wiping else content is just written back right away by the ldk process.
- Replaced `funding_txo` with corrected `funding_txid` field


src/screens/Onboarding/RestoreFromSeed.tsx
```diff
@@ -121,18 +121,16 @@ const RestoreFromSeed = (): ReactElement => {
...
+		verifyBackup();
+
+		const res = await restoreSeed({ mnemonic: seed.join(' ') });
 		if (res.isErr()) {
-			setIsRestoringWallet(false);
-			showErrorNotification({
-				title: 'Error Restoring Wallet',
-				message: res.error.message,
-			});
+			Alert.alert(res.error.message);
```
pwltr: Should use the styled toast to show user facing error messages:

```suggestion
			showErrorNotification({
				title: 'Error Restoring Wallet',
				message: res.error.message,
			});
```

pwltr (changes requested):

src/screens/Onboarding/RestoreFromSeed.tsx
Jasonvdb: Thanks done!

coreyphillips (approved):

pwltr (approved):

PR #570, opened 2022-10-24
fix(wallet): Transaction Amount Check

coreyphillips:
This PR:
- Ensures a previously set transaction amount does not get overrided.

### DIFF
--- a/src/utils/scanner.ts
+++ b/src/utils/scanner.ts
@@ -17,6 +17,7 @@ import {
 
 import {
 	getOnchainTransactionData,
+	getTransactionOutputAmount,
 	parseOnChainPaymentRequest,
 } from './wallet/transactions';
 import { getStore } from '../store/helpers';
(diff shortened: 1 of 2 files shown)


pwltr:
Only thing I can see is that paymentRequest data with an amount of 0 is not updating the amount. But I wouldn't consider this an issue. Would you? As an aside we should update our receive unified URIs to not attach a `?amount=0` param if no amount is specified in the Receive flow.

Example URI:
`bitcoin:bcrt1qzz2uh3e33t68hmmhl8nurf3xel53wgnyc72new?amount=0`

pwltr (approved):

coreyphillips:
> Only thing I can see is that paymentRequest data with an amount of 0 is not updating the amount. But I wouldn't consider this an issue. Would you? As an aside we should update our receive unified URIs to not attach a `?amount=0` param if no amount is specified in the Receive flow.
> 
> Example URI: `bitcoin:bcrt1qzz2uh3e33t68hmmhl8nurf3xel53wgnyc72new?amount=0`

Agreed, I wouldn't consider that an issue. I added an issue regarding the Bitcoin URI [here](https://github.com/synonymdev/bitkit/issues/573).

PR #590, opened 2022-10-25
fix: reset activity store when switching network

limpbrains:


pwltr (approved):

coreyphillips:
Does resetting the activity store and switching back to the network with the previous activity repopulate the activities list?

coreyphillips:
As a future todo, we may want to split up the activity store by networks to resemble the other stores so we can toggle between them as we switch.

limpbrains:
> Does resetting the activity store and switching back to the network with the previous activity repopulate the activities list?

yes. There activity store will be repopulated on wallet refresh action

limpbrains:
> As a future todo, we may want to split up the activity store by networks to resemble the other stores so we can toggle between them as we switch.

agree

PR #601, opened 2022-10-26
fix(wallet): Hide Textinput

coreyphillips:
This PR:
- Hides textinput when number pad is displayed.
- Only displays amount when textinput is present.


pwltr (approved):
Took the freedom to tweak the styles a bit to make it look more like the prototype.

PR #606, opened 2022-10-26
fix(wallet): Hide Bitcoin Network Toggle

coreyphillips:
This PR:
- Only displays the Bitcoin network toggle when dev options are enabled.


pwltr (approved):
Has this been discussed? We don't want users to be able to switch?

coreyphillips:
> Has this been discussed? We don't want users to be able to switch?

Yes, it's been discussed. Most users should have no need to switch between networks. In the event a user/developer does wish to, they can enable the dev options by tapping the bottom of the main setting menu five times.

PR #645, opened 2022-11-01
fix(wallet): Limit Address Scanning

coreyphillips:
This PR:
- Adds `scanAllAddresses` param to several methods.
- Setting `scanAllAddresses` to `false` will force adherence to the set gap limit (20 at the moment) in order to limit scanning all addresses every time the user refreshes.
- Adds condition to `QuickConfirm` & `CustomConfirm` components to enable lightning testing on iOS devices for `regtest` and `testnet` networks.

### DIFF
--- a/src/utils/startup/index.ts
+++ b/src/utils/startup/index.ts
@@ -149,6 +149,7 @@ export const startWalletServices = async ({
 					refreshWallet({
 						onchain: isConnectedToElectrum,
 						lightning: isConnectedToElectrum,
+						scanAllAddresses: restore,
 						updateAllAddressTypes: restore,
 					}),
 				]);
--- a/src/utils/wallet/electrum.ts
+++ b/src/utils/wallet/electrum.ts
@@ -40,15 +40,18 @@ export interface IGetUtxosResponse {
 
 /**
  * Returns utxos for a given wallet and network along with the available balance.
- * @param selectedWallet
- * @param selectedNetwork
+ * @param {string} [selectedWallet]
+ * @param {TAvailableNetworks} [selectedNetwork]
+ * @param {boolean} [scanAllAddresses]
  */
 export const getUtxos = async ({
 	selectedWallet,
 	selectedNetwork,
+	scanAllAddresses = false,
 }: {
 	selectedWallet?: string;
 	selectedNetwork?: TAvailableNetworks;
+	scanAllAddresses?: boolean;
 }): Promise<Result<IGetUtxosResponse>> => {
 	try {
 		if (!selectedNetwork) {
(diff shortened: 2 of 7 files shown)


src/utils/wallet/electrum.ts
```diff
@@ -73,22 +76,63 @@ export const getUtxos = async ({
...
+				let addresses: IAddress = {};
+				let changeAddresses: IAddress = {};
+				// Instead of scanning all addresses, adhere to the gap limit.
+				if (!scanAllAddresses && addressIndex >= 0 && changeAddressIndex >= 0) {
+					await Promise.all([
+						Object.values(allAddresses).map((a) => {
+							if (Math.abs(a.index - addressIndex) <= GAP_LIMIT) {
+								addresses[a.scriptHash] = a;
+							}
+						}),
```
limpbrains, line 109: Sorry, can you please explain the logic here.
Let's say we have 100 addresses. AddressIndex = 99, Looks like this condition will always be true.

coreyphillips: Sure thing! Here's an example:
```
const addresses = [1,2,3,4,5,6,7,8,9,10]; // This represents an array of address indexes.
const addressIndex = 4; // This is our current index.
const gapLimit = 2;

const addressesToScan = addresses.filter((a) => Math.abs(a - addressIndex) <= gapLimit);

console.log(addressesToScan);
// [2, 3, 4, 5, 6]
```

Grabbing the absolute value of an address index minus the current index allows us to retrieve the specified gap limit both behind and in front of the current index.

limpbrains: What if user receives transaction to address at index 1 ? He will not be able to get it on pull-to-refresh on main screen?

coreyphillips: Two things on that:
- When the app restores it will scan all addresses [here](https://github.com/synonymdev/bitkit/blob/cc7ce8cda3bc6c14a90dc142930618a8d3b22ecc/src/utils/startup/index.ts#L152). Subsequent refreshes after startup will adhere to the gap limit standard of 20.
- The app will not generate receiving addresses beyond the specified gap limit from the receive view.

coreyphillips: Scratch that. I need to replace `restore` with `true`. Getting a quick fix in now.

coreyphillips: Like other wallets, we may need to add a rescan button to the Settings menu that rescans from index 0.

src/utils/startup/index.ts
```diff
@@ -149,6 +149,7 @@ export const startWalletServices = async ({
 					refreshWallet({
 						onchain: isConnectedToElectrum,
 						lightning: isConnectedToElectrum,
+						scanAllAddresses: restore,
```
pwltr, line 152: This only scans all addresses when restoring a wallet, above you said it should run on every app start

pwltr (changes requested):

limpbrains:
I think pull-to-refresh on main screen should act like a Rescan everything button

src/utils/startup/index.ts
coreyphillips: I corrected my comment. Scanning all addresses on every start can get demanding when the app starts to manage hundreds of addresses. We need something more controlled and manageable for a mobile wallet like John suggested (Rotating a subset of old addresses outside of the scanning threshold on every start and/ or refresh).

coreyphillips:
> I think pull-to-refresh on main screen should act like a Rescan everything button

I think this is a good solution. This would make pull to refresh behave how the user expects it and keeps the default demand on scanning performant by using the gap limit as the basis for setting the scanning threshold.

pwltr (approved):

limpbrains (approved):

PR #650, opened 2022-11-03
Feat/delete auth widget

### LINKED ISSUE #648: [Feature]: delete auth widgets
### Describe the problem

> There is no delete button for the widget that appears with the Slashtags auth demo. So you can't delete it from the wallet. There is also no logo.

> Auth demo widget should also show the settings icon when tapping it (just like the other widgets), so a user is able to delete the widget.

### Describe the solution

Open Widget settings but skip the feed resolution, to enable deleting widgets!

### Additional context

_No response_

Nuhvi:
closes #648

https://user-images.githubusercontent.com/40009100/199691600-7c733410-74c7-434d-b0ec-f45330f19c96.mov


limpbrains (approved):

src/components/HeadlinesWidget.tsx
```diff
@@ -119,17 +131,37 @@ const HeadlinesWidget = ({
...
 						onPress={(): void => {
 							setTimeout(() => setShowButtons(false), 0);
 							navigate('WidgetFeedEdit', { url });
 						}}
 					/>
 				</View>
 			)}
+			<Dialog
+				visible={showDialog}
+				title={`Delete ${name} widget?`}
```
pwltr: `name` here should always be just the widget name imo

pwltr (changes requested):

pwltr (approved):

PR #658, opened 2022-11-04
fix(settings): About links

pwltr:


pwltr:
@limpbrains Can you test the `mailto` link on iPhone please? On the simulator there is no mail app for testing

coreyphillips (approved):

PR #660, opened 2022-11-05
fix(lightning): Reset Channel-Open Transactions

### LINKED ISSUE #655: [Bug]: Unable to open a channel
### Describe the bug

I am trying to open a channel, I received two onchain payments, so I have two utxo, then I'm trying to open a channel, I tried with the quick way and the advance way, both I get the same error.
```
Unable to create transaction
Outputs are spending more than inputs
```

### Reproduce

Receive two onchain payments, one with amount o 200k and the other one 400k.
Then try to open a channel
I tried to open a channel with 100% on the quick way but also I tried to open a channel of 400k, I have 600k so it should open but is not

### Screenshots / Recording

photo_2022-11-03_16-43-15

### Operating system

Android 13

### Bitkit version

v1.0.0

### Log output

_No response_

coreyphillips:
This PR:
- Adds `resetOnChainTransaction` method to both `CustomSetup` & `QuickSetup` components.
- This closes #655

### DIFF
--- a/src/screens/Lightning/CustomSetup.tsx
+++ b/src/screens/Lightning/CustomSetup.tsx
@@ -28,7 +28,10 @@ import NumberPadLightning from './NumberPadLightning';
 import type { LightningScreenProps } from '../../navigation/types';
 import Store from '../../store/types';
 import { useBalance } from '../../hooks/wallet';
-import { setupOnChainTransaction } from '../../store/actions/wallet';
+import {
+	resetOnChainTransaction,
+	setupOnChainTransaction,
+} from '../../store/actions/wallet';
 import {
 	fiatToBitcoinUnit,
 	getFiatDisplayValues,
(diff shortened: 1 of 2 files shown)


src/screens/Lightning/CustomSetup.tsx
```diff
@@ -178,6 +181,15 @@ const CustomSetup = ({
 			: maxSpendingLimit.fiatValue;
 	}, [currentBalance.fiatValue, selectedCurrency]);
 
+	useEffect(() => {
```
pwltr: Do we need to use `useFocusEffect` here instead? Otherwise it will only run once on app mount.

coreyphillips: Implemented [here](https://github.com/synonymdev/bitkit/pull/660/commits/f3bb845261db2a80655f723160a33115971133c8).

pwltr (approved):

limpbrains (approved):

PR #661, opened 2022-11-06
feat(wallet): Implemented react-native-v8

coreyphillips:
This PR:
- Implements `react-native-v8` for Android.
- Updates several dependencies.

Notes:
- This PR swaps out `hermes` for `v8` on Android. This switch makes the app much more performant on physical devices and significantly improves the UX on older, less capable Android devices.
- I had to remove any hardcoded integer with an underscore throughout the app. `v8` does not seem to enjoy it.


pwltr:
Damn, this really does feel a lot more snappy. Got it to build and run everywhere (Intel & M1 emulators, device). Will test a bit more though because I had the occasional black screen for a few seconds on initial launch.

limpbrains:
I see some open issues there regarding react-native-reanimated library. I wonder if using v8 might block us from upgrading RN version in future?

coreyphillips:
> I see some open issues there regarding react-native-reanimated library. I wonder if using v8 might block us from upgrading RN version in future?

I've never had issues implementing v8 with the latest version of RN in the past, but react-native-reanimated support has been spotty. Luckily, reanimated v2 is now supported and it seems react-native-reanimated version 3 support appears to have been merged [here](https://github.com/software-mansion/react-native-reanimated/pull/3736) so that shouldn't be an issue once we decide to upgrade the reanimated library.

pwltr (approved):

PR #668, opened 2022-11-07
feat: move to react-native-camera-kit

limpbrains:
Replacing `react-native-camera` with new `react-native-camera-kit`
Unfortunately it can't check permissions on android so it just shows black screen if user denies camera access.

@pwltr , please test in on android, I've only tried it on a emulator. Also there might be problems with m1


coreyphillips:
Didn't have a problem building it on an M1, but the camera view is black when first loaded even with permissions accepted. The camera will work if I close out and re-enter, however.

limpbrains:
> but the camera view is black when first loaded even with permissions accepted.

Android?

coreyphillips:
Sorry. Yes, on Android.

coreyphillips:
If a solution to the black screen on Android can't be found. We could try roping in `react-native-permissions` and call something similar to the following method prior to rendering the camera:

```
import { check, request, PERMISSIONS, RESULTS } from 'react-native-permissions';

const [hasCameraPermission, setHasCameraPermission] = useState<undefined | boolean>(undefined);

const handleCameraPermission = async () => {
		const cameraPermission =
			Platform.OS === 'ios'
				? PERMISSIONS.IOS.CAMERA
				: PERMISSIONS.ANDROID.CAMERA;
		const checkResponse = await check(cameraPermission);
		switch (checkResponse) {
			case RESULTS.UNAVAILABLE:
			case RESULTS.BLOCKED:
				break;
			case RESULTS.DENIED:
				const rationale = {
					title: 'Permission to use camera',
					message: 'Bitkit needs permission to use your camera',
					buttonPositive: 'Okay',
					buttonNegative: 'Cancel',
				};
				const requestResponse = await request(cameraPermission, rationale);
				if (requestResponse === RESULTS.GRANTED) {
					setHasCameraPermission(true);
				} else {
					setHasCameraPermission(false);
				}
				break;
			case RESULTS.LIMITED:
			case RESULTS.GRANTED:
				setHasCameraPermission(true);
				break;
		}
	};
```

pwltr:
> Didn't have a problem building it on an M1, but the camera view is black when first loaded even with permissions accepted. The camera will work if I close out and re-enter, however.

Got the same behaviour on emulator & device

limpbrains:
updated to use react-native-permissions

coreyphillips:
Nice! Working as expected on my end.

coreyphillips (approved):

limpbrains:
Sorry, didn't see there is no approval from @pwltr. Does it work for you as-well, Philipp?

pwltr:
No completely, I'm testing right now but there seem to be issues on Android when opening camera from Send flow

pwltr:
Freezing for me when opening scanner in Send flow (emulator and device):

https://user-images.githubusercontent.com/8538369/200815492-44f553b3-89cf-458b-b4f2-c99822a95e1c.mov

limpbrains:
Thx, I will revert it then

PR #684, opened 2022-11-11
docs: add docs for development & build

pwltr:


BUILD.md
```diff
@@ -0,0 +1,27 @@
...
+1. Clone the repository
+
+```shell
+git clone git@github.com:synonymdev/bitkit.git && cd bitkit
+```
+
+2. Switch Node version
+
+```shell
+nvm use
```
limpbrains: You don't have to use nvm. I don't. Let's just mention that it is better to use node of the certain version

limpbrains (approved):

PR #690, opened 2022-11-14
fix: reduce flickering on restore flow

limpbrains:


src/screens/Onboarding/Restoring.tsx
```diff
@@ -45,8 +48,11 @@ const RestoringScreen = (): ReactElement => {
 		})();
 	}, [onRemoteRestore]);
 
+	let color = 'blue';
```
pwltr: Color for Loading screen should always be "brand" I believe

pwltr (changes requested):

src/screens/Onboarding/Restoring.tsx
limpbrains: right

pwltr (approved):

PR #695, opened 2022-11-14
fix(scan): freeze on Android

pwltr:
Changing to correct `TouchableOpacity` imports actually fixes the camera freeze on Android. Also fixed for other components.


limpbrains (approved):
Omg, no idea why I used it in ScannerComponent

pwltr:
> Omg, no idea why I used it in ScannerComponent

It comes up first for auto-import, happened to me too a lot

Issue #708, opened 2022-11-16
[Bug]: Pay to LN zero amount invoice broken


aldertnl:
### Describe the bug

Paying to a zero amount LN invoice is currently broken. 

This is problematic, as copying and sharing a BIP21 URI with BOLT 11 invoice (with both on-chain address and Lightning invoice) is the default behavior in the receive flow, and thus users can currently only use LN receive functionality if they specify an invoice.

### Reproduce

1. Go to 'Receive'
2. Click on 'Copy'
3. Share invoice with someone
4. They try to pay the invoice
5. It will not work

### Operating system

iPhone (and probably Android)

### Bitkit version

1.0 (19)

### Log output

```shell
Error creating transaction
```

PR #709, opened 2022-11-16
fix(lightning): Update payWithTimeout Method

coreyphillips:
This PR:
- Updates `payWithTimeout` usage in `payLightningInvoice`.
- Updates response type for `payLightningInvoice`.
- Passes lightning tx fees to notification and activity object.
- Bumps lightning payment timeout to 30 seconds.
- This addresses #708

### DIFF
--- a/src/utils/lightning/index.ts
+++ b/src/utils/lightning/index.ts
@@ -7,6 +7,7 @@ import lm, {
 	TAccountBackup,
 	TChannel,
 	TChannelManagerPayment,
+	TChannelManagerPaymentSent,
 	TCloseChannelReq,
 	THeader,
 	TInvoice,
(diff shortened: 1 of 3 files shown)


src/utils/lightning/index.ts
```diff
@@ -927,22 +928,11 @@ export const payLightningInvoice = async (
...
-			payResponse = await lm.payWithTimeout({
-				paymentRequest: invoice,
-			});
-		}
-		if (!payResponse) {
-			return err('Unable to pay the provided lightning invoice.');
-		}
+		const payResponse = await lm.payWithTimeout({
+			paymentRequest: invoice,
+			amountSats: sats ?? 0,
```
Jasonvdb, line 933: Should be able to pass `sats` as an optional number instead of `0`. Although it should work either way.

Jasonvdb (approved):

PR #714, opened 2022-11-17
fix: various fixes

### LINKED ISSUE #593: State mutation
When creating new wallet Redux error appear on the console:
```
Error: Invariant failed: A state mutation was detected between dispatches, in the path 'wallet.wallets.wallet0.addresses.bitcoin.p2wpkh'.  This may cause incorrect behavior.
```

### LINKED ISSUE #625: [Bug]: Scan new contact from photo library
When arriving a contact using a screenshot from the photo library, the scan is ignored

**To Reproduce**
1. Go to contacts
2. Add contact using the scan option
3. On the top left corner , Select photo library and select the picture with he QR code —> ok
4. The QR code is ignored

**Expected behavior**
The QR code on the picture should be scan and contact added 

**Screenshots**
NA

 - OS: iOS

**Additional context**
NA

pwltr:
### Changes
- fix #625 
- fix(wallet): #593 when creating wallet
- fix(settings): update activity on network change
- fix(send): only show available balance if recipient is known
- fix(send): amount spacing
- chore: add some types & cleanup

### Notes
Contains a temp solution for state mutation using `immer` until we update all reducers to use RTK utils.

### Preview

https://user-images.githubusercontent.com/8538369/202548584-ccaef15c-ed51-4a98-adc8-678873eb92b2.mp4

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "fast-sha256": "^1.3.0",
     "fuzzysort": "^1.9.0",
     "i18next": "^21.8.3",
+    "immer": "^9.0.16",
     "intl": "^1.2.5",
     "jdenticon": "^3.1.1",
     "lodash.debounce": "^4.0.8",
--- a/src/styles/components.ts
+++ b/src/styles/components.ts
@@ -486,7 +486,6 @@ export const Display = styled.Text((props) => ({
 		? props.theme.fonts[props.font].fontFamily
 		: 'NHaasGroteskDSW02-65Md',
 	fontSize: props.size ? props.size : '48px',
-	lineHeight: props.lineHeight ?? '49px',
 }));
 
 export const Headline = styled.Text((props) => ({
--- a/src/styles/components.ts
+++ b/src/styles/components.ts
@@ -486,6 +486,7 @@ export const Display = styled.Text((props) => ({
 		? props.theme.fonts[props.font].fontFamily
 		: 'NHaasGroteskDSW02-65Md',
 	fontSize: props.size ? props.size : '48px',
+	lineHeight: props.lineHeight ?? '48px',
 }));
 
 export const Headline = styled.Text((props) => ({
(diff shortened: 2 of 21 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
     "fast-sha256": "^1.3.0",
     "fuzzysort": "^1.9.0",
     "i18next": "^21.8.3",
+    "immer": "^9.0.16",
```
limpbrains, line 58: Can use lodash cloneDeep or `JSON.parse(JSON.stringify(x))` hack from https://github.com/synonymdev/bitkit/blob/master/src/store/helpers.ts#L9 instead of adding another dependency here?

src/styles/components.ts
```diff
@@ -486,7 +486,6 @@ export const Display = styled.Text((props) => ({
 		? props.theme.fonts[props.font].fontFamily
 		: 'NHaasGroteskDSW02-65Md',
 	fontSize: props.size ? props.size : '48px',
-	lineHeight: props.lineHeight ?? '49px',
```
limpbrains, line 489: If I remeber correctly, it is not a default value for this font size/type. Are you sure we can remove this?

limpbrains (changes requested):

package.json
pwltr: It is already a peerDependency of `@reduxjs/toolkit`. It let's us write mutable code in reducers which I would like to use more in the near future.

limpbrains: Nice, let's use it then

src/styles/components.ts
pwltr: It is not the correct lineHeight for this font which causes it to be uncentered in the lineHeight. Not really noticeable anywhere but in `<Money />` component so I'm overwriting it there now.

Simulator Screen Shot - iPhone SE (3rd generation) - 2022-11-18 at 14 35 28

limpbrains: Yeah, but without it all other places, where we are using Display, like on Slideshow. Will be broken

limpbrains (approved):

PR #715, opened 2022-11-18
perf: hardcode a mirroring seeder to test its performance

Nuhvi:
This PR adds helps to join a certain peer (a seeder that doesn't announce all cores on the DHT), to compare it with the current seeder, it shouldn't disrupt anything else and is simple to revert, hopefully, we can test it in internal staging soon.


Jasonvdb (approved):

Jasonvdb:
@Nazeh rather move to .env

PR #717, opened 2022-11-21
fix(wallet): getStore Updates

coreyphillips:
This PR:
- Reduces usage of `getStore` where able.
- Splits `getStore` into several methods for individual stores.

### DIFF
--- a/src/store/helpers.ts
+++ b/src/store/helpers.ts
@@ -1,12 +1,48 @@
 import store from '../store';
-import Store from './types';
+import Store, { EStore } from './types';
 import { Dispatch } from 'redux';
+import { IWallet } from './types/wallet';
+import { ISettings } from './types/settings';
+import { IMetadata } from './types/metadata';
+import { IActivity } from './types/activity';
+import { ILightning } from './types/lightning';
+import { IBlocktank } from './types/blocktank';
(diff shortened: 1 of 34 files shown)


src/store/helpers.ts
```diff
@@ -1,13 +1,49 @@
...
+import { IUser } from './types/user';
 
 /*
 Used to retrieve the store outside of a component.
  */
 export const getStore = (): Store =>
 	JSON.parse(JSON.stringify(store.getState()));
 
+export const getWalletStore = (): IWallet =>
+	JSON.parse(JSON.stringify(store.getState()[EStore.wallet]));
```
pwltr, line 21: Is there a reason to first stringify and then parse in the same line here? Usually can just access `store.getState()`

coreyphillips: Prevents mutating state by writing directly to the store.

pwltr: Seems an expensive way to clone an object, or do you think it's optimized enough by the JS engine?

coreyphillips: It's optimized enough for now, but yes I agree, we'll eventually need a better solution. Previous attempts to try something else yielded odd/inconsistent results.

pwltr (approved):

PR #730, opened 2022-11-25
chore: move seeder topic and base url to env

Nuhvi:
@Jasonvdb This will need to be added to `env` before release obviously!


Jasonvdb:
Also add to the file `.env.development.template` in the root dir

Jasonvdb (approved):

PR #745, opened 2022-11-29
feat: draggable widgets

limpbrains:
- use react-native-draggable-flatlist for widgets on main screen
- change style of edit/delete buttons
- add new errors to LogBox ingnore list


src/store/index.ts
```diff
@@ -44,7 +44,7 @@ const persistConfig = {
 	key: 'root',
 	storage: mmkvStorage,
 	// increase version after store shape changes
-	version: 1,
+	version: 2,
```
limpbrains, line 47: would that be enought?
I added state.widgets.sortOrder branch

pwltr: I think add this in `migrations/index.ts`:
```
2: (state): PersistedState => {
		return {
			...state,
			widgets: {
				...state.widgets,
				sortOrder: [],
			},
		};
	},
```

But since you handle the case where `sortOrder` is undefined no migration is necessary. I personally like migrations more, and keeping app code clean.

```
const sortOrder = useSelector(
		(state: Store) => state.widgets.sortOrder ?? []
);
```

limpbrains: added migration

pwltr:
> Tapping outside of widget area > Resets view to default states (hides buttons).

This would be nice to have. Otherwise LGTM.

pwltr (approved):

pwltr:
This scrolling behaviour is a bit weird. Do we need it here?

https://user-images.githubusercontent.com/8538369/204604925-2e332dba-40ad-4d35-b36d-c2183c303693.mp4

limpbrains:
> This scrolling behaviour is a bit weird. Do we need it here?
> Simulator.Screen.Recording.-.iPhone.13.-.2022-11-29.at.18.47.02.mp4

I don't have it. Will test again

PR #760, opened 2022-12-07
fix(wallet): More Performance Updates

coreyphillips:
This PR:
- Implements `memo`/`useMemo`/`useCallback` in several components to prevent re-renders when able.
- Pulls several methods/variables out of render completely.

### DIFF
--- a/src/screens/Lightning/PieChart.tsx
+++ b/src/screens/Lightning/PieChart.tsx
@@ -1,4 +1,4 @@
-import React, { memo, ReactElement, useMemo } from 'react';
+import React, { memo, ReactElement } from 'react';
 import {
 	DashPathEffect,
 	BlurMask,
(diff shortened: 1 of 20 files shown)


limpbrains:
I'm worry that we are overusing memorization. We already have most components wrapped in memo(
- what is the point of using it inside Chart component, when you are sure it will be rendered only once.
- useMemo eats memory 
- I don't think putting simple comparisons, like `c = useMemo((a && b) => {},[a, b])`. There operations are super fast and using useMemo might make them actually slower
- it complicates the code

coreyphillips:
Agreed, we have to be careful to balance use of the feature. I reverted the overkill in `PieChart` [here](https://github.com/synonymdev/bitkit/pull/760/commits/261afe30d8f7a6ea91308b1247fb6faa2b822dda).

limpbrains (approved):

PR #762, opened 2022-12-07
feat(profile): add prefix to profile links if suggestions were used

Nuhvi:
This PR tries to improve profile links input, without going too crazy with regex and validating every URI under the sun!

I regret that I couldn't use `useRef` to focus on the input text too.

simulator_screenshot_4174559E-0242-4448-AAE3-27B7CFC9F3C3

### DIFF
--- a/src/screens/Profile/ProfileLinkSuggestions.tsx
+++ b/src/screens/Profile/ProfileLinkSuggestions.tsx
@@ -8,30 +8,30 @@ import { updateProfileLink } from '../../store/actions/ui';
 import SafeAreaInsets from '../../components/SafeAreaInsets';
 import type { RootStackScreenProps } from '../../navigation/types';
 
-const suggestions = [
-	'Email',
-	'Phone',
-	'Website',
-	'Twitter',
-	'Telegram',
-	'Instagram',
-	'Facebook',
-	'LinkedIn',
-	'Github',
-	'Calendly',
-	'Vimeo',
-	'YouTube',
-	'Twitch',
-	'Pinterest',
-	'TikTok',
-	'Spotify',
-];
+export const suggestions = {
+	Email: { prefix: 'mailto:' },
+	Phone: { prefix: 'tel:' },
+	Website: { prefix: 'https://' },
+	Twitter: { prefix: 'https://twitter.com/' },
+	Telegram: { prefix: 'https://t.me/' },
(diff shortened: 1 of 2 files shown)


src/screens/Profile/ProfileLinkSuggestions.tsx
```diff
@@ -8,30 +8,30 @@ import { updateProfileLink } from '../../store/actions/ui';
...
-	'Vimeo',
-	'YouTube',
-	'Twitch',
-	'Pinterest',
-	'TikTok',
-	'Spotify',
-];
+export const suggestions = {
+	Email: { prefix: 'mailto:' },
+	Phone: { prefix: 'tel:' },
```
pwltr, line 13: I feel like pre-filling especially these two strings will be confusing to the avg user. I suggest we move these prefixes to the render code in `ProfileLinks`.

EDIT:
Or maybe add the prefix before saving, whatever the intention is here.

pwltr (changes requested):

src/screens/Profile/ProfileLinkSuggestions.tsx
Nuhvi: I started with adding the prefix before saving, but it is kinda tough, for example this looks benign:
```js
const url = form.url.startsWith(prefix)) ? form.url : prefix + form.url 
```

But if the prefix is `https://twitter.com/` and the user enters `https://www.twitter.com` or vice versa, then you get a bad url. I basically chose the easry road that will just nudge users just a step closer to the correct url formats.

If you have any more elaborate checks and transformations, feel free to close this PR, I just wanted to suggest a quick fix.

Nuhvi: Also ... none of this actually solve the issue of validating each suffix according to the `form.title` for example phone numbers require different validation than a website or an email, and if someone writes `https://twitter.com/foo/bar` that shouldn't be valid either!

pwltr: I think the links are fine as a first step just the phone number and email I wouldn't do this way. Usually they aren't saved to a database in that format. Leave it up to the consumer how to render these two. I'm down to work on validation soon.

pwltr:
Closing in favor of https://github.com/synonymdev/bitkit/pull/1080

PR #763, opened 2022-12-07
feat(wallet): Implement Reselect

coreyphillips:
This PR:
- Implements reselect for all `selectedWallet`/`selectedNetwork` selectors throughout the app.
- Creates `reselect` directory.
- Upgrades `@reduxjs/toolkit`.

Notes:
- All selectors are contained in the reselect directory for reference and are separated by store.
- Only `selectedWallet` & `selectedNetwork` selectors are implemented in this PR, but most selectors that the app uses have already been created and added to the reselect directory within the relevant store's file.
- A great benefit of this approach is that we only need to write selectors once and can reuse them as needed and easily change them using a single reference. Reselect is also great for creating and reusing logic for filtered/mapped state data.
- A minor pain point I found while implementing this is that creating a selector with multiple arguments can be a bit cumbersome to write, but once written the convenience of using the selector outweighs having to create repeated/expensive logic, otherwise. However, if you know of a better way to pass arguments, please let me know.

Usage:
Before: `useSelector((state: Store) => state.wallet.selectedWallet);`
After: `useSelector(selectedWalletSelector);`

Usage with param: `useSelector((state: Store) => viewControllerSelector(state, 'sendNavigation'));`


limpbrains (approved):

src/store/reselect/fees.ts
```diff
@@ -0,0 +1,10 @@
+import Store from '../types';
+import { createSelector } from '@reduxjs/toolkit';
+import { IOnchainFees } from '../types/fees';
+
+const onChainFeesState = (state: Store): IOnchainFees => state.fees.onchain;
+
+export const onChainFeesSelector = createSelector(
+	onChainFeesState,
+	(onChainFees): IOnchainFees => onChainFees,
+);
```
pwltr, line 10: Looking at the [docs (Caution box)](https://redux.js.org/usage/deriving-data-selectors#createselector-overview) this pattern is apparently incorrect. I'm not sure it breaks anything but it probably doesn't add any value either.

So these two would be the same (but latter is more code):
```
const onChainFees = useSelector(onChainFeesState)
const onChainFees = useSelector(onChainFeesSelector)
```

coreyphillips: I see, good catch. Instead of passing back its state/input directly I need to actually derive selectedWallet in the createSelector from wallet in order for it to function as intended.

So instead of:
```
export const selectedWalletSelector = createSelector(
	selectedWalletState,
	(selectedWallet): string => selectedWallet,
);
```

It should be:
```
export const selectedWalletSelector = createSelector(
	[walletState],
	(wallet): string => wallet.selectedWallet,
);
```

Putting together a fix now.

coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/763/commits/3d98204db0890d791cc60ca036e28f43c8904a6f).

src/store/reselect/receive.ts
```diff
@@ -2,9 +2,9 @@ import Store from '../types';
...
 
-export const receiveState = (state: Store): IReceive => state.receive;
+const entireState = (state: Store): Store => state;
 
 export const receiveSelector = createSelector(
-	receiveState,
-	(receive): IReceive => receive,
+	[entireState],
+	(state): IReceive => state.receive,
 );
```
pwltr: > Similarly, a memoized selector should never use state => state as an input! That will force the selector to always recalculate.

This should probably be written as: 

```ts
export const receiveSelector = (state: Store): IReceive => state.receive;
```

coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/763/commits/ca8641412d2382a4c9e9531f84276d1f6d997e2e). I also removed `store.ts`.

src/store/reselect/lightning.ts
```diff
@@ -0,0 +1,88 @@
...
+import { createSelector } from '@reduxjs/toolkit';
+import { TAvailableNetworks } from '../../utils/networks';
+
+export const lightningState = (state: Store): ILightning => state.lightning;
+export const nodesState = (state: Store): TNodes => state.lightning.nodes;
+
+export const lightningSelector = createSelector(
+	lightningState,
+	(lightning): ILightning => lightning,
+);
```
pwltr, line 12:
```suggestion
export const lightningSelector = lightningState;
```

pwltr (changes requested):

pwltr (approved):

PR #764, opened 2022-12-12
fix: try to close the app after Wipe App action

limpbrains:
Because running async actions can mutate state after Wipe App action, we need to try to close it, if that is possible.
BackHandler.exitApp(); works on Android, but not on iOS, so there we just show warning, explainin that app needs to be restarted

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -331,6 +331,8 @@ PODS:
     - React-Core
   - react-native-randombytes (3.6.1):
     - React-Core
+  - react-native-restart (0.0.24):
+    - React-Core
   - react-native-safe-area-context (4.3.3):
     - RCT-Folly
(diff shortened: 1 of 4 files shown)


pwltr:
Also needs to be added in `ResetAndRestore.tsx`

limpbrains:
In new version I'm using react-native-restart to restart app on iOS, it doesn't work nicely on Android, so there I'm just closing the app using BackHandler.exitApp();

pwltr:
Getting a bit of broken UI on iOS after wiping on PinPad or ForgotPIN screen. Probably not a big issue as the user can just restart the app again to fix.

Simulator Screen Shot - iPhone 13 - 2022-12-12 at 15 23 05

pwltr (approved):

PR #778, opened 2022-12-15
Share LDK logs

Jasonvdb:
Zips up newest 10 logs files and uses react-native-share to share it

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -320,7 +320,7 @@ PODS:
     - React-Core
   - react-native-image-picker (4.10.0):
     - React-Core
-  - react-native-ldk (0.0.76):
+  - react-native-ldk (0.0.77):
     - React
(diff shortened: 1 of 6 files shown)


coreyphillips:
I'm able to successfully export 10 logs from a live Android device on mainnet, however, they appear to be empty.

Screen Shot 2022-12-15 at 8 35 31 AM

coreyphillips:
Upgrading `react-native-ldk` to 0.0.77 seems to prevent Bitkit from generating invoices and appears to have trouble updating claimable/pending balances when testing on an Android device (Pixel 6 Pro) using mainnet.

Jasonvdb:
Empty log files in zip should be patched

coreyphillips:
It works! LGTM

coreyphillips (approved):

PR #784, opened 2022-12-16
fix(wallet): Reset Transaction Object

coreyphillips:
This PR:
- Resets the transaction object after any failed broadcast attempt.
  - Otherwise, the user will most likely run into the same issue unless they specifically close out of and return to the `Send` view. Which is not an ideal experience.


limpbrains:
I don't understrand why do we need this? What if it was just a network connection error?

coreyphillips:
This is a quick fix to resolve the following issue prior to the next release:

https://synonymworkspace.slack.com/archives/C01H106JRQB/p1671206166070299?thread_ts=1671202573.549169&cid=C01H106JRQB

My next task is to improve error responses throughout the app where we'll be able to distinguish between certain errors more clearly and respond accordingly. The case you mentioned, being one example.

limpbrains (approved):

PR #798, opened 2022-12-28
feat: new send tests

limpbrains:
- new send wallet-send test
- rbf now false by default in `setupOnChainTransaction`
- `setupOnChainTransaction`, `updateBitcoinTransaction`, `updateSelectedFeeId` and `resetOnChainTransaction` can now return errors
- fix using of `InteractionManager.runAfterInteractions` in `resetLdk`
- replace `new Buffer` with `Buffer.from` to remove deprication warning


src/store/actions/wallet.ts
```diff
@@ -821,7 +821,7 @@ export const setupOnChainTransaction = async ({
 	selectedNetwork,
 	addressType,
 	inputTxHashes,
-	rbf = true,
+	rbf = false,
```
pwltr, line 824: Do we need to pass `rbf: true` now for usage of `setupOnChainTransaction` in `setupRbf` function?

__tests__/wallet-send.ts
```diff
@@ -0,0 +1,313 @@
...
+	updateFee,
+	validateTransaction,
+} from '../src/utils/wallet/transactions';
+import { addElectrumPeer } from '../src/store/actions/settings';
+import { getScriptHash } from '../src/utils/wallet';
+import initWaitForElectrumToSync from './utils/wait-for-electrum';
+
+jest.setTimeout(60_000);
+
+const bitcoinURL =
```
pwltr, line 27: Do I need to change this URL to be able to run the tests locally? I have the containers running and ran chmod on the `.cookie` file.

__tests__/wallet-send.ts
```diff
@@ -0,0 +1,313 @@
...
+	'http://electrumx:1VmSUVGBuLNWvZl0LExRDW0tvl6196-47RfXIzS384g=@localhost:43782';
```
pwltr, line 28: Do I need to change this URL to be able to run the tests locally? I have the containers running and ran chmod on the .cookie file.

src/store/actions/wallet.ts
limpbrains: Nice catch! I should probably revert it or add more tests, thx

__tests__/wallet-send.ts
limpbrains: You don't if you are running `docker-compose up` from `__tests__`. 
chmod needed on linux, on mac you don't have to do this

pwltr: Working now!

src/store/actions/wallet.ts
limpbrains: I've fixed usage of setupOnChainTransaction in setupRbf, please review again

pwltr (approved):

PR #799, opened 2022-12-28
fix(wallet): Update Dust Limit Logic

coreyphillips:
This PR:
- Updates `createPsbtFromTransactionData` method to remove dust outputs and apply them to the fee.
- Updates `createPsbtFromTransactionData` method to generate a `changeAddress` on the fly in the event there are additional sats when constructing a transaction.
- Creates `removeDustOutputs` & `getChangeAddress` methods.
- Implements `removeDustOutputs` in `createTransaction` method.
- Adds `dustLimit` value to `ETransactionDefaults`.

To Test:
  - Create an on-chain transaction where you are sending 200 sats shy of your max amount. (Toggle the "Max" button on, toggle "Max" button off, and remove 200 sats from the total value).
  - Send the transaction.
  - The on-chain transaction should apply what would have been a dust amount of 200 sats to the fee upon inspection.

### DIFF
--- a/src/screens/Wallets/Send/Result.tsx
+++ b/src/screens/Wallets/Send/Result.tsx
@@ -84,6 +84,11 @@ const Result = ({
 				data: { isOpen: false },
 			});
 		} else {
+			/*
+				TODO: Add ability to distinguish between errors sent to this component.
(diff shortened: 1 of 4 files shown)


pwltr (approved):
Working as intended!

PR #818, opened 2023-01-11
refactor: add & change types for wallet and components

pwltr:
This PR adds types for component props and wallet entities. Also includes a minor bugfix.

### Changes
- add some `@types` packages
- update eslint config
    - enable rules to require types
    - fix some instances of `implicit-arrow-linebreak` that are hard to read
    - add `prefer-optional-chain`
- add missing component prop types
- remove unused Bitcoin units mBTC, μBTC
- rename some enum from plural to singular
- adds `objectKeys` for a strongly-typed version of `Object.keys()`
- renamed some types to fix plural/singular and readability:
    - `IWallet` -> `IWalletStore`
    - `IDefaultWalletShape` -> `IWallet`
    - `IDefaultWallet` -> `IWallets`
    - `IAddress` -> `IAddresses`
    - `IAddressContent` -> `IAddress`
    - `IFormattedTransaction` -> `IFormattedTransactions`
    - `IFormattedTransactionContent` -> `IFormattedTransaction`
    - `IBoostedTransaction` -> `IBoostedTransactions`
- removed some occurences `no-misused-promises`

### Bugfixes
- fix UI store connectivity updates (isOnline, isConnectedToElectrum)
- fix `getPendingChannels` (unused)


src/utils/wallet/index.ts
```diff
@@ -1448,9 +1461,11 @@ export const getInputData = async ({
 			}
 			getTransactionsResponse.value.data.map(({ data, result }) => {
 				const vout = result.vout[data.vout];
-				const addresses = vout.scriptPubKey?.addresses
-					? vout.scriptPubKey?.addresses
-					: [vout.scriptPubKey.address];
+				const addresses = vout.scriptPubKey.addresses
```
coreyphillips, line 1464: What's the reason behind this change?

src/utils/wallet/index.ts
```diff
@@ -1577,9 +1591,11 @@ export const formatTransactions = async ({
 		//Iterate over each output
 		const vout = result?.vout || [];
 		vout.map(({ scriptPubKey, value }) => {
-			const _addresses = scriptPubKey?.addresses
+			const _addresses = scriptPubKey.addresses
```
coreyphillips, line 1594: Same as previous. What's the reason behind this change?

coreyphillips (changes requested):

src/utils/wallet/index.ts
pwltr: `vout.scriptPubKey.address` can be undefined, so we need to check for that to not end up with `const addresses = [undefined]`

coreyphillips (approved):

PR #823, opened 2023-01-13
fix(send): #441 auto-forward on invoice paste

### LINKED ISSUE #441: Auto Forward For Review On Invoice Paste
**Describe the feature**
When pasting an invoice into the send form's text input, forward the user to the review screen if either the invoice is greater than zero or the user has already entered an amount and the user can afford it.

**Expected behavior**
1. Tap Send.
2. With an invoice ready to go in the clipboard, tap the paste icon on the send form.
3. If the user has already set an amount or the invoice is greater than zero and they can afford the set amount, forward them on to the review screen.

pwltr:
### Changes
- auto-forward on invoice pase
- return decoded data for `processInputData` and `handleData`
- fix CoinSelection screen gradient

### Preview

https://user-images.githubusercontent.com/8538369/212318185-678e6841-5e2c-4fa0-afc6-3339eff325b4.mp4

### DIFF
--- a/src/screens/Wallets/Send/AddressAndAmount.tsx
+++ b/src/screens/Wallets/Send/AddressAndAmount.tsx
@@ -201,6 +201,15 @@ const AddressAndAmount = ({
 		}
 	}, [decodedInvoiceAmount, getOutput?.value, lightningInvoice]);
 
+	const onContinue = useCallback((): void => {
+		let view: keyof SendStackParamList = 'ReviewAndSend';
+		// If auto coin-select is disabled and there is no lightning invoice.
+		if (!coinSelectAuto && !transaction?.lightningInvoice) {
+			view = 'CoinSelection';
+		}
+		navigation.navigate(view);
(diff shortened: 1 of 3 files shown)


src/screens/Wallets/Send/AddressAndAmount.tsx
```diff
@@ -233,8 +240,11 @@ const AddressAndAmount = ({
 					},
 				}).then();
 			}
+			if (result.isOk() && result.value.amount) {
+				onContinue();
+			}
 		},
-		[index, value, selectedNetwork, selectedWallet, sdk],
+		[index, value, selectedNetwork, selectedWallet, sdk, onContinue],
```
coreyphillips, line 247: I believe `onContinue` will have to be wrapped in `useCallback` so that `handlePaste` can determine if it has changed. Otherwise, it will always return true as a dependency.

coreyphillips (changes requested):

coreyphillips (approved):

PR #831, opened 2023-01-17
feat(wallet): Add Address Viewer

coreyphillips:
This PR:
- Adds the `AddressViewer` component to the Advanced Settings list.
- Updates `getUtxos` method in `electrum.ts`.
- Creates `listUnspentAddressScriptHashes` & `getAddressUtxos` methods in `electrum.ts`.
- Updates `getPrivateKey` method in `utils/wallet/index.ts`.
- Fixes a bug in `getUtxos` that would prevent certain address types from being returned in the utxo array.

Notes/Included Functionality:
- This new view can be found by navigating to "Settings->Advanced->Address Viewer".
- This view is currently enabled for everyone, but can be hidden behind dev options if more appropriate.
- The address viewer generates addresses associated with the user's current wallet on the fly.
- The user can toggle between address types and receive/change addresses.
- If the dev settings are enabled the user can also toggle between networks (mainnet, testnet, regtest).
- The user can generate 20 new addresses per address type by tapping "Generate 20 More". (20 * Each Address Type (3))
- The user can search for an address from the generated list.
- The user can check the balances of the generated addresses.
- If balances are discovered, the user can select or deselect the addresses containing a balance and choose to spend from them.
  - When the user attempts to spend from their selected addresses, the sendNavigation modal will appear and pre-fill in their receive address from Bitkit. If they prefer to send to an external address they can modify the address like a normal transaction by scanning or entering a new address.

Simulator Screen Shot - iPhone 13 - 2023-01-17 at 11 00 07
Simulator Screen Shot - iPhone 13 - 2023-01-17 at 11 00 31

### DIFF
--- a/src/screens/Settings/AddressViewer/index.tsx
+++ b/src/screens/Settings/AddressViewer/index.tsx
@@ -179,6 +179,7 @@ const getAllAddresses = async ({
 	addressAmount?: number;
 }): Promise<Result<TAddressViewerData>> => {
 	const responseData = { ...defaultAllAddressesData };
+	const start = performance.now();
 	await Promise.all(
 		Object.values(addressTypes).map(async ({ path, type }) => {
 			const keyDerivationPathResponse = getKeyDerivationPathObject({
@@ -213,6 +214,12 @@ const getAllAddresses = async ({
 				changeAddresses,
 			};
 		}),
+	);
+	const end = performance.now();
+	console.log(
+		`Time To Generate ${
+			addressAmount * Object.keys(addressTypes).length
+		} Addresses: ${end - start} ms`,
 	);
 	return ok(responseData);
 };
@@ -403,7 +410,7 @@ const AddressViewer = ({
 				(a) => a.index,
 			);
 			const maxIndex = Math.max(...indexes);
-			await setConfig({
+			setConfig({
 				...config,
 				addressIndex: maxIndex + 1,
 			});
@@ -521,7 +528,7 @@ const AddressViewer = ({
 				(a) => a.index,
 			);
 			const maxIndex = Math.max(...indexes);
-			await setConfig({
+			setConfig({
 				...newConfig,
 				addressIndex: maxIndex + 1,
 			});
@@ -618,15 +625,11 @@ const AddressViewer = ({
 	 */
 	const utxoIsSelected = useCallback(
 		(utxo: IUtxo): boolean => {
-			let isSelected = false;
-			if (utxos) {
-				isSelected = !!selectedUtxos.find(
-					(u) => u.address === utxo.address && u.tx_pos === utxo.tx_pos,
-				);
-			}
-			return isSelected;
+			return selectedUtxos.some(
+				(u) => u.address === utxo.address && u.tx_pos === utxo.tx_pos,
+			);
 		},
-		[selectedUtxos, utxos],
+		[selectedUtxos],
 	);
 
 	/**
@@ -660,7 +663,7 @@ const AddressViewer = ({
 			if (utxosLength <= 0) {
 				return;
 			}
-			await resetOnChainTransaction({
+			resetOnChainTransaction({
 				selectedWallet,
 				selectedNetwork,
 			});
@@ -673,7 +676,7 @@ const AddressViewer = ({
 			if (transactionRes.isErr()) {
 				return;
 			}
-			const receiveAddress = await getReceiveAddress({
+			const receiveAddress = getReceiveAddress({
 				selectedWallet,
 				selectedNetwork,
 			});
--- a/src/screens/Settings/AddressViewer/index.tsx
+++ b/src/screens/Settings/AddressViewer/index.tsx
@@ -1,0 +1,1178 @@
+import React, {
+	memo,
+	ReactElement,
+	useCallback,
+	useEffect,
+	useMemo,
+	useRef,
+	useState,
+} from 'react';
+import { FlatList, LayoutAnimation, StyleSheet, View } from 'react-native';
+
+import {
+	TouchableOpacity,
+	View as ThemedView,
+} from '../../../styles/components';
+import { Subtitle, Text02S } from '../../../styles/text';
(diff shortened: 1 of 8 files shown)


src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+				...config,
+				selectedWallet,
+				keyDerivationPath,
+				addressAmount,
+				changeAddressAmount: addressAmount,
+				changeAddressIndex: config.addressIndex,
+				addressType: type,
+			});
+			if (generateAddressResponse.isErr()) {
+				return err(generateAddressResponse.error.message);
```
limpbrains: This will break .map, but getAllAddresses will not return error.

Why use async + map, if here is no async call?

src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+						],
+					};
+				}),
+			);
+
+			const indexes = newAllAddresses[config.addressType].addresses.map(
+				(a) => a.index,
+			);
+			const maxIndex = Math.max(...indexes);
+			await setConfig({
```
limpbrains: unneceserly await

src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+	const updateAddressType = useCallback(
+		async (_addressType): Promise<void> => {
+			if (_addressType === config.addressType) {
+				return;
+			}
+			const newConfig = {
+				...config,
+				addressType: _addressType,
+			};
+			await setConfig(newConfig);
```
limpbrains: unneceserly await

src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+
+	/**
+	 * Determines whether the checkmark for a given UTXO is selected.
+	 * @returns {boolean}
+	 */
+	const utxoIsSelected = useCallback(
+		(utxo: IUtxo): boolean => {
+			let isSelected = false;
+			if (utxos) {
+				isSelected = !!selectedUtxos.find(
```
limpbrains: Array.some() can be used to check if element is in array

src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+	 * This method will gather all selected UTXO's and setup an on-chain transaction.
+	 * The on-chain transaction will retrieve and include the app's receiving address by default.
+	 * Finally, this method will prompt the sendNavigation modal to appear for the user to finalize and confirm the transaction.
+	 */
+	const onSpendFundsPress = useCallback(
+		async (utxosLength, selectedUtxosLength): Promise<void> => {
+			if (utxosLength <= 0) {
+				return;
+			}
+			await resetOnChainTransaction({
```
limpbrains: unnesecerly await

src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+			const transactionRes = await setupOnChainTransaction({
+				selectedWallet,
+				selectedNetwork,
+				utxos: selectedUtxosLength > 0 ? selectedUtxos : utxos,
+				rbf: true,
+			});
+			if (transactionRes.isErr()) {
+				return;
+			}
+			const receiveAddress = await getReceiveAddress({
```
limpbrains: unnesecerly await

src/screens/Settings/AddressViewer/index.tsx
```diff
@@ -0,0 +1,1178 @@
...
+			}
+		},
+		[selectedUtxos, utxos],
+	);
+
+	const onGenerateMorePress = useCallback(async () => {
+		setIsGeneratingMoreAddresses(true);
+		await getMoreAddresses(ADDRESS_AMOUNT);
+		setIsGeneratingMoreAddresses(false);
+	}, [getMoreAddresses]);
```
limpbrains, line 784: It would be nice to show how much time did it take, this way we will be able to benchmark address generation

limpbrains (commented):
Unhandled Promise rejection when I'm switching to Regtest and press Check Balances

image

Can't find exectly where is it

src/screens/Settings/AddressViewer/index.tsx
coreyphillips: Ah, thanks. That's part of a separate PR.

src/screens/Settings/AddressViewer/index.tsx
coreyphillips: We're awaiting `generateAddresses`. Otherwise, `responseData` has a chance of not getting set properly.

src/screens/Settings/AddressViewer/index.tsx
coreyphillips: Added a timer to `getAllAddresses` that will log how long it took to generate the 60 addresses to console.

coreyphillips:
Regarding [this](https://github.com/synonymdev/bitkit/pull/831#pullrequestreview-1251913083), when attempting to check balances of addresses using a network that the wallet is not currently set to, the Address Viewer will switch to that network in the background so that it can utilize the correct electrum server for the balance request.

I'm not too worried about it since this feature of the viewer is only enabled when dev settings are on, but this is most likely something we'll need to fix when switching networks in general.

src/screens/Settings/AddressViewer/index.tsx
limpbrains: I think this return still needs to be fixed

coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/831/commits/8c5989fbd9e64cab31ff2b05e1a73eabef02c8d9).

limpbrains (approved):

PR #837, opened 2023-01-18
feat: add safe mode

### LINKED ISSUE #566: Add Safe Mode
**Describe the feature/enhancement**
Add a safe mode in the event a user does not not want to connect to certain services or start LDK.

**Expected behavior**
Starting the app in a certain manner (shaking the phone, tapping the screen or holding the phone upside-down while starting) will start the app in safe mode.

pwltr:
Shake the phone at startup to enter recovery screen. Time window to enter is 3 seconds after splash screen, `startWalletServices` will be delayed until this timeout is over.

### Changes
- add safe mode
- small fix for "Export LDK logs" feature
- update `react-native-gesture-handler` and `react-native-reanimated` minor versions to fix a iOS build error with latest Ruby version
- ignore a `react-navigation` warning


pwltr:
Any other debug features we should add?

coreyphillips:
We may also want the ability to disable Backups, Bitcoin & Lightning (Disabling Bitcoin also disables Lightning). Though that may be a bit much for the first iteration. May make sense for the disable options to be a checkbox where we can disable what we wish and continue to the app.

I know @Jasonvdb had some thoughts regarding this as well.

Jasonvdb:
Nice! In the call we chatted about just allowing the user to display their seed (after auth with a pin or biometrics), and then sharing logs like you have there and maybe a support button also. If we disable features then the whole app behaves in weird ways.

So buttons are:
- Export lightning logs
- Display seed
- Contact support (Using `createSupportLink()`)
- Wipe app

src/screens/Settings/Lightning/Channels.tsx
```diff
@@ -219,20 +219,22 @@ const Channels = ({
...
 			});
 		}
 
-		// Share the zip file
-		await Share.open({
-			type: 'application/zip',
-			url: `file://${res.value}`,
-			title: 'Export Lightning Logs',
-		});
+		if (result.isOk()) {
```
Jasonvdb: A bit of a nitpick but if the error is checked above and there is a return in the body then you don't have to check isOk() again just below. Same thing but less code.

Jasonvdb:
In testing I see LDK is still started even if displaying safe mode which won't help if LDK is causing the crash

src/screens/Settings/Lightning/Channels.tsx
pwltr: Nitpicks welcome :). I thought there was an issue here where the error message got lost when early returning but seems to work fine when testing again.

coreyphillips:
Waiting 3-secs feels a bit long before starting services.
Could we catch the shake if using a 0.5 sec window instead? Just curious how low we can reliably go.

I know the original idea was to check if the device is upside-down on app-start using a library similar to [this](https://github.com/wonday/react-native-orientation-locker) (via its `getDeviceOrientation` method it will return `PORTRAIT-UPSIDEDOWN`). This would prevent us from having to wait for additional user input before starting and reduce the potential for scenarios where a car rumbling may be mistaken for a shake and trigger the safe-mode.

Jasonvdb:
Yeah agree with Corey here, artificial delay not going to be great here for initial impression of the app. Is the upside down thing a lot of work?

pwltr:
I tried to make it work using [react-native-orientation-locker](https://github.com/wonday/react-native-orientation-locker) but the issue is that it doesn't read the correct orientation on app start and always defaults to `PORTRAIT`. So the user would have to wait for the js to run to turn the phone which would be even more awkward than shaking.

Tested with a 500ms delay and seems to be reliable as well if you keep shaking the phone. If we need to we can apply the 
delay to the LDK setup functions only as a further optimisation. I don't think users will be fast enough to run into any problems with such a short delay.

coreyphillips:
We'll want to wrap Recovery mode in the `AuthCheck` component to ensure it can't be used to bypass previously enabled pin or biometrics.

coreyphillips (approved):

PR #838, opened 2023-01-19
fix(ui): fix electrum connectivity toast falsely triggering

pwltr:
Stop the Electrum connectivity toast falsely showing when the app comes back to the foreground by unsubscribing when the app goes to background.

### DIFF
--- a/src/AppOnboarded.tsx
+++ b/src/AppOnboarded.tsx
@@ -1,11 +1,11 @@
 import React, { memo, ReactElement, useEffect, useRef } from 'react';
 import { Platform, NativeModules, AppState } from 'react-native';
+import { useSelector } from 'react-redux';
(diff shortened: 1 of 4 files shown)


src/AppOnboarded.tsx
```diff
@@ -70,69 +68,65 @@ const AppOnboarded = (): ReactElement => {
...
+		);
+
 		return () => {
-			unsubscribeElectrum();
+			appStateSubscription.remove();
+			electrumSubscription.remove();
 		};
-	}, [isConnectedToElectrum]);
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, []);
```
limpbrains, line 131: If eslint-disable is really neede here, can you please leave a comment why?

limpbrains (approved):

PR #861, opened 2023-01-31
fix: use Modal as a backdrop white dragging widgets

limpbrains:
This allows to just tap outside to quit widgets editing mode
Also new icons


pwltr:
I'm getting a weird render issue when clicking on settings button:

https://user-images.githubusercontent.com/8538369/215761188-51a02d69-8e89-4d50-bf66-eefb8ad3422c.mp4

pwltr:
Also maybe disable the edit state when the user deletes a widget so this doesn't happen:


https://user-images.githubusercontent.com/8538369/215762749-e748996e-beb2-48a5-abff-e6147137fa91.mp4

limpbrains:
Thanks, @pwltr, I didn't test it properly.
I've fixed first issue, unfortinatly second one is more complicated - there are now 2 Modals shows at the same time. 
I will fix it in another PR, probably I will need to rewrite everything to avoid 2 Modals issue

pwltr:
Looks good now

pwltr (approved):

PR #863, opened 2023-01-31
fix(wallet): Address Indexes Update

coreyphillips:
This PR:
- Updates the `updateAddressIndexes` method by partitioning out each index condition.
- Changes iterator in `updateAddressIndexes` to a `for await of` loop.

### DIFF
--- a/src/store/actions/wallet.ts
+++ b/src/store/actions/wallet.ts
@@ -169,82 +169,82 @@ export const updateAddressIndexes = async ({
 		);
 	}
 
-	let updated = false;
-
-	const promises = addressTypesToCheck.map(async (addressTypeKey) => {
-		if (!selectedNetwork) {
-			selectedNetwork = getSelectedNetwork();
-		}
-		if (!selectedWallet) {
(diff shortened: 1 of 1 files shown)


src/store/actions/wallet.ts
```diff
@@ -169,82 +169,82 @@ export const updateAddressIndexes = async ({
...
-		const { type } = addressTypes[addressTypeKey];
-		let addressIndex = currentWallet.addressIndex[selectedNetwork][type];
-		let changeAddressIndex =
-			currentWallet.changeAddressIndex[selectedNetwork][type];
-		let lastUsedAddressIndex =
-			currentWallet.lastUsedAddressIndex[selectedNetwork][type];
-		let lastUsedChangeAddressIndex =
-			currentWallet.lastUsedChangeAddressIndex[selectedNetwork][type];
+	try {
+		for await (const addressTypeKey of addressTypesToCheck) {
```
limpbrains, line 173: I think if addressTypesToCheck is a regular array, then you don't need `for await of` to iterate over it, regular `for of` will do?

PR #867, opened 2023-02-01
fix(ui): fix scroll issues on Home screen

pwltr:
### Changes
- remove safe mode `<Pressable />` after timeout
- update Recovery text

### DIFF
--- a/src/screens/Recovery/Recovery.tsx
+++ b/src/screens/Recovery/Recovery.tsx
@@ -71,9 +71,9 @@ const Recovery = ({
 			<NavigationHeader title="Recovery" displayBackButton={false} />
 			<View style={styles.content}>
 				<Text01S style={styles.text} color="gray1">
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -94,14 +95,15 @@ const App = (): ReactElement => {
 	}, [tapCount, availableUpdateType, walletExists, requiresRemoteRestore]);
 
 	const currentTheme = useMemo(() => themes[theme], [theme]);
+	const RootWrapper = isListening ? Pressable : View;
```
limpbrains: Maybe instead of View, use React.Fragment, so that it will not add another layer to the view hierarchy ?

limpbrains (approved):

pwltr:
Have to avoid updating the render here to avoid flickering on startup. I went another route by disabling event propagation on the affected screen. Have to watch out for issues on other screens but so far seems good.

limpbrains (approved):

PR #871, opened 2023-02-03
fix(wallet): Add & Implement lodash.clonedeep

coreyphillips:
This PR:
- Adds `lodash.clonedeep` to dependencies in `package.json`.
- Implements `cloneDeep` on default wallet state references to protect against mutations when used. 
- Creates and implements `getDefaultWalletShape` where needed.
- Removes a lingering address type condition in `createDefaultWallet` method.
  - This condition was a remnant from when we tried to limit address generation to only a single address type when creating wallets for the first time. Removing this allows Bitkit to generate addresses for all address types when creating new wallets.

### DIFF
--- a/src/store/shapes/wallet.ts
+++ b/src/store/shapes/wallet.ts
@@ -14,12 +14,14 @@ import {
 import { IHeader } from '../../utils/types/electrum';
 import { EAvailableNetworks } from '../../utils/networks';
 import { objectKeys } from '../../utils/objectKeys';
+import cloneDeep from 'lodash.clonedeep';
 
-const cloneDeep = require('lodash.clonedeep');
+export const assetNetworks: Readonly<TAssetNetwork[]> = [
+	'bitcoin',
+	'lightning',
+];
 
-export const assetNetworks: TAssetNetwork[] = ['bitcoin', 'lightning'];
-
-export const addressTypes: IAddressTypes = {
+export const addressTypes: Readonly<IAddressTypes> = {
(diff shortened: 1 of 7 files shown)


src/store/shapes/wallet.ts
```diff
@@ -15,6 +15,8 @@ import { IHeader } from '../../utils/types/electrum';
 import { EAvailableNetworks } from '../../utils/networks';
 import { objectKeys } from '../../utils/objectKeys';
 
+const cloneDeep = require('lodash.clonedeep');
```
limpbrains: Why require, if we use import everywhere?

limpbrains (approved):

pwltr:
Could also wrap the type in question in a `Readonly` and let the compiler catch any reassignments.

`export const defaultWalletShape: Readonly<IWallet> = { ... }`
`export const defaultWalletStoreShape: Readonly<IWalletStore> = { ... }`

Don't think there is any need for cloning these objects to be honest.

pwltr:
This does contain a fix for restoring on Android though by removing the address type condition. Would be nice to get this in the next release.

coreyphillips:
I've wrapped the remainder wallet state references in the `Readonly` type. It may be a good idea to do the same for state references in other stores in a future PR as well.
While `Readonly` will help us catch mutating a state when we shouldn't be, it's not perfect and mutations can still happen. Cloning will ensure we can safely work with and use sensitive reference items.

PR #877, opened 2023-02-07
feat(ui): add deep linking for slashtags

### LINKED ISSUE #720: [Feature]: iOS/Android special url intent support
### Describe the problem

When tapping a `slash:` or `slashauth:` url, the OS should have these registered to open in Bitkit and automatically processed by the app. This should include any other special URLs we support, like bitcoin:, etc

### Describe the solution

Register all relevant special URL prefixes with the OS on app install or whenever/however is appropriate

### Additional context

We discussed this in the past, but I don't know if it ever got implemented.

pwltr:
closes https://github.com/synonymdev/bitkit/issues/720

### Changes
- add deep linking support for `slash:` & `slashauth:`

### Notes
Can test this on the simulator by running `npx uri-scheme open "<slashtag>" --[ios|android]`

### Preview

https://user-images.githubusercontent.com/8538369/224822007-635771c2-9b18-4542-9a9d-dd1024e4ffa0.mp4

### DIFF
--- a/src/navigation/root/RootNavigator.tsx
+++ b/src/navigation/root/RootNavigator.tsx
@@ -9,9 +9,9 @@ import {
 } from '@react-navigation/stack';
 
 import { NavigationContainer } from '../../styles/components';
+import { processInputData } from '../../utils/scanner';
 import Store from '../../store/types';
 import AuthCheck from '../../components/AuthCheck';
-import Biometrics from '../../components/Biometrics';
 import WalletNavigator from '../wallet/WalletNavigator';
 import ActivityDetail from '../../screens/Activity/ActivityDetail';
 import ActivityAssignContact from '../../screens/Activity/ActivityAssignContact';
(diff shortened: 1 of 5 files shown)


BitcoinErrorLog:
Note that I think there is more than one prefix for slashtags to register, check with @Nazeh, but I think slashauth: at least is another

src/navigation/root/RootNavigator.tsx
```diff
@@ -43,6 +43,20 @@ import ForceTransfer from '../bottom-sheet/ForceTransfer';
 import CloseChannelSuccess from '../bottom-sheet/CloseChannelSuccess';
 import type { RootStackParamList, RootStackScreenProps } from '../types';
 
+const linking = {
+	prefixes: ['slash://'],
```
Nuhvi: prefix is `slash:` not `slash://`

Nuhvi: Is that doable? does it _have_ to end in `://`?

BitcoinErrorLog:
I am pretty sure slashes are not required, since `bitcoin:` works as expected afaik.

limpbrains (approved):

PR #882, opened 2023-02-09
fix: various fixes

pwltr:
### Changes
- Suggestions: fix a crash when removing last card
- Electrum connectivity: when the connection was actually lost retry once, only notify if that retry fails 
- fixes the touch listener for safe mode stealing scroll event for some child components
- skip Profile onboarding in case the user already has a profile


coreyphillips:
LGTM, but it seems to be throwing off the e2e test.

coreyphillips (approved):

pwltr:
They were already failing before, I think they're just a little bit brittle still.

PR #896, opened 2023-02-18
feat(wallet): Add Minimum Transaction Relay Fee Dialog

coreyphillips:
This PR:
- Adds `Dialog` to `ReviewAndSend` component that alerts users when setting the min-relay fee.
- Sets `forceUpdate` param to true in `updateOnchainFeeEstimates` when starting app.

Notes:
- Bitkit will now grab fee estimates every time the app launches.
- Bitkit will now present a Dialog when users have set a fee that is at or below the current minimum tx relay fee.


limpbrains (approved):
I can see dialog logic is broken in ReviewAndSend component - it can't show 2 dialogs one after another, but this should be fixed in another PR, I think

PR #898, opened 2023-02-20
feat(wallet): Add Settings & Widgets To Backups

coreyphillips:
This PR:
- Backs up settings and widgets state when instructed to do-so via `remoteSettingsBackupSynced` & `remoteWidgetsBackupSynced`.
- Replaces JSON copy with `cloneDeep`.
- Reverted `recommendedBaseFee` back to `256`.

Notes:
- Two additional `useEffect`'s were added to the `backups-subscriber.ts` file.
   - One keeps an eye on `remoteSettingsBackupSynced` and the other keeps an eye on `remoteWidgetsBackupSynced`.
   - Whenever a `UPDATE_SETTINGS` or `SET_SLASHTAGS_FEED_WIDGET` action is dispatched the backup reducer will set `remoteSettingsBackupSynced` or `remoteWidgetsBackupSynced` to `false` respectively. This triggers a backup to occur for the given state.
- When importing and restoring settings/widgets backups:
   - `performSettingsRestore` & `performWidgetsRestore` methods were added.
   - When restoring settings/widgets, Bitkit will attempt to retrieve the backup information and then compare it to the current reference object for a partial match using the `testObjIsPartialMatch` method.
       - If the restored backup object is a partial match (Meaning, all restored object keys are found in Bitkit's current reference object), we will restore using the backup object. Otherwise, we discard it and use Bitkit's default reference object. This is to prevent falling into a corrupt state when backing up non-critical data such as settings/widgets. One improvement we can make in a future PR is to pull out all matching keys and use what we can from that to proceed with a backup instead of simply discarding the entire object.
- In a future PR we may want to limit how frequently these types of backups occur since they are non-critical. They currently fire off whenever a setting is changed or widget added. If the frequency of these backups become a problem, one idea would be to limit these backups to only occur when the app starts up.

To Test:
- Create a new wallet.
- Backup the seed.
- Add a widget or two.
- Change a few settings.
- Give it some time by navigating around the wallet for a bit.
- Wipe the wallet.
- Attempt to restore the wallet.
- Observe that most, if not all, of your added widgets and settings have been restored.


src/utils/helpers.ts
```diff
@@ -307,6 +307,35 @@ export const objectsMatch = (obj1, obj2): boolean => {
...
 };
 
+/**
+ * Determines if all keys in the test object are found in the reference object.
+ * @param testObj
+ * @param referenceObj
+ * @param {string[]} [keysToExclude] Returns true when excluded object keys are encountered.
+ * @returns boolean
+ */
+export const testObjIsPartialMatch = (
```
pwltr: Nit: maybe rename this to `isObjPartialMatch` or similar to avoid confusion with actual tests.

Also I think a unit test for this would be good.

pwltr (changes requested):

pwltr:
Restoring settings worked, but I couldn't get it to restore my widgets yet.

src/utils/helpers.ts
coreyphillips: Renamed [here](https://github.com/synonymdev/bitkit/pull/898/commits/0c89ef1628729b57da4446bd9613ba37425b1791).

coreyphillips:
> Restoring settings worked, but I couldn't get it to restore my widgets yet.

Should be resolved as of [this commit ](https://github.com/synonymdev/bitkit/commit/0a10bd558706a0f93bac2912f5020e43e00a8e48).

pwltr (approved):

PR #902, opened 2023-02-22
fix: disable resolving profiles by default in useProfile hook

### LINKED ISSUE #892: [Bug]: Opening the contacts view after tapping on the contacts icon is still a bit sluggish
### Describe the bug

UI is blocking for half a second before switching to the Contacts view.

### Reproduce

1. Go to home screen
2. Click on contacts icon top right

### Screenshots / Recording

_No response_

### Operating system

Android, IOS

### Bitkit version

1.0.0-beta.35

### Log output

_No response_

Nuhvi:
Resolving profiles seem to block UI, which is most impactful in the Contacts List screen, disabling it by default, and enabling it while adding new contacts reduces that lag significantly

More optimizations might be needed, but I would rather test this minimum change independently from any wider refactoring.

closes #892 

### Before

https://user-images.githubusercontent.com/40009100/220600392-6824b22f-ebf9-41e2-b418-3976b7cb0caa.mov

### After

https://user-images.githubusercontent.com/40009100/220600525-d371b529-f7d9-449f-8a19-73224dfd0960.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@synonymdev/react-native-lnurl": "0.0.3",
     "@synonymdev/result": "0.0.2",
     "@synonymdev/slashtags-auth": "^1.0.0-alpha.5",
-    "@synonymdev/slashtags-sdk": "1.0.0-alpha.35",
+    "@synonymdev/slashtags-sdk": "^1.0.0-alpha.36",
     "assert": "^2.0.0",
     "backpack-client": "github:synonymdev/bitkit-backup-client#f08fdb28529d8a3f8bfecc789443c43b966a7581",
     "bip21": "^2.0.3",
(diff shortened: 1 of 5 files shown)


pwltr (approved):

aldertnl:
@Nazeh _'Resolving profiles seem to block UI, which is most impactful in the Contacts List screen, disabling it by default, and enabling it while adding new contacts reduces that lag significantly'_

 - Does that mean that after a wallet restore these contacts would never resolve? And that contacts would never update in general? Not sure is that is an ideal solution. Maybe we could load the Contacts list screen UI, and after that populate the contacts data (resolve these) asynchronously.

Nuhvi:
@aldertnl It will resolve if you go to their contact page, not in the contact list.

This can be solved easily by resolving profiles once after restoring from the backup. But it is not the case now!

Nuhvi:
@aldertnl the correct solution here is to backup the cached profiles in the Redux store, and restore that when the app is restored.

@coreyphillips what is the status there and do you need help?

PR #922, opened 2023-03-08
fix(wallet): Various Fixes

coreyphillips:
This PR:
- Updates replace logic in `scanner.ts` to catch case insensitive instances of "lightning:".
- Added `channelIsOpen` condition to the "Close Connection" button in `ChannelDetails.tsx`.

### DIFF
--- a/src/screens/Settings/Lightning/ChannelDetails.tsx
+++ b/src/screens/Settings/Lightning/ChannelDetails.tsx
@@ -30,7 +30,10 @@ import {
 	selectedWalletSelector,
 } from '../../../store/reselect/wallet';
 import { enableDevOptionsSelector } from '../../../store/reselect/settings';
-import { openChannelIdsSelector } from '../../../store/reselect/lightning';
+import {
+	channelIsOpenSelector,
+	openChannelIdsSelector,
+} from '../../../store/reselect/lightning';
 import { getStateMessage } from '../../../utils/blocktank';
 import {
 	ArrowCounterClock,
@@ -41,6 +44,7 @@ import {
 	XIcon,
 } from '../../../styles/icons';
 import type { SettingsScreenProps } from '../../../navigation/types';
+import { useAppSelector } from '../../../hooks/redux';
 
 export const getOrderStatus = (state: number): React.FC<SvgProps> => {
 	// possible order states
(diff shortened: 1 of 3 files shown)


src/utils/helpers.ts
```diff
@@ -251,6 +251,10 @@ export const truncate = (text: string, length: number): string => {
  * @returns {string}
  */
 export const ellipse = (text: string, length: number = 15): string => {
```
pwltr: This function is actually quite broken. Replaced [here](https://github.com/synonymdev/bitkit/pull/923)

src/screens/Settings/Lightning/ChannelDetails.tsx
```diff
@@ -191,6 +195,16 @@ const ChannelDetails = ({
 		return openChannelIdsSelector(state, selectedWallet, selectedNetwork);
 	});
 
+	const channelIsOpen = useAppSelector((state) => {
+		const channelId = channel?.channel_id ?? ''; //If a channel is not confirmed it's likely a channel id is not yet assigned.
```
pwltr: This may have to be updated later when we allow third-party channels but for now the `channel_id` will always be defined and in case of an unconfirmed blocktank channel will be the `orderId`. See https://github.com/synonymdev/bitkit/blob/b59b3505383afa594eb43e6be564519afc9de42e/src/screens/Settings/Lightning/Channels.tsx#L86-L88

Can also use `channel.is_usable` to determine if the channel is open I think?

pwltr (changes requested):

src/screens/Settings/Lightning/ChannelDetails.tsx
coreyphillips: In lightning, when a channel is first opened a temporary channel ID is assigned. Only once the funding transaction confirms is a proper channel id (`channel_id`) assigned. react-native-ldk should handle this so we never have an issue where `channel_id` is undefined, but this is me being cautious with LDK's behavior over time.

Regarding `channel.is_usable`, there are a number of reasons a channel can be unusable where we wouldn't necessarily want to block the user from being able to close a channel. That why referencing `openChannelIds` directly is used instead.

src/utils/helpers.ts
coreyphillips: Nice, reverted this change for your fix [here](https://github.com/synonymdev/bitkit/commit/70a9353771e8bca6e6546422c0ae3cc9eb1b74d2).

src/screens/Settings/Lightning/ChannelDetails.tsx
pwltr: In that case we should update the type for `TChannel[channel_id]` to reflect this.

coreyphillips: It wouldn't be the intended behavior though. This is just me being paranoid.
I gave react-native-ldk a check. If this behavior does change, we'll know pretty quickly in Bitkit and will be able to update our types/implementations accordingly.
I removed the channel_id check [here](https://github.com/synonymdev/bitkit/commit/e0d752e401ed21624df935df81b6a49cb4456e15).

pwltr: Sounds good.

pwltr (approved):

PR #924, opened 2023-03-09
chore: remove postinstall-slashtags.js

Nuhvi:
- remove postinstall-slashtags.js
- add transformBigIntLiteral babel transformer
- add @babel/plugin-proposal-async-generator-functions
- update slashtags-sdk to fix  import issue

outstanding issue: Object is not iterable in Hyperswarm/peer-discovery.js apparently babel didn't transform that file?


Nuhvi:
@limpbrains Any help here would be really appreciated, I tried to clean up all the hacky approaches, but babel doesn't seem to do what it should and transform `for await ()` in submodules.

Nuhvi:
The error I get, curiously
simulator_screenshot_764F793B-D1C0-459C-B77E-BE88C7E09FBC

limpbrains:
I believe babel did transform the file.
Error happends because in `async _refresh` method there is an attempt to iterate `for await (const data of this._activeQuery)` where `this._activeQuery` is not an iterable. 

`this._activeQuery[Symbol.asyncIterator]` and `this._activeQuery[Symbol.iterator]` are undefined, which should not be for true for objects that you can iterate over.

That is why `Object is not async iterable` is thrown.

I believe this is a bug in hyperswarm

Nuhvi:
@limpbrains no it is not a bug, chances are, rn-nodify or shims or whatever; something missed with the Readable stream somehow ... I have been fighting this since last year until I finally gave up and went with the postInstall-slashtags solution.

limpbrains:
Maybe we should wait for RN 0.70, before merging this ? cc @coreyphillips

coreyphillips:
Agreed, it may be best to hold off on troubleshooting this until we upgrade RN to the latest version. This may eliminate some of the issues we're currently experiencing. And/Or add some :-)

Nuhvi:
@coreyphillips @limpbrains I largely agree, but, notice that this simplifies and removes Slasthags hacks (less opaque and brittle things to break on a sudden change), so if the RN update caused any issues with the current hacks, it is worth trying the new RN with the simplifications in this PR.

limpbrains (approved):
`@babel/plugin-proposal-async-generator-functions` is already in master and in next RN branch, so I don't think anything can be broken by this PR.

PR #926, opened 2023-03-09
fix(widgets): better default decoder to handle Objects and stringifie…

Nuhvi:


Nuhvi:
this will never end apparently, maybe give it a review already? @limpbrains @pwltr

pwltr (approved):

PR #928, opened 2023-03-13
fix: Lightning UI & channel open/close issues

### LINKED ISSUE #659: [Bug]: Balance not adding up, too-high Blocktank fees, Blocktank takes many attempts
### Describe the bug

A few possibly related bugs:

Part 1 - "It looks like 14,793 Sats have gone missing somehow in setting up lightning. These numbers don't add up. I sent myself (received) 57,840. Then set up lightning. Now I only have 43,047 left. Where is the other 14,793 sats? If they went for fees then it needs to explain that somewhere. "

Part 2 - "Note: It costs $123 to add $2 to my lightning channel."

Part 3 - "It took 9 attempts over 2 days to get the Lightning channel open. It kept throwing an error about outputs exceeding inputs." 

### Screenshots / Recording

image

image

image

### Operating system

Android?

### Bitkit version

_No response_

### Log output

_No response_

### LINKED ISSUE #872: [Bug]: Text parser showing 0 as amount in LN invoice payment attempts
### Describe the bug

If I scan a QR code of a LN invoice, it shows the correct amount and I can attempt to pay it.

If I paste in a text version of the same invoice instead, it shows the amount as zero (amount is also not editable).

### Screenshots / Recording

_No response_

### Operating system

iOS

### Bitkit version

33

### Log output

_No response_

pwltr:
Closes https://github.com/synonymdev/bitkit/issues/659, https://github.com/synonymdev/bitkit/issues/872, [Asana #1](https://app.asana.com/0/0/1202464022210897/1203975785881699/f), [Asana #2](https://app.asana.com/0/1189635884316057/1204116236827217/f)

### Changes
- UI: fix spacing inconsistency in List views / rows
- UI: fix some style in LN onboarding & transfer, fix small design deltas for widgets
- Send: fix missing redirect when pasting non BIP21 invoices
- Settings: fix some issues in the Channel UI (list & detail screen)
- Lightning: fix issues when closing channels, fix suggestion cards when transferring
- Activity: show transfer to savings tx, hide receive notification when it's a transfer
- Receive: add waitForLdk helper, fix LN invoice missing when opening Receive sheet too quickly after opening Bitkit
- Lightning: improve fee setting for LDK
- remove some unnecessary async code, cleanup

### Notes
We're now adding the tx inputs to the store so that we don't have to fetch them when needed. Using `order.channel_close_tx` from BT to determing correct channel status and to show transfers to savings in activity.


src/utils/i18n/locales/ru/lightning.json
```diff
@@ -87,6 +87,7 @@
 	"total_size": "Общий размер канала",
 	"other": "Другое",
 	"opened_on": "Открыт",
+	"closed_on": "TODO",
```
pwltr: @limpbrains Okay to say 'TODO' here? Is there a better way to add new fields without translation?

limpbrains: You don't have to add anything to the translation files, except for the English.
If some string is missing in the RU translation, one from EN will be used instead automatically.
Please remove this

src/screens/Wallets/Receive/index.tsx
```diff
@@ -125,6 +125,8 @@ const Receive = ({
 			return;
 		}
 
+		await waitForLdk();
```
limpbrains, line 128: If LDK is not available, will user see any error?

limpbrains (changes requested):

src/screens/Wallets/Receive/index.tsx
pwltr: No not at the moment, when testing it was usually available on the second or third retry. I would like to add some error handling for this but we should think of a good approach first to give the user some action to follow if LDK doesn't start at all. (Will be needed in other places too). Got some stability improvements coming as well in RN-LDK afaik.

coreyphillips (approved):

limpbrains (approved):

PR #934, opened 2023-03-15
fix(wallet): #780 skipping addressIndex

### LINKED ISSUE #780: [Bug]: Bitkit misses address indexes when switching networks
### Describe the bug

After switching wallet network addresses at one or more indexes are missing. Unused address indexes are skipped when generating addresses when switching networks.

### Reproduce

1. Create new wallet
2. Switch to regtest
3. Check Redux store wallet -> wallets -> wallet0 -> addresses -> bitcoinRegtest -> p2wpkh

Addresses should start from index 0, but they starts with 1:
```
1a42c8f5e92577267cc1595c53d5f476f4b32a8f9d0f391cfc0e4a5b689851d4: {
	address: "bcrt1qxsvz97cp99ll3d3p55nfvs4ynwf2vuzw5wudcy"
	index: 2
	path: "m/84'/0'/0'/0/2"
	publicKey: "03a8f6bb568e2ac4135d0f7501e664625c0988983a632c3503f0e28ce4bfc957dd"
	scriptHash: "1a42c8f5e92577267cc1595c53d5f476f4b32a8f9d0f391cfc0e4a5b689851d4"
},
222088841d13b644374c36420a27bd2bee613c97f43c725618c55a65116cb696: {
	address: "bcrt1q389nnnnua2q9d8h48asn5fvkugumk3dh74vthn"
	index: 4
	path: "m/84'/0'/0'/0/4"
	publicKey: "02432d476301b983e7096c5cf481706722c8d146f305c648375e44fae0260a6fe0"
	scriptHash: "222088841d13b644374c36420a27bd2bee613c97f43c725618c55a65116cb696"
},
a517806b70620a720fe571e3449176092876e6575a3b54d9fc01223db533fbbb: {
	address: "bcrt1qg7hrrg0swlh6mepm7r6cjy6gkr3zr9pwtcxekd"
	index: 1
	path: "m/84'/0'/0'/0/1"
	publicKey: "025d38a2e7f02a7448072bf34e53b28e78c23f48c56f322d398e5b48ac97a57552"
	scriptHash: "a517806b70620a720fe571e3449176092876e6575a3b54d9fc01223db533fbbb"
},
bc9e5d449d70377b350ec2cb45228bb22d91f04516cddce8c0db57c248e9743c: {
	address: "bcrt1q5cmptwr4gh0d78uqv8v06tr3w5suuv2y44a9v6"
	index: 5
	path: "m/84'/0'/0'/0/5"
	publicKey: "0216503106f7f0efe34569548f387cd07d9726dc654eb333925b7ed4883623c809"
	scriptHash: "bc9e5d449d70377b350ec2cb45228bb22d91f04516cddce8c0db57c248e9743c"
},
ca28648ab53e145c9835504c974293ceea4257fd1d7da93b2fdad66dcc3448d8: {
	address: "bcrt1qvd47eknexlkgrumgzuvad893qy0elj8257wqq9"
	index: 3
	path: "m/84'/0'/0'/0/3"
	publicKey: "03d1c3dd66f91343f255b7d0c5cc54fed0e0497c3ed9853eccef7c187658741d1b"
	scriptHash: "ca28648ab53e145c9835504c974293ceea4257fd1d7da93b2fdad66dcc3448d8"
}
```

### Screenshots / Recording

_No response_

### Operating system

iOS

### Bitkit version

_No response_

### Log output

_No response_

pwltr:
Closes https://github.com/synonymdev/bitkit/issues/780

~This PR fixes some incorrect usages of `.map()` together with `async` that most likely caused a race condition when generating addresses causing the above issue. Also removes other unnecessary async code.~

~I'll probably enable the `@typescript-eslint/no-misused-promises` eslint rule and fix other occurrences of this type in another PR.~

This PR fixes a race condition caused by calling `updateAddressIndexes()` twice when switching networks causing the above issue. Also removes other unnecessary async code.

### Changes
- remove all function calls in `BitcoinNetworkSelection.tsx` that are already called in `startWalletServices()`
- remove unnecessary async code, cleanup

### DIFF
--- a/src/store/actions/wallet.ts
+++ b/src/store/actions/wallet.ts
@@ -166,14 +166,10 @@ export const updateAddressIndexes = async ({
 		selectedWallet,
 		selectedNetwork,
 	});
-	const addressTypes = getAddressTypes();
-	let addressTypesToCheck = Object.keys(addressTypes) as EAddressType[];
+
+	let addressTypesToCheck = Object.keys(getAddressTypes()) as EAddressType[];
 	if (addressType) {
-		addressTypesToCheck = await Promise.all(
-			addressTypesToCheck.filter(
-				(_addressType) => _addressType === addressType,
-			),
-		);
+		addressTypesToCheck = [addressType];
 	}
 
 	let updated = false;
@@ -193,42 +189,40 @@ export const updateAddressIndexes = async ({
 		if (response.isErr()) {
 			throw response.error;
 		}
+		const result = response.value;
 
-		const { type } = addressTypes[addressTypeKey];
-		let addressIndex = currentWallet.addressIndex[selectedNetwork][type];
+		let addressIndex =
+			currentWallet.addressIndex[selectedNetwork][addressTypeKey];
 		let changeAddressIndex =
-			currentWallet.changeAddressIndex[selectedNetwork][type];
+			currentWallet.changeAddressIndex[selectedNetwork][addressTypeKey];
 		let lastUsedAddressIndex =
-			currentWallet.lastUsedAddressIndex[selectedNetwork][type];
+			currentWallet.lastUsedAddressIndex[selectedNetwork][addressTypeKey];
 		let lastUsedChangeAddressIndex =
-			currentWallet.lastUsedChangeAddressIndex[selectedNetwork][type];
+			currentWallet.lastUsedChangeAddressIndex[selectedNetwork][addressTypeKey];
 
 		if (
-			currentWallet.addressIndex[selectedNetwork][type]?.index < 0 ||
-			currentWallet.changeAddressIndex[selectedNetwork][type]?.index < 0 ||
-			response.value?.addressIndex?.index >
-				currentWallet.addressIndex[selectedNetwork][type]?.index ||
-			response.value?.changeAddressIndex?.index >
-				currentWallet.changeAddressIndex[selectedNetwork][type]?.index ||
-			response.value?.lastUsedAddressIndex?.index >
-				currentWallet.lastUsedAddressIndex[selectedNetwork][type]?.index ||
-			response.value?.lastUsedChangeAddressIndex?.index >
-				currentWallet.lastUsedChangeAddressIndex[selectedNetwork][type]?.index
+			addressIndex.index < 0 ||
+			changeAddressIndex.index < 0 ||
+			result.addressIndex.index > addressIndex.index ||
+			result.changeAddressIndex.index > changeAddressIndex.index ||
+			result.lastUsedAddressIndex.index > lastUsedAddressIndex.index ||
+			result.lastUsedChangeAddressIndex.index >
+				lastUsedChangeAddressIndex?.index
 		) {
-			if (response.value?.addressIndex) {
-				addressIndex = response.value.addressIndex;
+			if (result.addressIndex) {
+				addressIndex = result.addressIndex;
 			}
 
-			if (response.value?.changeAddressIndex) {
-				changeAddressIndex = response.value?.changeAddressIndex;
+			if (result.changeAddressIndex) {
+				changeAddressIndex = result.changeAddressIndex;
 			}
 
-			if (response.value?.lastUsedAddressIndex) {
-				lastUsedAddressIndex = response.value.lastUsedAddressIndex;
+			if (result.lastUsedAddressIndex) {
+				lastUsedAddressIndex = result.lastUsedAddressIndex;
 			}
 
-			if (response.value?.lastUsedChangeAddressIndex) {
-				lastUsedChangeAddressIndex = response.value?.lastUsedChangeAddressIndex;
+			if (result.lastUsedChangeAddressIndex) {
+				lastUsedChangeAddressIndex = result.lastUsedChangeAddressIndex;
 			}
 
 			dispatch({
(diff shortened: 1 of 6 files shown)


src/utils/wallet/index.ts
```diff
@@ -216,93 +215,91 @@ export const generateAddresses = async ({
...
+				const addressPath = formatKeyDerivationPath({
+					path,
+					selectedNetwork,
+					accountType,
+					changeAddress: false,
+					addressIndex: `${index}`,
+				});
+				if (addressPath.isErr()) {
+					return err(addressPath.error.message);
+				}
```
limpbrains: If you return in map, you only skip one element
If you return in for loop, you exit from it
IMHO here should be return, but let's check if @coreyphillips

src/store/actions/wallet.ts
```diff
@@ -238,17 +225,12 @@ export const updateAddressIndexes = async ({
...
+
 			updated = true;
 		}
-	});
-
-	try {
-		await Promise.all(promises);
-	} catch (e) {
-		return err(e);
 	}
```
limpbrains: Not sure if this can be removed without adding giant try/catch for the whole function

src/utils/wallet/index.ts
```diff
@@ -2738,18 +2724,18 @@ export const createZeroIndexAddresses = async ({
...
+	for (const addressType of addressTypes) {
+		await addAddresses({
+			selectedWallet,
+			selectedNetwork,
+			addressAmount: 1,
+			changeAddressAmount: 1,
+			addressIndex: 0,
+			changeAddressIndex: 0,
+			addressType,
+		});
```
limpbrains: Same here. With promise.all if any `addAddresses` call throws an error, it will be ignored. With for ... loop this will stop the loop.

IMHO for loop is better here, but what if `addAddresses` return an error? Should we handle stop the loop or ignore it?

limpbrains (changes requested):
I think that using Promise.all with map to just iterate over something is very confusing and may lead to race condition bugs.

src/store/actions/wallet.ts
```diff
@@ -169,32 +169,22 @@ export const updateAddressIndexes = async ({
...
-		}
+	for (const type of addressTypesToCheck) {
 		const response = await getNextAvailableAddress({
 			selectedWallet,
 			selectedNetwork,
-			addressType: addressTypeKey,
+			addressType: type,
 		});
 		if (response.isErr()) {
 			throw response.error;
```
limpbrains, line 190: replace with return

src/utils/wallet/index.ts
pwltr: I think we should definitely let it error out and not ignore because it's critical that this does what it should. But need to handle errors here.

src/store/actions/wallet.ts
pwltr: This code was redundant since `promises` is of type `Promise<void>` since the `.map()` didn't return anything. I think with [this](https://github.com/synonymdev/bitkit/pull/934#discussion_r1137383518) changed there isn't any need for more error handling here? Other than `getNextAvailableAddress` (handled) there isn't a lot going on.

pwltr: Apparently you don't need to return anything in async `.map()`s for it to run as intended. So some of these changes are unnecessary.

pwltr:
Found the problem is not caused by the promises being resolved in parallel but rather the whole `updateAddressIndexes()` function being called twice racing against each other. I still believe error handling should be improved in some of the async/await loops but that would be for another PR.

limpbrains:
How is this possible? We wait `updateAddressIndexes` to be completed before calling `startWalletServices`. 
If `await updateAddressIndexes()` resolves before the job is completed, there may be a bug in this function. Have we resolved this issue?

As far as I recall, the purpose of calling updateAddressIndexes here is to generate more addresses than what is typically done in startWalletServices. By default, startWalletServices only generates five native segwit addresses, similar to when creating a new wallet. However, when switching networks or restoring a wallet, it is necessary to scan for alternative wallet types that the user may have used. If they have used a wallet type other than native segwit, they will be unable to access their funds.

pwltr:
Good question, let me check again.

src/utils/wallet/index.ts
coreyphillips: We should be most careful about exiting from the loop entirely when iterating over any method performing calls to electrum. This could result in unsaved work.

src/store/actions/wallet.ts
```diff
@@ -166,14 +166,10 @@ export const updateAddressIndexes = async ({
...
+
+	let addressTypesToCheck = Object.keys(getAddressTypes()) as EAddressType[];
 	if (addressType) {
-		addressTypesToCheck = await Promise.all(
-			addressTypesToCheck.filter(
-				(_addressType) => _addressType === addressType,
-			),
-		);
+		addressTypesToCheck = [addressType];
 	}
```
pwltr, line 173: return value of `.filter()` was not used.

src/utils/wallet/index.ts
```diff
@@ -853,340 +848,328 @@ export const getNextAvailableAddress = async ({
...
-					selectedNetwork,
-					addressAmount: GENERATE_ADDRESS_AMOUNT,
-					changeAddressAmount: 0,
-					keyDerivationPath,
-					addressType,
-				});
-				if (generatedAddresses.isErr()) {
-					return resolve(err(generatedAddresses.error));
-				}
-				const key = Object.keys(generatedAddresses.value.addresses)[0];
```
pwltr, line 910: It was wrongly assumed that the `addresses` object was sorted by `address.index` here. In effect `key` pointed to a random address of the newly generated.

src/utils/wallet/index.ts
```diff
@@ -853,340 +848,328 @@ export const getNextAvailableAddress = async ({
...
-					addressAmount: 0,
-					changeAddressAmount: GENERATE_ADDRESS_AMOUNT,
-					keyDerivationPath,
-					addressType,
-				});
-				if (generatedChangeAddresses.isErr()) {
-					return resolve(err(generatedChangeAddresses.error));
-				}
-				const key = Object.keys(
-					generatedChangeAddresses.value.changeAddresses,
```
pwltr, line 927: same here

limpbrains (approved):

PR #950, opened 2023-03-22
fix: Update Crash Screen

pwltr:
### Changes
- update ErrorBoundary screen
- update `createSupportLink` function
- add `react-native-exit-app` dependency
- also adds a fix for the UpdatePrompt triggering when it shouldn't

### Preview

https://user-images.githubusercontent.com/8538369/226980626-fac03499-0225-48e1-8402-6a8a002ba880.mp4

### DIFF
--- a/Root.tsx
+++ b/Root.tsx
@@ -27,14 +27,8 @@ const Root = (): ReactElement => {
 		</Provider>
 	);
 
-	if (__DEV__) {
-		return content;
-	}
-
 	return <ErrorBoundary>{content}</ErrorBoundary>;
 };
-
-export default Root;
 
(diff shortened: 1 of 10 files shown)


limpbrains (approved):

limpbrains:
we need an e2e test for that :)

Also , should we use react-native-exit-app in wipe action?

pwltr:
> we need an e2e test for that :)
> 
> Also , should we use react-native-exit-app in wipe action?

Agreed, I need take some time to look into writing detox tests. `react-native-exit-app` is only for closing the app, I think in the wipe action what we want is a restart?

PR #953, opened 2023-03-23
test(settings): add tests for dev settings & crash screen

pwltr:
### Changes
- add tests for dev settings & crash screen
- add env var for running e2e tests in development to be used in tests


pwltr:
@limpbrains Please review docs for `waitFor` detox function [here](https://github.com/nubank/detox/blob/master/docs/APIRef.waitFor.md). I think there are some changes needed in the tests.

> NOTE: Every waitFor call must set a timeout using withTimeout(). Calling waitFor without setting a timeout will do nothing.

> NOTE: waitFor will not throw when reaching timeout, instead it will just continue to the next line. To make sure your tests work as you expect them to add expect() at the following line.

For example this should fail (the client is not disconnected but stays connected to the current server in the test)

```js
// disconnected warning should appear
await waitFor(element(by.id('Disconnected'))).toBeVisible();
await sleep(1000);
```

limpbrains:
> @limpbrains Please review docs for `waitFor` detox function [here](https://github.com/nubank/detox/blob/master/docs/APIRef.waitFor.md). I think there are some changes needed in the tests.
> 
> > NOTE: Every waitFor call must set a timeout using withTimeout(). Calling waitFor without setting a timeout will do nothing.
> 
> > NOTE: waitFor will not throw when reaching timeout, instead it will just continue to the next line. To make sure your tests work as you expect them to add expect() at the following line.
> 
> For example this should fail (the client is not disconnected but stays connected to the current server in the test)
> 
> ```js
> // disconnected warning should appear
> await waitFor(element(by.id('Disconnected'))).toBeVisible();
> await sleep(1000);
> ```

😱😱😱 thx, will do

limpbrains (approved):
because these 2 tests affects each other and you can't choose tests order in jest, I suggest we merge them into one.

PR #954, opened 2023-03-24
feat(wallet): Implement Wallet Checks

coreyphillips:
This PR:
- Implements wallet/app checks on app-start.
- Adds address storage check.
- Creates `checks` reducer containing a `warnings` array.

Notes:
- This PR sets the app up to check for and address known issues and locally store warnings and data associated with a particular warning for potential troubleshooting and debugging.
- Each warning id is associated with a particular Github issue.
- The check implemented in this PR checks the lowest and highest stored addresses in storage and compares them with newly generated addresses to ensure they were not impacted by a memory bug. If a discrepancy is found, the app compares all addresses and change addresses against newly generated addresses and replaces all impacted addresses and change addresses in storage accordingly. The app will then take note of this by adding an entry to the warnings array along with any data that may come in handy for troubleshooting/debugging at a later date.
- In a future PR, we'll want to add a way to export data saved to the warnings array for troubleshooting/debugging.
- In a future PR, we'll want to add a notification to alert the user of any warnings that can not be addresses and may be cause for concern.


src/utils/wallet/index.ts
```diff
@@ -222,7 +233,11 @@ export const generateAddresses = async ({
 			addressArray.map(async (_item, i) => {
 				try {
 					const index = i + addressIndex;
-					let path = { ...keyDerivationPath };
+					// This check isn't needed, but satisfies a ts bug.
+					if (!keyDerivationPath) {
+						return err('');
+					}
+					const path = { ...keyDerivationPath };
```
pwltr: This also works:

```js
const path = { ...keyDerivationPath! };
```

or a bit safer:

```js
// outside callback
const constKeyDerivationPath = keyDerivationPath
```

__tests__/wallet-send.ts
```diff
@@ -95,6 +96,16 @@ describe('Wallet - new wallet, send and receive', () => {
...
 
+		// run storage check
+		res = await runStorageCheck({
+			selectedWallet: 'wallet0',
+			selectedNetwork: 'bitcoinRegtest',
+		});
+		if (res.isErr()) {
+			throw res.error;
+		}
+		expect(typeof res.value).toEqual('All Match');
```
pwltr:
```suggestion
		expect(res.value).toEqual('All Match');
```

src/store/actions/wallet.ts
```diff
@@ -1584,3 +1585,105 @@ export const setZeroIndexAddresses = async ({
...
+				// Remove invalid address
+				delete newAddresses[addressType][invalidScriptHash];
+				// Add valid address data
+				newAddresses[addressType][validScriptHash] =
+					impactedAddress.generatedAddress;
+				// Update address index info
+				const currentAddressIndex = newAddressIndexes[addressType].index;
+				if (currentAddressIndex === impactedAddress.storedAddress.index) {
+					newAddressIndexes[addressType] = impactedAddress.generatedAddress;
+				}
```
pwltr: Will this not mutate state? Recommend using `.filter()` instead or cloning objects above.

pwltr (changes requested):

src/utils/wallet/index.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/commit/32b6eb2da64fa08fe4fe2a75fdaab9927aaff079).

__tests__/wallet-send.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/commit/dc8a615b33a764b0261a7220f6a86878cca72a54).

src/store/actions/wallet.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/commit/fc308ac80eef47a9fb9e09c6df255ff63f4d0fa2).

src/store/actions/wallet.ts
```diff
@@ -1584,3 +1585,111 @@ export const setZeroIndexAddresses = async ({
...
+					newAddresses[addressType],
+				);
+				// Add valid address data
+				newAddresses[addressType][validScriptHash] =
+					impactedAddress.generatedAddress;
+				// Update address index info
+				const currentAddressIndex = newAddressIndexes[addressType].index;
+				if (currentAddressIndex === impactedAddress.storedAddress.index) {
+					newAddressIndexes[addressType] = impactedAddress.generatedAddress;
+				}
```
pwltr, line 1648: This still looks like a mutation to me since `newAddresses` is a reference. But maybe it's fine because somewhere up the flow there is this:

```
export const getWalletStore = (): IWalletStore => {
	return cloneDeep(store.getState()[EStore.wallet]);
};
```

Also there is already a `removeKeysFromObject` if you wanna use that :)

pwltr (approved):

PR #956, opened 2023-03-28
feat: Activity calendar filter

limpbrains:
- Activity can be filtered by date using calendar
- calendar generator aware of user timezone and locale (week starts with Sunday for en-US)
- e2e tests Tags and Activity filter
- react-native-localize to get timezone from the device
- there is still bug with 2 tags in activity filter, but this will be resolved later


src/components/TabBar.tsx
```diff
@@ -23,6 +27,13 @@ const TabBar = ({
 	const insets = useSafeAreaInsets();
 	const { t } = useTranslation('wallet');
 	const betaRiskAccepted = useAppSelector(betaRiskAcceptedSelector);
+	const viewControllers = useAppSelector(viewControllersSelector);
+	const anyBottomSheetIsOpen = useMemo(() => {
+		const viewControllerKeys = objectKeys(viewControllers);
+		return viewControllerKeys
+			.filter((view) => view !== 'backupPrompt')
```
pwltr: remove exception for `backupPrompt`

src/store/migrations/index.ts
```diff
@@ -85,6 +85,17 @@ const migrations = {
...
+		return {
+			...state,
+			ui: {
+				...state.ui,
+				viewControllers: defaultViewControllers,
+				timeZone: 'UTC',
+				language: 'en',
+			},
+		};
+	},
```
pwltr: No need for a migration here since `ui` reducer is not persisted. See `persistConfig`

src/screens/Activity/ActivityFiltered.tsx
```diff
@@ -77,7 +87,10 @@ const ActivityFiltered = ({
 		};
 	}, [radiusContainerHeight, insets.bottom]);
 
-	const addTag = (tag): void => setTags((tg) => [...tg, tag]);
+	const addTag = (tag): void => {
```
pwltr:
```suggestion
	const addTag = (tag: string): void => {
```

src/screens/Activity/TagsPrompt.tsx
```diff
@@ -0,0 +1,87 @@
...
+} from '../../hooks/bottomSheet';
+import Tag from '../../components/Tag';
+import { useSelector } from 'react-redux';
+import Store from '../../store/types';
+
+const TagsPrompt = ({
+	onAddTag,
+	tags,
+}: {
+	onAddTag: Function;
```
pwltr:
```suggestion
	onAddTag: (tag: string) => void;
```

src/components/SearchInput.tsx
```diff
@@ -39,7 +42,7 @@ const styles = StyleSheet.create({
 		alignItems: 'center',
 		borderRadius: 32,
 		backgroundColor: 'rgba(255, 255, 255, 0.08)',
-		height: 48,
+		minHeight: 48,
```
pwltr: Is there an issue with the height here? Aldert asked me to fix it in the past, it needs to be exactly 48px

pwltr (changes requested):

pwltr:
Maybe show an empty state here or hide the tags icon when there are no tags?

Simulator Screen Shot - iPhone 13 - 2023-04-05 at 21 19 28

pwltr:
There is a bug when you select a range at the end of the month and it shows the first day of the next month as selected. Maybe only for when Monday is set as first day of the week.

Screenshot_1680722488
Screenshot_1680722491

pwltr:
Sometimes the calendar has 6 rows and the spacing is not 100% right.

Simulator Screen Shot - iPhone 13 - 2023-04-06 at 01 18 11

src/components/TabBar.tsx
limpbrains: fixed in #0f70dea

src/store/migrations/index.ts
limpbrains: fixed

src/screens/Activity/ActivityFiltered.tsx
limpbrains: fixed

src/screens/Activity/TagsPrompt.tsx
limpbrains: fixed

src/components/SearchInput.tsx
limpbrains: Before your fix searchInput was able to stretch vertically if there are many tags. I was trying to restore this behaviour. But I couldn't, so I'm reverting it. Still waiting for the discussion in figma

limpbrains:
> Maybe show an empty state here or hide the tags icon when there are no tags?

done

> There is a bug when you select a range at the end of the month and it shows the first day of the next month as selected. Maybe only for when Monday is set as first day of the week.

Can't reproduce, even tried to set different timezone and date of 5th March. Maybe you changed time or timezone without restarting the app?

limpbrains:
> Sometimes the calendar has 6 rows and the spacing is not 100% right.

After discussing it with Alders, I added another BottomSheet size just for this case 

image

src/screens/Activity/TimeRangePrompt.tsx
```diff
@@ -0,0 +1,403 @@
...
+		flex: 1,
+	},
+	divider: {
+		width: 16,
+	},
+	range: {
+		flex: 1,
+		justifyContent: 'center',
+		alignItems: 'center',
+	},
```
pwltr: I would change these styles to avoid the text jumping when the number of rows changes:

```suggestion
	range: {
		marginTop: 'auto',
		marginBottom: 36,
		alignItems: 'center',
	},
```

pwltr (changes requested):

pwltr:
> > Maybe show an empty state here or hide the tags icon when there are no tags?
> 
> done
> 
> > There is a bug when you select a range at the end of the month and it shows the first day of the next month as selected. Maybe only for when Monday is set as first day of the week.
> 
> Can't reproduce, even tried to set different timezone and date of 5th March. Maybe you changed time or timezone without restarting the app?

I can't reproduce it anymore either

src/screens/Activity/TimeRangePrompt.tsx
limpbrains: Still jumps for me

pwltr: I forgot, also need to  remove `marginTop: 'auto',` from `buttonContainer` styles

limpbrains: Nice, added

pwltr (approved):
Really nice work

PR #963, opened 2023-04-10
fix(profile): avoid 'undefined is not a function' by never closing hyperdrives

Nuhvi:
# Description

`undefined is not a function` is an old error while saving profiles, while I can't figure out what is it, I know it happens in `Hyperdrive` module, and caused when a drive is closed (as a part of cleaning up `useProfile` hook).

Not closing any drives in any hook, avoids all of these issues hidden in Hyper* modules.

The tradeoff is that number of opened cores will keep increasing in memory.
An update for the SDK will deduplicate the core sessions (will publish it as soon as it is reviewed and merged), but at least, all the drives opened while checking a profile or a feed, will stay in memory for the lifetime of the session. But that is ok I think.

---

## Type of change

Please delete options that are not relevant.

## QA Notes

- [x] Open the app, try to edit profile and save it a couple of times, you should not get `undefined is not a function` error ever.

## Bitkit Version
1.0.0-beta.24

### DIFF
--- a/src/screens/Settings/SlashtagsSettings/index.tsx
+++ b/src/screens/Settings/SlashtagsSettings/index.tsx
@@ -58,8 +58,6 @@ const SlashtagsSettings = (): ReactElement => {
 				if (!unmounted) {
 					setProfileError(error.message);
 				}
-			} finally {
-				drive.close();
 			}
 		})();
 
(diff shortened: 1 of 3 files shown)


limpbrains (approved):

Nuhvi:
Try #967 before merging this.

PR #967, opened 2023-04-13
fix: add polyfill for Promise.allSettled

limpbrains:
# Description

Currently if you try to save Slashtags profile app shows `undefined is not a function` error. This is because our React Native version doesn't support `Promise.allSettled` which is used inside hypercore library. So I'm adding it.

---

## Type of change

Bug fix

## Tests

No test

## QA Notes

Open app, edit the Profile name and try to save it, the should be no errors


Nuhvi (approved):
Great, let's try this.

dzdidi (approved):
Beautiful!

PR #978, opened 2023-04-17
feat: show failed backup warning

limpbrains:
# Description

BitKit now remembers time of each backup attempt. If last attempt was more than 30 minutes ago and still no success, BitKit will start to show backup error message every 10 minutes.

---

## Type of change

New feature (non-breaking change which adds functionality)

## Tests
No test

## Screen shot / Video
Insert Relevant screenshot / Video share

image

## QA Notes

- Turn device into airplate mode
- Change settings or add a few tags
- wait 30-40 minutes and then you should see backup failed warning


coreyphillips (changes requested):
The new backup types should get added to `shapes/backup.ts` along with a new entry to `migrations/index.ts` that resembles the following:
```
10: (state): PersistedState => {
		return {
			...state,
			backup: {
                             ...defaultBackupShape,
                             ...state.backup,
                         },
		};
	},
```
Otherwise, we risk a crash when trying to access them from the backup selector.

PR #984, opened 2023-04-20
fix: replace Orange color with Brand color, remove unused icons

limpbrains:
# Description

- orange color is now the same as brand - #FF6600
- #F7931A is now called bitcoin. It is only used for bitcoin logo
- fix bitcoin logo svg and remove it's duplicates
- remove all unused icons
- remove all hardcoded colors in icons, except white

## Type of change

Bug fix

## Tests

No test

## QA Notes

No tests required it is just UI polishing update


src/styles/colors.ts
```diff
@@ -65,10 +66,11 @@ const colors: IColors = {
 	green: '#75BF72',
 	green2: '#22C846',
 	yellow: '#FFD200',
-	orange: '#F7931A',
+	orange: '#FF6600',
```
pwltr, line 69: Can we not remove this color entirely and use `brand` instead?

limpbrains: There are still 2 colors in Figma - Brand and Orange, that is why I didn't remove it completely.
I don't think Orange will come back, but I would rather wait a few releases before removing

pwltr: Okay yeah mapping to Figma makes sense

pwltr (approved):

PR #982, opened 2023-04-19
chore(wallet): Upgrade react-native to 0.71.7

coreyphillips:
# Description

- Upgrades `react-native` to `0.71.7`.
- Upgrades dependencies in `package.json`.
- Android is now using `hermes` instead of the `v8` engine.

## Type of change
- [x] Dependency Upgrade (non-breaking changes which adds functionality)

## QA Notes

- Prior to running Bitkit we need to first specify either `yarn setup-ios` or `yarn setup-android`.
    - This will be removed once we have a stable `libsodium` branch for both iOS & Android.
- Please suggest changes or create PR's against this branch if you see anything that needs fixing or additional attention.
- All functionality throughout the app should be tested to ensure that functionality is retained or improved.

PR #996, opened 2023-04-24
fix: remove flipper during e2e tests

limpbrains:
# Description

remove Flipper during e2e tests 

---

## Type of change

Bug fix

## Tests

Detox test

## QA Notes

Nothing to QA here

### DIFF
--- a/.github/workflows/e2e-ios.yml
+++ b/.github/workflows/e2e-ios.yml
@@ -1,12 +1,9 @@
 name: e2e-ios
 
-on:
-  pull_request:
-    branches:
-      - 'master'
-  push:
-    branches:
-      - 'master'
+on: pull_request
+
+env:
+  NO_FLIPPER: 1
 
 jobs:
   e2e:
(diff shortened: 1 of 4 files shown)


pwltr:
To fix the modal bug we need to revert [this commit](https://github.com/synonymdev/bitkit/pull/982/commits/250690c31bdce9277e4839a99c626e49b3671855). I thought it was fixed in the latest RN version but looks like it's still there 👀 

We can probably move to native Alert in the next step which should work.

pwltr:
Have you seen this? Maybe a cleaner solution
https://dev.to/retyui/disable-flipper-on-ci-builds-for-react-native-complete-guide-bnd

limpbrains:
> Have you seen this? Maybe a cleaner solution https://dev.to/retyui/disable-flipper-on-ci-builds-for-react-native-complete-guide-bnd

Thx, will try!

limpbrains:
Indeed, NO_FLIPPER is much cleaner. I also removed unused cache config, now the job is 10 min faster.
I will try to add them back later, after benchmarking

pwltr (approved):

Issue #937, opened 2023-03-17
[Bug]: Wiping wallet freezes app


SeverinAlexB:
### Describe the bug

After clicking `Reset and restore wallet` and confirming `Yes, Reset`, the app freezes. I need to kill the app to continue. The wallet is wiped.

### Reproduce

1. Go to `Settings`, `Back Up Or Restore`, `Reset and restore wallet`.
2. Click on `Yes, Reset`.
3. App freezes.

### Screenshots / Recording

![IMG_1241](https://user-images.githubusercontent.com/8782386/225851812-ebcc76ff-b908-44b7-8773-7fbdbcbc77f4.png)

### Operating system

iOS 16.1.1

### Bitkit version

Build 37

JeanlChristophe:
Reopen the ticket because we can find a bug sometimes.

pwltr:
Issue is likely caused by `RNFS.unlink()`, see https://github.com/itinance/react-native-fs/issues/1174

JeanlChristophe:
Thanks

PR #1002, opened 2023-04-27
UI: use native Alerts

pwltr:
_tbd_

### DIFF
--- a/src/components/Suggestions.tsx
+++ b/src/components/Suggestions.tsx
@@ -19,7 +19,7 @@ import { removeTodo } from '../store/actions/todos';
 import { showBottomSheet } from '../store/actions/ui';
 import { useAppSelector } from '../hooks/redux';
 import { useBalance } from '../hooks/wallet';
-import Dialog from './Dialog';
+import { showAlert } from '../utils/alert';
 import type { RootNavigationProp } from '../navigation/types';
 import { todosSelector } from '../store/reselect/todos';
 import {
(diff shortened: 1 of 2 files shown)


pwltr:
@limpbrains Could you please check if the option for Alert `userInterfaceStyle: 'dark'` works on your iOS? [It should work](https://snack.expo.dev/n7uXEtGO5) but for some reason it is not rendering in dark mode for me on the simulator. Maybe it is a simulator issue?

Would like to replace our custom `<Dialog />` with the native alerts to avoid the bugs and sometimes freezes we're seeing.

limpbrains:
> @limpbrains Could you please check if the option for Alert `userInterfaceStyle: 'dark'` works on your iOS? [It should work](https://snack.expo.dev/n7uXEtGO5) but for some reason it is not rendering in dark mode for me on the simulator. Maybe it is a simulator issue?
> 
> Would like to replace our custom `<Dialog />` with the native alerts to avoid the bugs and sometimes freezes we're seeing.

Doesn't work for me either :(

Can we just use 
```
{show && <View><Modal /><View>}
```

hack or it doesn't solve all the problems?
Animation still works in this case

pwltr:
Ok thanks for checking. Yes your solution could work. There is also [this issue](https://github.com/synonymdev/bitkit/issues/937) and I remember trying something similar with `{show && <Modal />}` but maybe the additional `<View />` makes the difference? Let me try.

limpbrains:
I don't think #937 is related to Alerts. 
Same thing happens to me when I do reset throught Dev menu.

pwltr:
> I don't think #937 is related to Alerts. Same thing happens to me when I do reset throught Dev menu.

Okay I wasn't 100% sure. Then that's a separate issue and for the Dialogs we can do the above solution 👍

PR #1006, opened 2023-04-28
fix: remove lnurlpay from Payment Preference

limpbrains:
# Description

Remove lnurlpay from Payment Preference, because it is not used and only brings confusion.

## Type of change

Bug fix

## Tests
No test

## QA Notes

Make sure that LNURL-Pay is disapeared from Payment Preference menu

### DIFF
--- a/src/store/migrations/index.ts
+++ b/src/store/migrations/index.ts
@@ -85,6 +85,20 @@ const migrations = {
 			checks: defaultChecksShape,
 		};
 	},
+	10: (state): PersistedState => {
+		// remove lnurlpay from receivePreference
+		const receivePreference = state.settings.receivePreference.filter(
+			(i) => i.key !== 'lnurlpay',
+		);
+
+		return {
+			...state,
+			settings: {
+				...state.settings,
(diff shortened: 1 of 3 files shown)


src/store/migrations/index.ts
```diff
@@ -85,6 +85,20 @@ const migrations = {
 			checks: defaultChecksShape,
 		};
 	},
+	10: (state): PersistedState => {
+		// remove lnurlpay from receivePreference
+		const receivePreference = state.settings.receivePreference.filter(
+			(i) => i.key !== 'lnurlpay',
+		);
```
pwltr, line 92: This will be an array `TReceiveOption[]` instead of `TReceiveOption`. I would say just reset this setting to the initial value of `defaultReceivePreference`.

pwltr (changes requested):

src/store/migrations/index.ts
pwltr: Oh wait my bad, array value is correct here

pwltr (approved):

PR #1008, opened 2023-05-01
feat: show backup status

limpbrains:
# Description

There are 2 types of backup in Bitkit. And we need to validate them differently

#### Backpack ####

Data just pushed to the seeder using backpack client and SlashtagsRPC. Because SlashtagsRPC promise doens't throw timeout error, we just memorise time of the last attempt. If it was too long ago (30 minutes), then we show an error and failed state in Setting/Backup screen.

#### Profile and Contacts Hypercores #####

They are synced in the background so we still can't get backup status easily. So to do that we periodically ask Seeder-server if it has these hypercores. If it does - backup is successfull. 

One more difference here, that Backpack backups can be easily retried, by just altering redux state (`backup.remoteLdkActivityBackupSynced` for example). But we can't do that for Hypercores, so refresh button is disabled for them.

---

## Type of change
New feature

## Tests
Unit test

## Screen shot / Video

image

## QA Notes

- Turn device into airplate mode
- Change settings or add a few tags
- wait 30-40 minutes and then you should see backup failed warning
- go to setting/back up screen and check backup status


limpbrains:
migration is not needed, because all new keys in store are undefined by default

coreyphillips:
> migration is not needed, because all new keys in store are undefined by default

The object keys that we're attempting to access, however, won't exist and may crash on certain devices. The updated `defaultBackupShape` and migration implemented [here](https://github.com/synonymdev/bitkit/pull/1009) should prevent that by creating the keys and settings these values to `undefined` by default.

limpbrains:
Alright, thx for the PR

coreyphillips (approved):

PR #1039, opened 2023-05-12
chore(wallet): Upgrade Reanimated Library

coreyphillips:
### Description
- Upgrades `react-native-reanimated` to `3.1.0`.
- Updated `SwipeToConfirm` component.
- Removes all instances of `LayoutManager`.

### Type of change

### Tests
- [x] No test

### QA Notes
- Ensure the reanimated components continue to work as expected throughout the app.
- To test the updated `SwipeToConfirm` component, navigate to the "ReviewAndSend" view by pasting a Bitcoin/Lightning address after tapping "Send", select "Max" amount, slide to confirm and select "No". The slider should reset prior to navigating back to the previous view.


limpbrains (changes requested):
App now throws a lot of annoying `Overriding previous layout animation...` warning. Can this be fixed?

Also I don't understand why new `ref.reset()` method is needed, if there is already an animation when you change `loading` prop from true to false ? I've tried to comment it out, still works as expected. Am I missing something?


https://github.com/synonymdev/bitkit/assets/155891/d4a13e6a-ae21-4146-bb40-2ec05d75c3ea

coreyphillips:
Unfortunately, I haven't found a way to rid that warning. Looking into it.

On my Android device at least, the loading indicator and circle position would never reset and remain locked in the final state. So providing parent components a way to reset the state after certain actions (Selecting no when the alert appears in this case) fixes it.

limpbrains:
> On my Android device at least, the loading indicator and circle position would never reset and remain locked in the final state. So providing parent components a way to reset the state after certain actions (Selecting no when the alert appears in this case) fixes it.

To archive this try to replace `await swipeToConfirmRef.current?.reset();` with `setIsLoading(false);`

[untitled.webm](https://github.com/synonymdev/bitkit/assets/155891/c74b38a5-a024-4e21-918e-90f4a2a1dd82)

coreyphillips:
Looks like the warning is due to the use of `LayoutAnimation` in a component. Ex `LayoutAnimation.easeInEaseOut();` We could just remove them completely. Looking to see if we can get around that though.

coreyphillips:
> > On my Android device at least, the loading indicator and circle position would never reset and remain locked in the final state. So providing parent components a way to reset the state after certain actions (Selecting no when the alert appears in this case) fixes it.
> 
> To archive this try to replace `await swipeToConfirmRef.current?.reset();` with `setIsLoading(false);`
> 
>  untitled.webm

Updated [here](https://github.com/synonymdev/bitkit/pull/1039/commits/79ff28c22b4afaf718789c2c559c5021424642cf).

coreyphillips:
I opted to remove the old `LayoutAnimation` implementations in order to remove the warnings that we're seeing [here](https://github.com/synonymdev/bitkit/pull/1039/commits/6014f9d7a56f2db522eb009872fb75ee21fba743).

limpbrains (approved):

PR #1040, opened 2023-05-12
refactor: add eslint rules for promises, fix ensuing errors

pwltr:
### Description

- add eslint rules for promises, fix ensuing errors
- split up some helpers into smaller files
- fix types for function props
- fix e2e tests for onchain, settings & slashtags
- rename __DISABLE_ANIMATION__ & __DISABLE_PERIODIC_REMINDERS__ to __E2E__

### Type of change

Refactoring

### Tests

Fixed some e2e tests


src/utils/nodejs-mobile/index.ts
```diff
@@ -29,17 +28,17 @@ const listeners = {};
  * @param {string} [id]
  * @param {ENodeJsMethod} method
  * @param {(data: any) => void} resolve
- * @returns {Promise<void>}
+ * @returns {void}
  */
-async function setupListener({
-	id = uuidv4(),
```
coreyphillips, line 35: It shouldn't be necessary to always require an id when calling `setupListener`. It's only if we need to setup special listeners for a given request. I would revert this to keep the id as optional and auto-generate the id for the listener method if empty as before (`id = uuidv4()`) since we always need this id to track listeners.

coreyphillips (changes requested):

src/utils/nodejs-mobile/index.ts
pwltr: Reverted [here](https://github.com/synonymdev/bitkit/pull/1040/commits/f5e75d7c143363a6ad072415b28bfdc82b05abe9)

coreyphillips (approved):

PR #1041, opened 2023-05-13
chore: update e2e workflow

pwltr:
Speed up Docker setup step in CI

### DIFF
--- a/.github/workflows/e2e-ios.yml
+++ b/.github/workflows/e2e-ios.yml
@@ -15,24 +15,16 @@ jobs:
         attempt: [1, 2]
 
     steps:
-      - name: Echo attempt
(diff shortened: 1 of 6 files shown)


limpbrains:
This is awesome, also take a look at 
https://github.com/douglascamata/setup-docker-macos-action
I tried this before, but now it should work with `docker compose`

pwltr:
@limpbrains Should we remove matrix now? We can add more test attempts if it still fails too often and it should be faster than two parallel jobs.

limpbrains:
Sure, let’s give it a try. Additionally, caching is not functioning correctly in matrix mode, so we will be able to optimise for it.
To ensure it is stable, let’s run tests 3-5 times

limpbrains (approved):

PR #1045, opened 2023-05-15
refactor(wallet): remove unused wallet store fields

pwltr:
### Description

- remove all unused fields from wallet reducer
- move `addressTypes` out of reducer

### Type of change

Refactoring (improving code without creating new functionality)

### Tests

No test


coreyphillips:
This LGTM. My only suggestion is that we leave `blacklistedUtxos` in. Otherwise, we'll have to migrate it back in at a later date once we add that feature.

coreyphillips (approved):

PR #1046, opened 2023-05-15
fix(send): fixes & refactoring in send flow

pwltr:
### Description

- refactor Recipient screen
- reset transaction state when opening Send flow
- fix max send when using Lightning
- fix ReviewAndSend LN invoice rendering
- fix ReceiveQR tooltip position
- fix settings e2e test

### Type of change

Bug fix (non-breaking change which fixes an issue)

### Tests

No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/8538369/718927d0-4bfe-411b-933c-96587ec4f1cc

### QA Notes

Test all ways to input valid data:
- from main scanner
- from send scanner
- from paste (button)
- from paste (system paste)
- from contacts
- onchain & lightning
- iOS & Android

valid data: unified, Bitcoin address, LN invoice, slashtag


limpbrains (approved):

limpbrains:
If I remember correctly, we also reset onchain transaction on Send close, maybe we should only keep one solution ?

pwltr:
> If I remember correctly, we also reset onchain transaction on Send close, maybe we should only keep one solution ?

We can remove the `onClose`. The issue with it is if you close the app it doesn't run and you end up with broken transaction state.

limpbrains:
> > If I remember correctly, we also reset onchain transaction on Send close, maybe we should only keep one solution ?
> 
> We can remove the `onClose`. The issue with it is if you close the app it doesn't run and you end up with broken transaction state.

Agree

pwltr:
@limpbrains I was able to simplify the `Recipient` screen by a lot while keeping all functionality. Please review again. I need find out how to work with clipboard in detox and then I will add some tests.

src/navigation/bottom-sheet/SendNavigation.tsx
```diff
@@ -92,23 +98,30 @@ export const sendNavigation = {
...
+	const selectedWallet = useSelector(selectedWalletSelector);
+	const selectedNetwork = useSelector(selectedNetworkSelector);
 	const { isOpen, screen } = useSelector((state) => {
 		return viewControllerSelector(state, 'sendNavigation');
 	});
 
 	const initialRouteName = screen ?? 'Recipient';
 
 	const onOpen = async (): Promise<void> => {
+		await updateOnchainFeeEstimates({ selectedNetwork, forceUpdate: true });
```
limpbrains, line 111: Looks like we are always using updateOnchainFeeEstimates with force: true

limpbrains (approved):

Issue #502, opened 2022-10-13
Pre-Launch Tasks


pwltr:
Things to do before publishing Bitkit to the App Store.

### Tasks
- update AppleID in `About` and `EasterEgg` screens
- update About screen links
- remove `usesCleartextTraffic` from `AndroidManifest.xml`

Nuhvi:
should we close this now @pwltr?

pwltr:
No, these are tasks for when we do a full launch, meaning moving from Testflight to App Store. We also use some non TLS endpoints currently that we should swap at some point.

PR #1053, opened 2023-05-19
chore(wallet): Update Android Manifest

coreyphillips:
### Description
-Removes usesCleartextTraffic from AndroidManifest.xml.

### Linked Issues/Tasks
- #502

### Tests
- [x] No test

### QA Notes
- Ensure Android still functions and is able to make api requests as expected.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -12,9 +12,6 @@
     <uses-permission tools:node="remove" android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission tools:node="remove" android:name="android.permission.SYSTEM_ALERT_WINDOW" />
 
-    <!-- TODO: Remove this line before launch -->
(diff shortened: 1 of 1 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -12,9 +12,6 @@
     <uses-permission tools:node="remove" android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission tools:node="remove" android:name="android.permission.SYSTEM_ALERT_WINDOW" />
 
-    <!-- TODO: Remove this line before launch -->
-    <!-- android:usesCleartextTraffic="true" -->
```
pwltr, line 16: This is just a comment, the actual rule is further down (XML doesn't support inline comments). Not sure we want to remove this rule though, at the very least in development we want to connect to electrum servers over http.

pwltr (changes requested):

PR #1058, opened 2023-05-23
fix(lightning): show error if unable to afford channel

### LINKED ISSUE #1050: [Bug]: Can't open channel: fee error
### Describe the bug

I tried to open a channel and got this.

```
Unable To Broadcast Transaction

the transaction was rejected by network rules.

min relay fee not met, 79 < 113
```
How do I fix that, and how could the error be easier for a user to understand?

### Reproduce

Any attempt to open a channel does this.

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/8538369/c8a9ae17-6ca6-46e0-977a-f4faa0dc3031

### Operating system

Android

### Bitkit version

1.0 (46)

### Log output

_No response_

pwltr:
### Description

Improve channel purchase UX by showing an error toast when the user is unable to afford the cost for the channel.

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/1050

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/8538369/892c90b3-97b1-4462-9cfc-2eb869010634

### QA Notes

If unable to afford the cost for a channel an error toast should appear.

### DIFF
--- a/src/utils/wallet/transactions.ts
+++ b/src/utils/wallet/transactions.ts
@@ -1084,7 +1084,7 @@ export const updateFee = ({
 	selectedNetwork?: TAvailableNetworks;
 	index?: number;
 	transaction?: ISendTransaction;
-}): Result<string> => {
+}): Result<{ fee: number }> => {
 	if (!selectedWallet) {
 		selectedWallet = getSelectedWallet();
 	}
@@ -1107,19 +1107,13 @@ export const updateFee = ({
 		inputs: transaction.inputs,
 	});
 
-	const max = transaction.max;
-	const message = transaction.message;
-	const outputs = transaction.outputs;
+	const { max, message, outputs } = transaction;
 	let address = '';
 	if (outputs.length > index) {
 		address = outputs[index]?.address ?? '';
 	}
 
-	//Check that the user has enough funds
-	const newFee = getTotalFee({
-		satsPerByte,
-		message,
-	});
+	const newFee = getTotalFee({ satsPerByte, message });
 
 	//Return if the new fee exceeds half of the user's balance
 	if (newFee >= inputTotal / 2) {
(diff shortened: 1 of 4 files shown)


src/utils/wallet/transactions.ts
```diff
@@ -1107,27 +1107,13 @@ export const updateFee = ({
...
-		satsPerByte,
-		message,
-	});
-
-	//Return if the new fee exceeds half of the user's balance
-	if (newFee >= inputTotal / 2) {
-		return err(
-			'Unable to increase the fee any further. Otherwise, it will exceed half the current balance.',
-		);
-	}
```
pwltr: Not sure why this limit is needed, maybe we can remove it now that we have warning dialogs for various cases in the send flow?

coreyphillips: We should leave this limit in in order to prevent the user from creating a transaction with a fee greater than the amount they are trying to spend. Electrum also has rules to prevent this from occurring, but it's better they encounter that limit here then after they try to broadcast and have it fail.

coreyphillips (changes requested):

src/utils/wallet/transactions.ts
pwltr: Is the rule for half or the full amount? We need to handle this case in multiple more places and show errors if the limit is `fee >= inputTotal / 2`. With a onchain balance of ~50000 sats I run into issues here quite often.

coreyphillips: Sorry. Yes, meant to say if it's greater than 50% of the full amount (Ex: Any fee >= 500 for a 1000 sat tx).

coreyphillips (approved):

PR #1064, opened 2023-05-24
fix(send): only process payment data

pwltr:
### Description

This reverts pasting via OS paste to not automatically navigate the user and also prevents processing non payment data (slashAuth, slashFeed, slashtag without payment data) in the send flow.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/8538369/d4ceea7c-352a-41bf-9ee8-d5f678a775c3

### QA Notes

In the Send flow only payable data should be processed. Pasting via OS paste should not forward the user but data should still be validated. Using the paste button or using the scanner should show errors for non-payable data. The main scanner should still process all supported data.

Test URIs:

- invalid BIP21: bitcoin:bcrt1qfwysev5dyu56xga9egkdemgd28hy9cv8fdt4v
- valid BIP21: bitcoin:bcrt1qfwysev5dyu56xga9egkdemgd28hy9cv8fdt4vm?amount=0.00012
- valid slashtag w/ payment data: slash:3kdqkopdjdkx5umunigzyw8ckoucfqe1k96obxu4tds1urm5a6uo
- valid slashtag without payment data: slash:3kdqkopdjdkx5umunigzyw8ckoucfqe1k96obxu4tds1urm5a6u
- invalid slashtag: slash:1234
- valid slashAuthUrl: slashauth:unrreuy3r4qkadioomgrcfonqkf5d1smyegqpod3pp59aqxpruhy?q=tqs0dfzdvk
- valid slashFeedUrl: slashfeed:53egeirg88ghrg9ssjqf6136dqra3ixdbsikx3rhaarrpjbox8oy#encryptionKey=5da9xnmn6wboyiu3zihzr9k3tc11t5p4d9ypaimuwk7sbo3k7u5o


limpbrains:
I find it confusing, that I need to tap on the paste button after I've already pasted the data.

pwltr:
> I find it confusing, that I need to tap on the paste button after I've already pasted the data.

No, you would tap the 'Continue' button. For the OS paste we don't want to redirect the user. Here is the explanation from Aldert:

> The initial reasoning behind this decision to have a slightly different behavior for these two paste functions is that this allows advanced or security savvy users to use the input field more precisely, and to double check the full address they paste in there. In contrast, the paste action button functions as a shortcut to the review&send for daily/quick use.

limpbrains (approved):

PR #1074, opened 2023-05-26
chore: update dev dependencies

pwltr:
### Description

Update dev dependencies, remove unused package `semver`


package.json
```diff
@@ -138,68 +138,68 @@
...
     "browserify-zlib": "^0.2.0",
     "dns.js": "^1.0.1",
     "electrum-client": "github:BlueWallet/rn-electrum-client#76c0ea35e1a50c47f3a7f818d529ebd100161496",
-    "esbuild": "^0.17.10",
-    "eslint": "^8.19.0",
+    "esbuild": "^0.17.19",
+    "eslint": "^8.41.0",
+    "eslint-plugin-ft-flow": "^2.0.3",
     "eslint-plugin-import": "^2.27.5",
+    "eslint-plugin-jest": "^27.2.1",
```
limpbrains: What are these for ? some peer dependencies?

limpbrains (approved):

package.json
pwltr: Maybe not directly peer dependencies but I think these plugins are included in `@react-native-community/eslint-config` now, Couldn't find a way to exclude or disable them so I just added them. Not the best but it's just dev dependencies.

limpbrains: For some reason these dependencies of `@react-native-community/eslint-config` are installed in 
node_modules/@react-native-community/eslint-config/node_modules/
while others goes under  
`node_modules/`
that is why they are not found 

I was able to fix this with `yarn upgrade`, but that touched too many packages, probably not a good idea

limpbrains: I was able to fix this issue by running 
```
yarn upgrade <list of all dev dependencies>
```

after that I was able to remove 
```
    "@typescript-eslint/eslint-plugin": "^5.59.7",
    "@typescript-eslint/parser": "^5.59.7",

    "eslint-plugin-ft-flow": "^2.0.3",
    "eslint-plugin-import": "^2.27.5",
    "eslint-plugin-jest": "^27.2.1",
    "eslint-plugin-jsx-a11y": "^6.7.1",
    "eslint-plugin-react": "^7.32.2",
    "eslint-plugin-react-hooks": "^4.6.0",
    "eslint-plugin-react-native": "^4.0.0",
```
because all of the are part of @react-native-community/eslint-config

from packages.json and everything works fine, I can push this commit if you want

limpbrains: eslint-plugin-jsx-a11y and eslint-plugin-import are just unused packages

pwltr: Yes I arrived at the same result by now by deleting `yarn.lock` first for a clean install. Also removed peer deps & unused packages.

limpbrains: Deleting yarn.lock is not cool - this way you upgrade a lot of production dependencies.

__tests__/wallet.ts
```diff
@@ -17,10 +17,10 @@ describe('Wallet Methods', () => {
 		}
 		expect(res.value.onchain).toEqual(mnemonic);
 		expect(res.value.lightning).toEqual(
-			'old found shock tenant merit tower foster chase sauce stool book enhance public key whip group retreat member cabin blanket sorry pole gym wink',
+			'lawn host tribe green enrich crime anxiety volume item pill soon steak hip mother orbit balcony avocado half scrap topic race near cool change',
 		);
 		expect(res.value.tokens).toEqual(
-			'inspire sick rule wild near rebuild pride tomato shell come drip reduce street steel warrior project radar sister day title spice execute evolve outside',
+			'receive original tilt luxury autumn burden weapon mix bonus joke glow mobile west detect orient midnight slice control bargain height light clock argue oval',
```
pwltr, line 23: @limpbrains Do you know why this changed now? Maybe some packages never upgraded because of `yarn.lock` like `bitcoinjs-lib` & `bip39`, but why  is the derivation different now? Want to be careful here

pwltr:
Okay, this updated too many packages. I'm going to close here, can you open a PR with your commit instead @limpbrains ?

PR #1081, opened 2023-05-26
feat(wallet): Detect Reorgs And RBFs

coreyphillips:
### Description

- Bitkit now tracks transactions with less than 6-confirmations and notifies users of a relevant reorg or RBF.
- Added `unconfirmedTransactions` to `wallet` reducer.
- Created several new methods for tracking unconfirmed txs.
- Added a button to inject a fake transaction in the Dev menu.
- Added `exists` to `IActivityItem` types.
- Updated `mergeActivityItems` method.
- Added methods to clear store objects when rescanning.

With this update we're concerned with two main things that can happen to transactions with less than 6 confirmations.
 1. A transaction can get pushed back to the mempool due to a reorg.
 2. A zero-conf transaction can get removed from the mempool if it has insufficient fees for the given fee environment, a reorg occurs that knocks it from the mempool or it's RBF'd.

In the event number one occurs, Bitkit will detect that a previously confirmed transaction is now no longer confirmed and notifies the user accordingly. This is a best-case scenario where a user may only need to wait for the tx to confirm again. The only action that needs to be taken is ensuring the activity item is no longer marked as confirmed and the stored transaction height is adjusted.

In the event number two occurs, Bitkit will detect that a tracked transaction is no longer in the mempool and alert the user accordingly of a potential reorg or RBF. This situation is more severe since the transaction in question is now gone, reverting ownership of any received or sent funds. In this situation, Bitkit will set the `exists` state to false in the activity list item and highlight the tx entry in red in the activity list. Bitkit will also perform a rescan of the wallet to ensure funds and stored transactions are accurate.

### Type of change

### Tests

- [x] No test

### Screenshot / Video
ghost-tx
reorg
removed-from-mempool

### QA Notes

To test a reorg:
 1. Send funds to the wallet: `/regtest-deposit <address> 0.00005000`
 2. Mine a block: `/regtest-mine 1`
 3. Invalidate the block you just mined: `/regtest-invalidate-block <blockhash>`
 Expected: A blue notification should appear that matches the attached screenshot after detecting the reorg. The transaction should no longer appear as confirmed in the activity list.

To test an RBF or getting bumped from the mempool:
 1. Navigate to "Settings->Dev Settings->Inject Fake Transaction"
Expected: A red notification should appear that matches the attached screenshot after detecting the ghost/invalid tx. The tx in question should be red in the activity list. When tapping the activity list item the status should display "Removed from Mempool".

### Notes

- I added translations for the new notifications using Google translate. Please let me know if they need further adjustments.


src/utils/i18n/locales/ru/wallet.json
```diff
@@ -90,6 +90,8 @@
 	"activity_copied_tx_text": "ID транзакции скопирован в буфер обмена.",
 	"activity_confirming": "Подтверждение",
 	"activity_confirmed": "Подтверждена",
+	"activity_removed": "Удалено из мемпула",
+	"activity_removed_msg": "Пожалуйста, проверьте свой список действий. Любые затронутые транзакции будут выделены красным цветом.",
```
limpbrains, line 94: Thank you for the translation :) 
In future I think it is better to ingore languages, other than english. If new string is missing, english will be shown anyway. And we better use Transifex as a main store of translations, so we don't have to sync it back and forth.

limpbrains (changes requested):
LN transactions are now shown in red too

image

limpbrains:
Also I've pushed fix for e2e tests

src/store/migrations/index.ts
```diff
@@ -119,6 +119,21 @@ const migrations = {
...
+				wallets: {
+					...state.wallet.wallets,
+					[state.wallet.selectedWallet]: {
+						...state.wallet.wallets[state.wallet.selectedWallet],
+						unconfirmedTransactions: getNetworkContent({}),
+					},
+				},
+			},
+		};
+	},
```
pwltr: With our current state reconciler setting it's not possible to write migrations for state nested this deep, see note at the top of this file. Maybe we should switch to `hardSet` as state reconciler but this will have to be tested.

src/store/types/activity.ts
```diff
@@ -37,6 +38,7 @@ export type TLightningActivityItem = {
 	address: string;
 	message: string;
 	timestamp: number;
+	exists: boolean;
```
pwltr: This field seems unnecessary for a LN activity item.

pwltr (changes requested):

src/utils/i18n/locales/ru/wallet.json
coreyphillips: > LN transactions are now shown in red too
> 
> image

Thanks. Updated default params/checks [here](https://github.com/synonymdev/bitkit/pull/1081/commits/fccb8d74ebd98d3eeb1cf43cd0b58cd2f5f531cf).

src/store/types/activity.ts
coreyphillips: Removed exists from types [here](https://github.com/synonymdev/bitkit/pull/1081/commits/fccb8d74ebd98d3eeb1cf43cd0b58cd2f5f531cf).

src/store/migrations/index.ts
coreyphillips: Thanks for pointing that out. I'll set it to `hardSet` and run some tests on my end to ensure it's behaving as expected.

coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1081/commits/dbebbb99fee1d14c8e64db7382af5f8738854706).

coreyphillips: `hardSet` seems to prevent the app from starting. I changed `hardSet` back to `autoMergeLevel2` for now while I attempt to figure out the cause of the issue.

coreyphillips: Updated the migration for `unconfirmedTransactions` [here](https://github.com/synonymdev/bitkit/pull/1081/commits/60200af4d3afee5785e44ffd24d9e7808b857e0f). Should be good to go.

coreyphillips:
Tested the `unconfirmedTransactions` migration from old state to new without issue on my end. Please let me know if you encounter any issues during your testing.

limpbrains (approved):

src/screens/Activity/ActivityDetail.tsx
```diff
@@ -325,6 +326,15 @@ const OnchainActivityDetail = ({
 		);
 	}
 
+	if (activityType === EActivityType.onchain && !exists) {
```
pwltr, line 329: this will always be true since we're in `OnchainActivityDetail`

src/store/actions/wallet.ts
```diff
@@ -619,6 +696,374 @@ export interface ITransactionData {
...
+	}
+};
+
+/**
+ * Parses and adds unconfirmed transactions to the store.
+ * @param {TWalletName} [selectedWallet]
+ * @param {TAvailableNetworks} [selectedNetwork]
+ * @param {IFormattedTransactions} transactions
+ */
+export const addUnconfirmedTransactions = async ({
```
pwltr: unnecessary usage of `async`

src/store/actions/wallet.ts
```diff
@@ -619,6 +696,374 @@ export interface ITransactionData {
...
+		if (processRes.isErr()) {
+			return err(processRes.error.message);
+		}
+
+		const { unconfirmedTxs, outdatedTxs, ghostTxs } = processRes.value;
+		if (outdatedTxs.length) {
+			// Notify user that a reorg has occurred and that the transaction has been pushed back into the mempool.
+			const singular = i18n.t('wallet:reorg_msg_begin_singular');
+			const plural = i18n.t('wallet:reorg_msg_begin_plural');
+			await showInfoNotification({
```
pwltr: unnecessary `await`

pwltr (changes requested):

src/screens/Activity/ActivityDetail.tsx
coreyphillips: This is to appease typescript. Otherwise, exists would have to be added to each ActivityItem type.

pwltr: No this is not true. TS will know that `activityType` will always be `EActivityType.onchain` here and that `exists` will be there.

src/store/actions/wallet.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1081/commits/2e01f0f1948de799bdf3ca188cd9cdf85fee885c).

src/store/actions/wallet.ts
coreyphillips: Update [here](https://github.com/synonymdev/bitkit/pull/1081/commits/2e01f0f1948de799bdf3ca188cd9cdf85fee885c).

pwltr (approved):

pwltr (approved):

PR #1120, opened 2023-06-21
feat(lang): change locale format to ICU

limpbrains:
### Description

Problem: Transifex doesn't support i18next native file format.
Solution: Use ICU file format for language files
Which causes another problem: it is not expandable, we can't use cool i18next formatters like date or relative time
Solution: add another instance of i18next to format dates and relative time.

ICU file format uses single brackets instead of double, so I'me updating all language file.

### Type of change

Refactoring

### Tests

No test


src/utils/i18n/locales/en/wallet.json
```diff
@@ -1,54 +1,4 @@
...
-	"send_fee_and_speed": "Speed and fee",
-	"send_fee_speed": "Speed",
-	"send_fee_error": "Error Updating Fee",
-	"send_fee_custom": "Set Custom Fee",
-	"send_fee_total": "{{feeSats}} sats for this transaction",
-	"send_fee_total_fiat": "{{feeSats}} sats for this transaction ({{fiatSymbol}}{{fiatFormatted}})",
-	"note": "Note",
-	"tags": "Tags",
-	"tags_add": "Add Tag",
-	"tags_add_button": "Add",
```
pwltr: Deleted by accident?

pwltr (changes requested):

src/utils/i18n/locales/en/wallet.json
limpbrains: omg, yes

limpbrains: fixed

src/store/actions/wallet.ts
```diff
@@ -863,28 +863,22 @@ export const checkUnconfirmedTransactions = async ({
...
-			const titleIntro =
-				ghostTxs.length === 1
-					? i18n.t('wallet:transaction')
-					: i18n.t('wallet:transactions');
-			await showErrorNotification({
-				title: `${ghostTxs.length} ${titleIntro} ${i18n.t(
-					'wallet:activity_removed',
-				)}`,
+			showErrorNotification({
+				title: i18n.t('wallet:activity_removed_title_one', {
```
pwltr: I think this needs to be

```suggestion
				title: i18n.t('wallet:activity_removed_title', {
```

pwltr (changes requested):

src/store/actions/wallet.ts
limpbrains: indeed

pwltr (approved):

PR #1167, opened 2023-07-19
chore(wallet): Upgrade RN to 0.72.3

coreyphillips:
### Description
This PR ensures we're on top of RN upgrades and changes so it doesn't pile up too much down the road. This PR should also improve readability of error messages within the app when crashes occur and help with troubleshooting.
- Upgrades react-native to version 0.72.3.
- Upgrades other dependencies as needed.
- Applies patch to react-native-camera-kit.

### Type of change
- [x] Dependency Upgrade

### Tests
- [x] No test


pwltr:
Slashtags seems to be broken on this build:

Simulator Screenshot - iPhone 14 - 2023-07-20 at 21 13 26
Simulator Screenshot - iPhone 14 - 2023-07-20 at 21 11 26

coreyphillips:
Thanks! Able to reproduce. Looking into it.

pwltr:
The error looks like it has to do with async generator code which I believe hyperswarm doesn't support. Think we had a polyfill for that once maybe that got removed or is now needed again? Maybe @limpbrains knows more.

coreyphillips:
@pwltr, I made the mistake of regenerating the `yarn.lock` file. As such, it bumped some of the slashtag-sdk dependency versions to something incompatible. Updated it [here](https://github.com/synonymdev/bitkit/pull/1167/commits/a2660d0b0181f027addab3a2168fecac974b7e79). Should be working as expected now.

limpbrains:
> @pwltr, I made the mistake of regenerating the `yarn.lock` file. As such, it bumped some of the slashtag-sdk dependency versions to something incompatible. Updated it [here](https://github.com/synonymdev/bitkit/pull/1167/commits/a2660d0b0181f027addab3a2168fecac974b7e79). Should be working as expected now.

this is why I'm for pinning all deps in package.json. Of course it doesn't solve all the issues

pwltr:
Any idea why e2e tests are failing? It looks like some async tasks are taking way longer in this build in the detox environment. I can make the tests pass locally by increasing the timeouts but on CI it's still not enough it seems.

coreyphillips:
> Any idea why e2e tests are failing? It looks like some async tasks are taking way longer in this build in the detox environment. I can make the tests pass locally by increasing the timeouts but on CI it's still not enough it seems.

Same. I'm at a loss as to why this is failing.

pwltr:
> > Any idea why e2e tests are failing? It looks like some async tasks are taking way longer in this build in the detox environment. I can make the tests pass locally by increasing the timeouts but on CI it's still not enough it seems.
> 
> Same. I'm at a loss as to why this is failing.

Detox currently doesn't support this version of react-native so probably we will have to wait for that. I don't think it's worth breaking e2e tests unless there is something major in this upgrade that we need?

coreyphillips:
Sounds good. Lets wait. One of the added benefits of the upgrade is the potential for slightly improved error messages for troubleshooting.

limpbrains:
image

limpbrains:
Before merging I think we need to test it ourselfs, especially android 
Also, please, sqash it

coreyphillips:
Agreed. I'll run through some tests on Android in the morning.

coreyphillips:
Android build is working as expected. No hiccups thus far. Seems we've lost the e2e-ios test again? Not sure what happened there.

limpbrains:
I think we can ignore it for now, I will try to fix it in another PR

limpbrains (approved):

PR #1173, opened 2023-07-24
feat(recovery): switch to quick-action to enter recovery

pwltr:
### Description

Switches to a more reliable way of entering recovery by adding a quick-action for it.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/8538369/ac06ed74-b12f-4b93-9300-80c6b2f57a8a

### QA Notes

The quick-action will be added only after starting the app once, should be fine since there is no reason to enter recovery on a fresh install.


coreyphillips:
Received the following error when running this version of the app for the first time from a previously installed master branch version. The app worked as expected afterwards.

Any way we can wrap `QuickActions.setShortcutItems` in a function and/or try/catch to prevent this error when opening the app after this update?

Screenshot_1690292865

pwltr:
@coreyphillips I can reproduce this but only if omitting `yarn android`. There is a native component to this so you'll have to rebuild first.

coreyphillips (approved):

PR #1174, opened 2023-07-25
fix: activity total = value + fee

limpbrains:
### Description

- Money now uses decimalLength=long by default
- convert onchain fee to sats, so it matches value and LN activity 
- show real fee for LN payments
- LN activity items value should be positive. I'm adding a migration for that
- show activity total as value + fee

### Linked Issues/Tasks

https://app.asana.com/0/1189635884316057/1205072416958950/f

### Type of change

Bug fix

### Tests

No test

### Screenshot / Video

image

image

image

### QA Notes

Make sure that activity represents tx history


pwltr:
Onchain Receive screen looks different compared to design, I think fees should not be included for receive? 

Simulator Screenshot - iPhone 14 - 2023-07-31 at 23 06 45

src/screens/Activity/ListItem.tsx
```diff
@@ -99,6 +100,7 @@ const OnchainListItem = ({
 	const isTransferringToSpending = isTransfer && isSend;
 
 	let title = t(isSend ? 'activity_sent' : 'activity_received');
+	const amount = isSend ? value + fee : value;
```
pwltr, line 103: In `utils/activity/index` we are subtracting the fee for onchain send, maybe cleaner to remove the code there?

pwltr (changes requested):

limpbrains:
> Onchain Receive screen looks different compared to design, I think fees should not be included for receive?

fixed

src/screens/Activity/ListItem.tsx
limpbrains: Yes, but we can't do this for LN payments, since activity list there is the only sorce of truth for now.
So I would prefer to keep it this way for now

pwltr (approved):

PR #1216, opened 2023-08-22
chore(lightning): Upgrade To Blocktank v2 API

coreyphillips:
### Description

Upgrades Bitkit to the Blocktank v2 API [here](https://github.com/synonymdev/blocktank-lsp-http/tree/master/client).


coreyphillips:
@limpbrains, any thoughts on why this test might be failing?

limpbrains:
test expected QuickSetupReserveNote to be shown, but in reality QuickSetupBlocktankNote is shown. I've changed test to reflect it.
I've removed all sleep(1000), it is really not needed, when you run test with E2E=true.
Also test was failing with `[BlocktankClientError: Failed to create order. "Channel size too big. Max size is USD999."]` on the last step when buying big barrel on CustomSetup screen. Looks like a rounding bug in Bitkit ot Blocktank. I changed test to use medium barrel - $500. Maybe this needs to be fixed and reverted?

Here is a video of the test run:

https://github.com/synonymdev/bitkit/assets/155891/b06fd547-6c41-48f9-9863-54c92b1d3f08

coreyphillips:
Looks like the total channel size (incoming + receiving) was exceeding the `maxChannelSizeSat` value when the user opted for a spending value + the max receiving value in the custom flow. So in [this commit](https://github.com/synonymdev/bitkit/pull/1216/commits/21ab801752a851302976b5cf7e19e5dc70eb1f68) and [this commit](https://github.com/synonymdev/bitkit/pull/1216/commits/d4552e24e61e4c07e2683bf61ea645f0c7ea4b8b) we subtract any spending amount from the max receiving to ensure it remains under that limit in both Quick & Custom setups.

coreyphillips:
I no longer understand how to appease the testing gods as @limpbrains did, but it's at least ready for review. @pwltr

limpbrains:
Slider on the quick setup screen feels broken

https://github.com/synonymdev/bitkit/assets/155891/b8a2e7e2-0398-4e2e-811f-8a665ccf319e

limpbrains:
Testing Gods are Pleased

pwltr:
Bitkti UI is showing channel as 'paid' on home screen even though payment was not confirmed with BT yet. On the ChannelDetails screen shows the correct state.

```
{"id":"30798b6f-eabc-49f3-a1e3-a32a5cec7d07","state":"created","feeSat":193730,"lspBalanceSat":194564,"clientBalanceSat":192638,"zeroConf":false,"channelExpiryWeeks":12,"channelExpiresAt":"2023-11-20T10:54:42.561Z","orderExpiresAt":"2023-08-28T10:54:42.561Z","channel":null,"lspNode":{"alias":"Synonym-Regtest-1","pubkey":"03b9a456fb45d5ac98c02040d39aec77fa3eeb41fd22cf40b862b393bcfc43473a","connectionStrings":["03b9a456fb45d5ac98c02040d39aec77fa3eeb41fd22cf40b862b393bcfc43473a@35.233.47.252:9400"]},"lnurl":"LNURL1DP68GUP69UHKCMMRV9KXSMMNWSARJVPSXQHKZURF9AKZ7VESXUUNSC3KVCKK2CTZVVKNGWTXXVKKZVT9XVKKZVEJVY6KXETRXAJRQDCVA7KYJ","payment":{"state":"created","paidSat":0,"bolt11Invoice":{"request":"lnbcrt1937300n1pjwca02pp5xfc4ltue0jsn3w74fch6frfsyevewfn845defgkxargryw3t7g7sdzzgfkx7cmtw3sku6eqgd5xzmnwv4kzqge4gdq4gnn4gemkynrr23x85cnswf54y7fcx5cqzzsxqrrsssp59f4g0xedn5e7rfhxl0302r6mqls7kj9yya43hvnlztqwltclj78q9qyyssq7s4cmzzq79hngyf693vgfv6a767fp75wz6gv50s74np0egmr603za5kwusdzwp4d7ppgdjnwvh44xa24eutyy72w58l378gw5m9ev0qqxt6j5u","state":"pending","expiresAt":"2023-08-28T11:24:42.000Z","updatedAt":"2023-08-28T10:24:42.000Z"},"onchain":{"address":"bcrt1qld3dg32d2ual9whsz9y3uf4nj3qtadcz2hdjvf","confirmedSat":0,"transactions":[{"amountSat":193730,"txId":"9d33bc46dc65f627381eeceab0cea67984ee209b0e6d5e1f9d1a55116291b268","vout":1,"blockConfirmationCount":0,"feeRateSatPerVbyte":11.03914590747331,"confirmed":false}]}},"couponCode":"","discountPercent":0,"updatedAt":"2023-08-28T10:24:54.067Z","createdAt":"2023-08-28T10:24:42.561Z"}
```

Simulator Screenshot - iPhone 14 - 2023-08-28 at 12 26 49
Simulator Screenshot - iPhone 14 - 2023-08-28 at 12 26 47
Simulator Screenshot - iPhone 14 - 2023-08-28 at 12 26 54

src/store/actions/blocktank.ts
```diff
@@ -367,30 +336,31 @@ export const addPaidBlocktankOrder = ({
...
 
 	// queued for opening
-	if (order.state === 200) {
+	if (!order.channel?.state) {
 		setLightningSettingUpStep(2);
 	}
 
 	// opening connection
-	if (order.state === 300) {
+	if (order.channel?.state === BtOpenChannelState.OPENING) {
```
pwltr: `[TypeError: Cannot read property 'OPENING' of undefined]`

It would make sense to remove `declare` keyword from `BtOpenChannelState` so the enum is compiled to JS to be usable.

pwltr (changes requested):

src/store/actions/blocktank.ts
coreyphillips: Waiting on this [PR](https://github.com/synonymdev/blocktank-lsp-http/pull/5) in `blocktank-lsp-http`. Will then update accordingly.

coreyphillips: Set them as strings for now in this [commit](https://github.com/synonymdev/bitkit/pull/1216/commits/0c58170fac0d390f87b2b14bc99c7eaae4d22196) until the PR above is merged.

coreyphillips: Set them as string for now in this [commit](https://github.com/synonymdev/bitkit/pull/1216/commits/0c58170fac0d390f87b2b14bc99c7eaae4d22196) until the PR above is merged.

pwltr:
- "Connecting" and "Ready" cards are missing now, as well as the toast:
Screenshot 2023-08-30 at 15 05 30


- I have to manually pull to refresh for the channel to be opened

- After the channel was opened the ChannelDetails view still shows the wrong status "Opening connection".

- Wrong fee in transfer flow:
Screenshot 2023-08-30 at 15 05 30

- ~Orange bar overshoots the marker~ (not a new issue):
Screenshot 2023-08-30 at 15 05 30

- Getting the receive modal when closing channels

- After the channel was closed the ChannelDetails view still shows the wrong status.

coreyphillips:
As of the latest commit:
- The cards/notification should be restored when setting up and opening a channel.
- `ChannelDetails` should show the correct status.

I was not able to reproduce:
- I have to manually pull to refresh for the channel to be opened.
- Wrong fee in transfer flow.

"Getting the receive modal when closing channels" was always present as well. We can address that in a separate PR.

pwltr:
> Wrong fee in transfer flow.

The way to reproduce this is to open 2 channels: first one for 20% spending and then a transfer for 80% spending balance.

> I have to manually pull to refresh for the channel to be opened.

Getting this behaviour consistently. This time I had to restart the app for the channel to be opened. There is a call to `restartLdk()` in `actions/blocktank.ts`, maybe that's not running?

> Getting the receive modal when closing channels" was always present as well. We can address that in a separate PR.

There is logic [here](https://github.com/synonymdev/bitkit/blob/8aa9a2802e37980c8c8c86ad01ab50f5c146f3a9/src/store/actions/wallet.ts#L1206C4-L1206C4) that should prevent that. Probably has to be update for your PR?

- I'm getting an empty toast message on channel open.


- The cards for "connecting" and "ready" should only be shown for the initial connection (first channel), all subsequent channels should only show the "transfer to spending" card. See Figma "Transfer Funds to Savings (LN channel closing)"

coreyphillips:
In the latest set of commits I've fixed the empty toast message and the connecting/ready cards. I've also swapped out the `restartLdk` method with the `refreshLdk` method in blocktank's `handleOrderStateChange` so that should help resolve the issue with channel opens.

Unfortunately, I'm still not able to reproduce the wrong fee bug. 

Regarding the logic [here](https://github.com/synonymdev/bitkit/blob/8aa9a2802e37980c8c8c86ad01ab50f5c146f3a9/src/store/actions/wallet.ts#L1206C4-L1206C4). The reason this never worked is because the closing txid between Blocktank and what Bitkit ends up seeing are not the same when closing a channel. LDK broadcasts a separate tx to the receiving address provided by Bitkit. So that will need some additional work to get sorted.

src/screens/Transfer/Confirm.tsx
```diff
@@ -47,17 +47,27 @@ const Confirm = ({
...
 	const orders = useSelector(blocktankOrdersSelector);
 	const order = useMemo(() => {
-		return orders.find((o) => o._id === orderId);
+		return orders.find((o) => o.id === orderId);
 	}, [orderId, orders]);
 
-	const blocktankPurchaseFee = useDisplayValues(order?.price ?? 0);
+	const feeSat = order?.feeSat ?? 0;
+	const clientBalanceSat = order?.clientBalanceSat ?? 0;
+	const blocktankPurchaseFee = useDisplayValues(feeSat + clientBalanceSat);
```
pwltr: I think this is the cause for the wrong fee, you're adding the local balance to the fee

pwltr (changes requested):

pwltr:
> In the latest set of commits I've fixed the empty toast message and the connecting/ready cards. I've also swapped out the restartLdk method with the refreshLdk method in blocktank's handleOrderStateChange so that should help resolve the issue with channel opens.

Seems to be better now.

> Regarding the logic [here](https://github.com/synonymdev/bitkit/blob/8aa9a2802e37980c8c8c86ad01ab50f5c146f3a9/src/store/actions/wallet.ts#L1206C4-L1206C4). The reason this never worked is because the closing txid between Blocktank and what Bitkit ends up seeing are not the same when closing a channel. LDK broadcasts a separate tx to the receiving address provided by Bitkit. So that will need some additional work to get sorted.

Follow up issue [here](https://github.com/synonymdev/bitkit/issues/1221)

pwltr:
The custom setup flow is broken when selecting the maximum local balance:

https://github.com/synonymdev/bitkit/assets/8538369/24d08f14-430c-4ae1-bbb4-3541bf626f36

src/screens/Transfer/Confirm.tsx
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1216/commits/0eeed2f00cd549e0917c9d674552e8f146f0b142).

coreyphillips:
Updated the calculations in [this commit](https://github.com/synonymdev/bitkit/pull/1216/commits/0eeed2f00cd549e0917c9d674552e8f146f0b142). Custom channels should get calculated correctly now no matter how many sats the user has.

pwltr:
Fee is still incorrect here:

Untitled

pwltr:
I opened a channel with 80% spending balance and my receiving capacity is very small. I believe BT should always give me a balanced channel.

Simulator Screenshot - iPhone 14 - 2023-09-01 at 12 50 10

coreyphillips:
I'm surprised Blocktank allowed that channel purchase. The new api usually does not allow the client balance to exceed the lsp balance with the following error: `"clientBalanceSat is too big compared to lspBalanceSat. Max ratio: 1."`. 

In the latest commit, I've updated the `QuickSetup` component to set the max value on load to prevent the issue you posted above. I also adjusted the `Cost:` value to reflect the fee instead of the total sats needed to send to Blocktank.

Should be good to go for testing, just need to update the test script so that it passes with the new values for lightning onboarding.

pwltr:
> In the latest commit, I've updated the QuickSetup component to set the max value on load to prevent the issue you posted above.

I don't think we want to default to 80% from a design perspecive. Also it doesn't resolve the issue with the unbalanced channel for me.

> I also adjusted the Cost: value to reflect the fee instead of the total sats needed to send to Blocktank.

This still doesn't show the correct amount:

Simulator Screenshot - iPhone 14 - 2023-09-01 at 15 31 26
Simulator Screenshot - iPhone 14 - 2023-09-01 at 15 31 29

coreyphillips:
We've made it pretty far in this PR. I think we should call a checkpoint here so we can get some additional testing in with the new api and apply needed fixes in separate PR's from there

The Cost is the Blocktank fee. I don't believe we're able to calculate the cost of the transaction at this point. We might want to change the verbiage to "Blocktank Fee" or something similar.

The 80% is part of a larger discussion especially as anchor channels emerge.

pwltr:
> The Cost is the Blocktank fee. I don't believe we're able to calculate the cost of the transaction at this point. We might want to change the verbiage to "Blocktank Fee" or something similar.

This is working on the master branch. Unless I'm missing something I don't see the reason to introduce a regression here.

> The 80% is part of a larger discussion especially as anchor channels emerge.

I don't understand the need for this change. The unbalanced channel issue is independent of the 80% limit. I'm unable to get a balanced channel with any spending amount. This really needs to be fixed otherwise BT v2 channels will be pretty much useless for receiving.

coreyphillips:
The fee Cost discrepancy should be sorted as of the latest set of commits.

coreyphillips:
Also increased the lsp balance when setting 0 sats on the clients end to something more usable. Should be good to go.

pwltr:
Channel balancing seems better now, it's still defaulting to 80% spending balance in QuickSetup though. From my tests that doesn't affect balancing at all, and it's not in the design this way, so this should be reverted. Also there are a lot of leftover console logs.

coreyphillips:
@pwltr Could you expand on this? Do you mean it's defaulting to a value greater than $1000 at 80%?
"It's still defaulting to 80% spending balance in QuickSetup though. From my tests that doesn't affect balancing at all, and it's not in the design this way"

pwltr:
I'm talking about the slider on the QuickSetup screen, you've changed it to be set at 80% initially for a reason that is not apparent to me. It should be set to 20%.

image

coreyphillips:
Gotcha. Updated [here](https://github.com/synonymdev/bitkit/pull/1216/commits/8f05e630082b34952e8ffbbbc53c32b114f0494a). Changed it from setting it to the max of 80% to 20%.

src/store/actions/blocktank.ts
```diff
@@ -265,19 +258,33 @@ export const startChannelPurchase = async ({
...
+					address: buyChannelData.payment.onchain.address,
+					value: buyChannelData.feeSat,
 					index: 0,
 				},
 			],
 		},
 	});
 
-	return ok({ orderId: buyChannelData.order_id, channelOpenCost });
+	const feeRes = await updateFee({
```
pwltr: unnecessary `await`

src/store/actions/blocktank.ts
```diff
@@ -265,19 +258,33 @@ export const startChannelPurchase = async ({
 		);
 	}
 
-	updateSendTransaction({
+	await updateSendTransaction({
```
pwltr: unnecessary `await`

src/store/actions/blocktank.ts
```diff
@@ -98,11 +78,16 @@ export const refreshOrder = async (
...
-		if (order.state === 100) {
+		if (
+			order.payment.state === BtPaymentState.PAID &&
+			order.state !== BtOrderState.OPEN
+		) {
 			setLightningSettingUpStep(1);
-			const finalizeRes = await finalizeChannel(orderId);
+			const finalizeRes = await openChannel(orderId);
 			if (finalizeRes.isOk()) {
+				//setLightningSettingUpStep(2);
```
pwltr: remove comment

src/store/actions/blocktank.ts
```diff
@@ -367,30 +375,31 @@ export const addPaidBlocktankOrder = ({
...
-		return otherOrderStates.some((state) => states.includes(state));
-	};
+	//TODO: Not sure what to do with this
+	// const currentOrders = getBlocktankStore().orders;
+	// const otherOrders = currentOrders.filter((o) => o.id !== order.id);
+	// const otherOrderStates = otherOrders.map((o) => o.state);
+	//
+	// const oneOtherOrderHasState = (states: number[]): boolean => {.
+	// 	return otherOrderStates.some((state) => states.includes(state));
+	// };
```
pwltr: Can this be removed? Cards seem to work okay

pwltr (changes requested):
Couple of nitpicks other than that LGTM

src/store/actions/blocktank.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1216/commits/e852886655563df0f4cf8f21e885e456b5cafb7d).

src/store/actions/blocktank.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1216/commits/e852886655563df0f4cf8f21e885e456b5cafb7d).

src/store/actions/blocktank.ts
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1216/commits/e852886655563df0f4cf8f21e885e456b5cafb7d).

src/store/actions/blocktank.ts
coreyphillips: Yes, that was sorted. Forgot to remove the todo. Updated [here](https://github.com/synonymdev/bitkit/pull/1216/commits/e852886655563df0f4cf8f21e885e456b5cafb7d).

pwltr (approved):

PR #1237, opened 2023-09-08
feat: add lugano feed widget

Nuhvi:
### Description

Add Lugano feed widget.

### Linked Issues/Tasks

[Event widget feed timetable](https://app.asana.com/0/1200529269167147/1205322182636536)

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

Simulator Screen Shot - iPhone 14 - 2023-09-08 at 13 34 34


pwltr:
Resolve conflicts please.

limpbrains (changes requested):
Did you remove yarn.lock ? Please do not do this. You have upgraded a lot of peer dependencies. 
Can you please revert yarn.lock, ios/Podfile.lock, install packages you need and add yarn.lock again. 
Upgrading packages can break things, we better do it gradually and in separate PR

pwltr:
Can you fix the tests? It looks to me like your `react-native-fetch-hack` is trying to use `node-fetch` on the native mobile side, will that work? https://github.com/synonymdev/bitkit/actions/runs/6122214804/job/16617501219?pr=1237

Nuhvi:
> Can you fix the tests? It looks to me like your `react-native-fetch-hack` is trying to use `node-fetch` on the native mobile side, will that work? https://github.com/synonymdev/bitkit/actions/runs/6122214804/job/16617501219?pr=1237

You know better, but it is the only way I know of to load binary data, in this case images. If you know how to do it otherwise with the provided `fetch()` function, let me know, or just revert to the published npm version of the web-relay.

pwltr:
> You know better, but it is the only way I know of to load binary data, in this case images. If you know how to do it otherwise with the provided fetch() function, let me know, or just revert to the published npm version of the web-relay.

I don't know without looking into it, but trying to run `node-fetch` on the mobile client sounds like a massive workaround and judging by the tests it doesn't seem to work? Simple solution for this case would be to just render a static image for the banner in this widget since, the design looks like it has only one image. Would that work?

Nuhvi:
@pwltr This issue needs to be solved, the simple solution doesn't work for the rest of the Web Relay use-cases, when someone uploads a profile picture or whatever.

Isn't `node-fetch` already using the normal `globalThis.fetch` in browser environment as in Bitkit? so this is only a CI issue? I will try to make the CI hapyp, but it works in Bitkit fine.

If @rbndg thinks we should just hardcode these images fine.

limpbrains:
I was able to fix tests locally by chaning 
```patch
- import { Client } from '@synonymdev/web-relay';
+ import { Client } from '@synonymdev/web-relay/lib/client';
```
in src/components/SlashtagsProvider2.tsx


also ios/bitkit.xcodeproj/project.pbxproj needs to be resetted

Nuhvi:
@limpbrains Can you push that with the correct `ios/bitkit.xcodeproj/project.pbxproj`?

Nuhvi:
Pushed a version where `node-fetch` isn't required at all https://github.com/slashtags/web-relay/blob/react-native-fetch-hack/lib/client/fetch-react-native.js

Nuhvi:
Ok this latest error doesn't seem the web relay's fault @pwltr

pwltr:
Feeds are not working for me:

Simulator Screenshot - iPhone 14 - 2023-09-11 at 18 53 15

Nuhvi:
@pwltr @limpbrains all checks passed

limpbrains:
Works now!

limpbrains:
`LOG  Error while reading banner png [TypeError: Cannot read property 'byteLength' of null]`
On both android and iOS I can add a widget, but I can't see an image

Nuhvi:
@limpbrains Sigh... I will just read images as base64 and be done with it, but this is not sustainable.

src/components/Widgets.tsx
```diff
@@ -111,6 +112,10 @@ const Widgets = (): ReactElement => {
 					Component = FactsWidget;
 					testID = 'FactsWidget';
 					break;
+				case SUPPORTED_FEED_TYPES.LUGAON_FEED:
```
MiguelMedeiros:
```suggestion
				case SUPPORTED_FEED_TYPES.LUGANO_FEED:
```
typo mistake :)

src/hooks/widgets.ts
```diff
@@ -40,36 +45,42 @@ export const useSlashfeed = (options: {
...
-				const _fields = options.fields ?? _config.fields ?? [];
+				const _fields = options.fields ?? reader.config.fields ?? [];
 
 				// Don't continue for news & facts feeds
 				if (
-					_config.type === SUPPORTED_FEED_TYPES.FACTS_FEED ||
-					_config.type === SUPPORTED_FEED_TYPES.HEADLINES_FEED
+					reader.config.type === SUPPORTED_FEED_TYPES.FACTS_FEED ||
+					reader.config.type === SUPPORTED_FEED_TYPES.HEADLINES_FEED ||
+					reader.config.type === SUPPORTED_FEED_TYPES.LUGAON_FEED
```
MiguelMedeiros:
```suggestion
					reader.config.type === SUPPORTED_FEED_TYPES.LUGANO_FEED
```
typo mistake :)

src/screens/Widgets/Widget.tsx
```diff
@@ -169,18 +164,26 @@ const Widget = ({
 											);
 										case SUPPORTED_FEED_TYPES.FACTS_FEED:
 											return <FactsWidget key={url} url={url} />;
+										case SUPPORTED_FEED_TYPES.LUGAON_FEED:
```
MiguelMedeiros:
```suggestion
										case SUPPORTED_FEED_TYPES.LUGANO_FEED:
```

typo mistake :)

src/utils/widgets/index.ts
```diff
@@ -6,6 +6,7 @@ export enum SUPPORTED_FEED_TYPES {
 	HEADLINES_FEED = 'news.headlines',
 	BLOCKS_FEED = 'bitcoin.stats',
 	FACTS_FEED = 'quotes',
+	LUGAON_FEED = 'to.synonym.lugano-feed',
```
MiguelMedeiros:
```suggestion
	LUGANO_FEED = 'to.synonym.lugano-feed',
```
typo mistake :)

MiguelMedeiros:
@Nuhvi,

I've observed an inconsistency with the timestamp format on the client side. It currently displays as HH:M (e.g., 6:4). For better clarity, I believe it should be in the HH:MM format, such as 6:40.

Could you please confirm this?



simulator_screenshot_36B12486-AE92-46AF-AF3B-0FB8FF7ED42E

package.json
```diff
@@ -67,7 +67,7 @@
     "@synonymdev/slashtags-widget-facts-feed": "1.1.0",
     "@synonymdev/slashtags-widget-news-feed": "1.1.0",
     "@synonymdev/slashtags-widget-price-feed": "1.1.0",
-    "@synonymdev/web-relay": "1.0.0",
+    "@synonymdev/web-relay": "^1.0.2",
```
MiguelMedeiros:
```suggestion
    "@synonymdev/web-relay": "1.0.2",
```
Let's remove the ^ to ensure we have explicit control over dependency versions.

MiguelMedeiros:
@Nuhvi  I noticed that when we add the widget, or when we click to see its settings, a screen appears with part of the widget appearing behind another layer of the app. Is it possible to make everything fit there without the user having to scroll down? Otherwise, I think it would be better if the user could scroll down to see the preview of the widget.

edit: im using iphone 14 simulator

simulator_screenshot_98F490D3-97A7-402B-A3FD-647B473F54E8

Nuhvi:
> @Nuhvi I noticed that when we add the widget, or when we click to see its settings, a screen appears with part of the widget appearing behind another layer of the app. Is it possible to make everything fit there without the user having to scroll down? Otherwise, I think it would be better if the user could scroll down to see the preview of the widget.
> 

I agree. However, this is out of the scope of this widget so I don't want to couple the delivery of this widget with changes to how the Edit Widget screen behaves. Happy to do what @pwltr and @limpbrains suggest.

MiguelMedeiros:
> > 
> I agree. However, this is out of the scope of this widget so I don't want to couple the delivery of this widget with changes to how the Edit Widget screen behaves. Happy to do what @pwltr and @limpbrains suggest.

I agree with you that this change would be out of scope for this PR. I don't want to add any unnecessary complexity to the PR, so I'll create an issue to track this change and we can address it in a separate PR. 👍 

edit: created an issue here https://github.com/synonymdev/bitkit/issues/1249

MiguelMedeiros:
I've noticed an issue in the calendar events table where the event titles are being truncated in the middle of the text, resulting in something like "Bitcoin p...ments are." 

To improve the user experience, I suggest modifying the text truncation behavior to cut off the end of the string instead. For example, it should display as "Bitcoin payments..." 

I believe there's a function in the code that handles string truncation to fit within the screen, but it seems optimized for invoices, where we need to see both the beginning and end of the string. In this case, it would greatly enhance the user's ability to read and understand the event title if only the beginning of the presentation title is displayed. Do you agree, @Nuhvi ?

simulator_screenshot_7D9CB60E-46E9-4307-A55D-48A9A697529C

src/components/LuganoFeedWidget.tsx
```diff
@@ -0,0 +1,247 @@
+import React, { memo, ReactElement, useEffect, useState } from 'react';
+import { View, StyleSheet, StyleProp, ViewStyle, Image } from 'react-native';
+
+import { Caption13M, Text01M, Text02M } from '../styles/text';
+import BaseFeedWidget from './BaseFeedWidget';
+import { TouchableOpacity } from 'react-native-gesture-handler';
```
pwltr: Rather import from `react-native` as this one will cause issues

```suggestion
import { TouchableOpacity } from 'react-native';
```

src/components/LuganoFeedWidget.tsx
```diff
@@ -0,0 +1,247 @@
...
...WMBLzxa0tw6lh7l+9Sw8QXz53Jt7O3RNu+1a9j9/MviaGnKPeMBZH0b1gGzlXA6AV/iaqGKfAB3i0T1/Taemx7/NkuKeJ0KK6pFCSonjYJZwqKM5Dgj2zeak36YJGFJuWzyAWJLpozCyPm3Apf4OU6geT173ruFzkfX1xdifL6SzHi1uvTDilTQZqm9Z33cPCdjTjKBW96ujVBRU//ssL9VQ2TeO+/Fz7B24Uis5EgWzUOkaF6+mRhIlJBvX3Hv/2U5xh2pqh4I3g5SdkigppiXnXdZIxZcA98AxFgB9/XwUPsfekNITANs+ijMQ92m72//JGHIk4lDSHJvzAZRqXmlV3Xy8zp91IDyPfDb5mO2nvvgIc4RKOAQ/AUw7lKr3/qYdGrbaSrGUjdUwTRxon9B1LSyyabF65TFCZ8ZMCnqKwMtffM8VX4hAQRYCM8BiWLJiz9W+HopQo+NYcQ/dy95QsneeiJHoIQcvJH6ahajPvHb3ASToPHaKSKIMqp3g2PQd7+2IfmCZW+k8LCKhiU6gdkc3fsdLM+1j5eLzpZNKgUF905EyNmeS14jQ2aAHmBoVwL0Nb0+nrs5g5itqggRRdt9IsPFPt0JCFS0PdkckbwHt+vrAR5grXkBA7mb8gRHIw8w+H3P0TzG2/jCN/beYKBw6pw+T89JDSHDcon2PUFJOiy0bVC8OU8LZ1HghcIAqXvOAnW8N1U5DHatu8SwqX+XXP95IzCbH17A88stiDObXwZ9y/yTegWXuLin6UJMJXv1qCA8wnzOQGe0QTIWzNQQFokuPBH0RuRDeQfKBawHAWcLwhlbVIaBS1wviDBtwbe6Wlv6gGUFliMAvIdi7XwCV3Kbc+HEcF5jLVc+A12QzoC1PEdFboXTEF+IQFL9Wt0KQlTB8xHAfmGB1OFT0hbE8gPXIWCP5BPWKxk2gVZp65yc7CI776NAnwGCX9Rpi+7nbtcIIHXyCp8QrcEIHASUI3z4/C0cug8BA3p52dS+zZyIgBBjQ5WoDBEdB1rIYWfyOXgnCeG0AXVGJJGCAkU4Bqo15On35Cr8Ak5awAbShuQWViAglnoa+g8zhMqmtsjnBEBbHAyzOM7mgnhx1yo/AAJmiq5aKi+4UwEr9FrAthQdQW00XzuOrUVIoq9QwJG4BvVvldCt/EnMDxFI4N+JxQAAAAASUVORK5CYII=';
```
pwltr, line 26: Why hardcode the icon for this widget? Currently there is no icon on the overview and preview screen. Also the widget title probably needs to be cleaned up.

https://user-images.githubusercontent.com/5798170/267653830-b671c69e-d160-49f7-bc28-74aaaecbcc9a.png

src/screens/Widgets/Widget.tsx
```diff
@@ -131,13 +132,7 @@ const Widget = ({
...
 
-						{loading && (
-							<ThemedView style={styles.previewLoading} color="white08">
-								<Spinner />
-							</ThemedView>
-						)}
-
-						{!loading && (
+						{
 							<>
```
pwltr: Unnecessary nesting, remove

```
{
<>
...
</>
}
```

src/hooks/widgets.ts
```diff
@@ -40,36 +45,42 @@ export const useSlashfeed = (options: {
...
+					reader.config.type === SUPPORTED_FEED_TYPES.FACTS_FEED ||
+					reader.config.type === SUPPORTED_FEED_TYPES.HEADLINES_FEED ||
+					reader.config.type === SUPPORTED_FEED_TYPES.LUGANO_FEED
 				) {
 					setLoading(false);
 					return;
 				}
 
 				const promises = _fields.map(async (field) => {
+					// TODO: Use reader.getField(field.name) after merging and publishing https://github.com/synonymdev/slashtags-feeds/pull/11
```
pwltr: This is merged now

src/hooks/widgets.ts
```diff
@@ -12,6 +12,9 @@ type Field = {
 	unit?: string;
 };
 
+// Cache config files to reduce widgets layout shifts.
+const cache: { [url: string]: { config: SlashFeedJSON } } = {};
```
pwltr, line 16: Can we use the web relay client store as a caching layer since it already has all the data? Or what is the purpose of having it? Curious if you say this makes sense. Maybe the client/reader can be smart and use the store as cache by default (and take a param to skip the cache). I think this is a feature every user of the lib would like to have so should be included if possible.

pwltr (changes requested):

src/components/LuganoFeedWidget.tsx
Nuhvi: Hardcode because Bitkit still can't reliably load png images (raw binary), and converting this icon into SVG was an enormous pain that didn't work out in the end.

What do you mean by cleaning up the title?

src/hooks/widgets.ts
Nuhvi: Not published yet, will check with @rbndg

src/hooks/widgets.ts
Nuhvi: We can, but because it is async and on disk React keeps having flashes of empty images until the web relay client loads the image. This cache is blocking and in memory.

Let me know if I should remove it and you don't care about such flashes.

Nuhvi: I can make an in-memory cache in the web relay, but the interface is still async, which means React will still render the empty image slot then hydrate it it when the promise resolves on the nextTick, so again you will probably get this flash.

src/components/LuganoFeedWidget.tsx
pwltr: Okay, for the title it should probably be 'Plan B Forum' instead of 'PlanB forum'. But can hardcode this as well since we're already doing that for the image.

src/hooks/widgets.ts
pwltr: I'm okay with leaving it this way, doesn't need to happen now. I feel like loading the image async from the store would allow us to show the fallback image while loading but we can try that another time.

pwltr (approved):

limpbrains (approved):

PR #1248, opened 2023-09-11
chore: remove native libsodium

pwltr:
### Description

Remove `react-native-libsodium` libs as they are not actually used.

### DIFF
--- a/.github/workflows/react-native-skia-stub.patch
+++ b/.github/workflows/react-native-skia-stub.patch
@@ -5,7 +5,6 @@ index 935e19e8..fa64a969 100644
 @@ -178,7 +178,8 @@
      "stream": "stream-browserify",
      "vm": "vm-browserify",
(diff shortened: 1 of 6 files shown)


MiguelMedeiros (approved):
Hey @pwltr, I've tested your PR and noticed that some scripts from `package.json` were removed, but they were still mentioned in the documentation files. To maintain consistency, I've made the necessary adjustments in the READMEs, ensuring they match the changes you made.

limpbrains:
So sodium-native actually works in RN, but with pure JS, right?

pwltr:
> pure JS

Slashtags depends on [sodium-universal](https://github.com/sodium-friends/sodium-universal) which uses `sodium-native` in node and `sodium-javascript` on the browser. So Bitkit is using `sodium-javascript` which works but is slow and blocking the UI thread in RN.

limpbrains (approved):

PR #1272, opened 2023-09-25
test(scan): improve scan tests

pwltr:
### Description

Improved tests for scanning feature

### Type of change

### Tests

- [x] Detox test
- [ ] Unit test
- [ ] No test

### QA Notes


src/screens/TreasureHunt/Airdrop.tsx
```diff
@@ -0,0 +1,342 @@
...
+}: TreasureHuntScreenProps<'Airdrop'>): ReactElement => {
+	const interval = useRef<NodeJS.Timer>();
+	const maxInboundCapacitySat = useLightningMaxInboundCapacity();
+	const { chestId } = useSelector((state) => {
+		return viewControllerSelector(state, 'treasureHunt');
+	});
+
+	const { treasureChests } = useAppSelector((state) => state.settings);
+	const chest = treasureChests.find((c) => c.chestId === chestId)!;
+	const { attemptId, state, winType } = chest;
```
coreyphillips, line 57: May need a wrapper or check around this. Managed to receive the following error. "Cannot read property 'attemptId' of undefined". The image is from an error in `Prize.tsx`, but the same logic applies.

Simulator Screenshot - iPhone 14 - 2023-09-26 at 08 48 48

src/screens/TreasureHunt/Airdrop.tsx
```diff
@@ -0,0 +1,342 @@
...
+					return c !== chest ? c : updated;
+				});
+
+				updateSettings({ treasureChests: newChests });
+
+				if (result.state === 'FAILED') {
+					navigation.replace('Error');
+				}
+			}
+		};
```
coreyphillips: If able, we should define these methods outside of useEffect.

src/navigation/bottom-sheet/TreasureHuntNavigation.tsx
```diff
@@ -32,16 +41,97 @@ const screenOptions: StackNavigationOptions = {
...
+					setInitialScreen('Airdrop');
+				} else {
+					setInitialScreen('Chest');
+				}
+			} else {
+				setInitialScreen('Error');
+			}
+
+			setIsLoading(false);
+		};
```
coreyphillips: If able, we should define `getChest` outside of `useEffect`.

src/screens/TreasureHunt/Prize.tsx
```diff
@@ -41,15 +51,33 @@ const xml = `
...
+	const { chestId } = useSelector((state) => {
+		return viewControllerSelector(state, 'treasureHunt');
+	});
 
-	const prize = prizes.find((p) => p.id === id)!;
+	const chest = treasureChests.find((c) => c.chestId === chestId)!;
+	const chests = treasureChests.filter((c) => !c.isAirdrop);
+	const chestIndex = chests.indexOf(chest);
+
+	const { attemptId, state, winType } = chest;
```
coreyphillips, line 68: May need a wrapper or check around this. Managed to receive the following error. "Cannot read property 'attemptId' of undefined".

Simulator Screenshot - iPhone 14 - 2023-09-26 at 08 48 48

src/screens/TreasureHunt/prizes.ts
```diff
@@ -1,124 +1,154 @@
...
+
+export const prizes: TPrize[][] = [
+	[
+		{
+			winType: 'consolation',
+			amount: 100,
+			image: require('../../assets/treasure-hunt/consolation-1.jpg'),
+			title: 'Lightning Strikes',
+			description:
+				'Clouds gather, and a huge thunder bolt hits the treasure chest, liquifying most of the coins. Only 100 sats are left intact.',
```
coreyphillips:
```suggestion
				'Clouds gather, and a huge thunderbolt hits the treasure chest, liquifying most of the coins. Only 100 sats are left intact.',
```

src/utils/scanner.ts
```diff
@@ -382,12 +385,23 @@ export const decodeQRData = async (
...
 		if (chestId) {
+			setFeedWidget({
+				url: LuganoFeedURL,
+				type: SUPPORTED_FEED_TYPES.LUGANO_FEED,
+				fields: [],
+			});
+
 			return ok([{ qrDataType: EQRDataType.treasureHunt, chestId }]);
 		}
 	}
```
coreyphillips: I don't think it's needed because I don't think it's likely, but it's possible some users may try testing Bitkit on testnet during the event. So for these two conditions, we may want to check that unless `__DEV__` they are using mainnet. If they are not using mainnet, we can display a toast message suggesting that they switch to it in Settings.

coreyphillips (changes requested):

src/screens/TreasureHunt/Prize.tsx
pwltr: Should be fixed by moving logic to reducer.

src/utils/scanner.ts
pwltr: Added checks

src/screens/TreasureHunt/Airdrop.tsx
pwltr: I wouldn't bother with optimizing performance for this code since it will be probably be removed again soon, if we do end up keeping it I'll clean it up in a couple of places. Also these functions are defined in an onMount useEffect so they will only ever be instantiated once anyways.

coreyphillips (approved):

PR #1284, opened 2023-10-02
chore(settings): add export store dev action

pwltr:
### Description

Adds a menu item to developer settings for exporting redux store.


src/screens/Settings/DevSettings/index.tsx
```diff
@@ -47,6 +49,29 @@ const DevSettings = ({
 		return warningsSelector(state, selectedWallet, selectedNetwork);
 	});
 
+	const exportStore = async (): Promise<void> => {
+		const time = new Date().getTime();
+		const store = JSON.stringify(getStore());
```
limpbrains: I suggest format it nicely before export

limpbrains (approved):

Issue #770, opened 2022-12-13
Internal node not working


louneskmt:
### Describe the bug

Since a few days, the LDK node doesn't seem to be working:
- I'm no longer able to see the node id in "Advanced > Lightning Node" 
- The only channel I had is detected as "Closed connection" while there is no closing tx on the chain

The channel is marked as "Channel Ready Yes", "Channel Usable False". The local balance of the balance no longer appears on the main screen balance + Spending balance. 

### Reproduce

There isn't really any step to reproduce. 

### Screenshots / Recording

![AD26AF68-D8E2-4ABD-B188-A870A1792A34](https://user-images.githubusercontent.com/38882571/207255875-b1308ac2-83d3-4d92-88b2-100395752107.png)
![4B2F4EEC-F624-45FC-AE46-91B7EC0E5784](https://user-images.githubusercontent.com/38882571/207255957-a3b85f61-1c6f-4780-b97f-f92a43b181c4.png)

### Operating system

iOS 16

### Bitkit version

1.0 (23)

coreyphillips:
Hey @louneskmt, thanks for creating an issue. Sorry to read you're having trouble with the lightning node's connection. 

There are some instances where using a VPN will prevent a successful connection to your lightning node's peers. To prevent this, please ensure you're not using a VPN service on your device and attempt to restart the application to reconnect.

If this isn't the case, another way to troubleshoot this issue would be to enable "Developer mode" and attempt to restart and refresh LDK from the Settings menu. To enable dev mode, tap the Cog at the bottom of the Settings menu ~5 times. If successful, you should see a notification appear letting you know dev options were enabled. From there, you can navigate to "Advanced"->"Lightning Connections". You should now see several new buttons. Two of which are labeled "Restart LDK" & "Refresh LDK". Tapping "Restart LDK", wait several seconds and then tapping "Refresh LDK" should attempt to re-add and connect to your lightning peers.

Please let us know if this helps resolve your issue or if you notice any additional connection issues.

louneskmt:
Hey @coreyphillips! You're right, I use a VPN on my phone at times. Unfortunately, I tried your protocol multiple times, either on 5G or WiFi, without VPN, and nothing changed. Is there a way to get the logs?

coreyphillips:
[...]

Issue #713, opened 2022-11-17
[Bug]: Funds lost after channel close


steepdawn974:
### Describe the bug

After having closed a Lightning connection, the funds that were in the channel never reappear in the on-chain wallet.  

Under Networks > Ligthning Connections the channel is under "Closed Connections", so it appears it has properly closed. 
I imported the seed words into a third-party wallet to double-check, and can confirm that the funds never arrived.

Where can I provide further info (channelId etc) to help troubleshoot?

### Reproduce

1. Open a LN channel (Blocktank) with wallet on-chain funds
2. LN Balance appears
3. Close this Lightning Connection
4. Channel is closed, but Funds are lost

### Operating system

CalyxOS 4.3.0 (Android13)

### Bitkit version

v1.0.0

Jasonvdb:
Can you please confirm the version in the app? Go to Settings -> About Bitkit. It should be version `1.0 (19)` for the latest. 

We just added a pending balance to your total so you should see that there for a while after the channel closes. It's likely a mutual close didn't happen resulting in a force close and the funds could take 2 weeks to be claimable. You won't see it in your onchain wallet even if you import it elsewhere until the outputs become spendable again so the wallet can claim it.

If you're on the latest version and not seeing a pending balance above your total on the home screen then send us your channel ID or Blocktank order ID (Settings -> General -> Blocktank Orders) to `support@synonym.to` and we can check on our side what the status is.

steepdawn974:
It reads: 
`Version: 1.0.0`  


EDIT: I see a new update just dropped. Installing `1.0 (19)` now

Jasonvdb:
Cool let me know if you see a pending balance now?

steepdawn974:
Negative. Still no pending balance on `v1.0 (19)`.   Will engage support@synonym.to now.

Jasonvdb:
@steepdawn974 just confirming you didn't try restore your wallet into Bitkit? It's still the original one you created on your device? Just want to rule out a bug in the backup method.

steepdawn974:
Correct- It was a first-time fresh install of 1.0.0 back then. Create a fresh wallet, transferred some sats, opened a channel, then two weeks later closed the channel --> remaining channel balance is not showing up in the on chain wallet, also not a pending. 
Then upgraded to 1.0 (19) as suggested on Github --> still no fund pending. 

SD 

-- 
 Sent with Tutanota, enjoy secure & ad-free emails. 



[...]

PR #1294, opened 2023-10-06
feat: new ldk backup server restore

Jasonvdb:
### Description

- Enables backing up and restoring from new backup server.
- Reverts back to deprecated backup server if no backup found on new server

### Linked Issues/Tasks

Does not resolve existing restore issues it should prevent further ones from occurring.
https://github.com/synonymdev/bitkit/issues/713
https://github.com/synonymdev/bitkit/issues/770

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA Notes

Backup and restoring lightning funds should be reliable now


limpbrains:
should performRemoteLdkBackup be removed?

Also list in  Settings -> Backup and Restore screen doesn't represent actual state now. We need a way to get backup info from ldk to update it

src/store/actions/backup.ts
```diff
@@ -179,12 +184,86 @@ export const performRemoteBackup = async <T>({
...
+
+	if (backupSetupRes.isErr()) {
+		return err(backupSetupRes.error);
+	}
+
+	//Check if backup exists on new server
+	const fileListRes = await ldk.backupListFiles();
+	if (fileListRes.isErr()) {
+		return err(fileListRes.error);
+	}
```
limpbrains, line 236: this throws "Error: Request failed. Missing response from backup server."
I believe we should return { backupExists: false }  in this case

limpbrains (changes requested):

src/store/actions/backup.ts
Jasonvdb: Might be a server issue here, if there is no backup then the list is just empty. I don't want to assume an error means there's no backup. I'll get back to you on this.

Jasonvdb:
> should performRemoteLdkBackup be removed?

I was thinking let me remove it right after we've done a few rounds of live testing. Just in case.

> Also list in Settings -> Backup and Restore screen doesn't represent actual state now. We need a way to get backup info from ldk to update it

True. Now we don't really have a last backup time to display. I do have a self check function for the backups maybe we can  run this when this view is loaded to show the state.

src/store/actions/backup.ts
limpbrains: I'm testing with server from development.template. I've created a wallet using master branch, then I'm trying to restore in using new-ldk-backups and I'm getting this error

limpbrains (approved):

src/store/actions/backup.ts
Jasonvdb: I've updated the env with new backup server. SSL was required as well as a backend update.

PR #1323, opened 2023-10-13
fix(receive): update minimum invoice amount

pwltr:
### Description

Updates the minimum for an order. The minimum is the fee for the channel which is a dynamic value. For now anything greater than ~1092 is okay. Also updates max amount (half the max channel size) and adds a missing migration.

### Linked Issues/Tasks

https://app.asana.com/0/0/1205713367218911/f

### DIFF
--- a/src/screens/Wallets/Receive/ReceiveAmount.tsx
+++ b/src/screens/Wallets/Receive/ReceiveAmount.tsx
@@ -22,7 +22,6 @@ import { getNumberPadText } from '../../../utils/numberpad';
 import { createCJitEntry } from '../../../utils/blocktank';
 import { showToast } from '../../../utils/notifications';
 import { DEFAULT_CHANNEL_DURATION } from '../../Lightning/CustomConfirm';
-import { primaryUnitSelector } from '../../../store/reselect/settings';
 import { blocktankInfoSelector } from '../../../store/reselect/blocktank';
 import { refreshBlocktankInfo } from '../../../store/actions/blocktank';
 import type { ReceiveScreenProps } from '../../../navigation/types';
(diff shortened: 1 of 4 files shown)


limpbrains (approved):

src/screens/Wallets/Receive/ReceiveAmount.tsx
```diff
@@ -38,9 +42,9 @@ const ReceiveAmount = ({
 	const unit = useSelector(primaryUnitSelector);
 	const blocktank = useSelector(blocktankInfoSelector);
 
-	const { minChannelSizeSat, maxChannelSizeSat } = blocktank.options;
+	const { maxChannelSizeSat } = blocktank.options;
 	// Subtract from max to keep a buffer for dust
-	const maxInvoiceSats = maxChannelSizeSat - minChannelSizeSat;
+	const maxInvoiceSats = maxChannelSizeSat - MINIMUM_AMOUNT;
```
pwltr: Not sure what is the point of the buffer, the invoice will be paid via LN so dust shouldn't play a part here? @coreyphillips Can you elaborate?

coreyphillips: The comment is not correct and can be removed, but I believe this was initially added to prevent Blocktank from denying the CJIT channel opens in older versions of it. We can most likely remove it now and test that it continues to work and open as expected.

coreyphillips (approved):

src/screens/Wallets/Receive/ReceiveAmount.tsx
pwltr: Changed maxAmount and a added missing migration

coreyphillips (approved):

PR #1346, opened 2023-10-18
fix(lightning): prevent refresh from being called multiple times 

Jasonvdb:
### Description

Prevent refresh from being called multiple times simultaneously. Promises will still be resolved normally on success/error.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### DIFF
--- a/src/utils/lightning/index.ts
+++ b/src/utils/lightning/index.ts
@@ -464,6 +464,25 @@ export const unsubscribeFromLightningSubscriptions = (): void => {
 	onSpendableOutputsSubscription?.remove();
 };
 
+let isRefreshing = false;
+let pendingRefreshPromises: Array<(result: Result<string>) => void> = [];
+
+const resolveAllPendingRefreshPromises = (result: Result<string>): void => {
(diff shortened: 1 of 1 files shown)


src/utils/lightning/index.ts
```diff
@@ -464,6 +464,24 @@ export const unsubscribeFromLightningSubscriptions = (): void => {
 	onSpendableOutputsSubscription?.remove();
 };
 
+let isRefreshing = false;
+let pendingRefreshPromises: Array<(result: Result<string>) => void> = [];
+
+const resolveAllPendingRefreshPromises = (result: Result<string>): void => {
+	while (pendingRefreshPromises.length > 0) {
```
limpbrains, line 472: I believe `isRefreshing = false` is missing here

Jasonvdb: you're right thanks just fixed

limpbrains (approved):

coreyphillips (approved):

PR #1351, opened 2023-10-19
feat(lightning): Add Custom RGS Server

coreyphillips:
### Description

- Adds ability for users to add custom RGS servers in Advanced Settings.
- Will currently attempt to connect to any valid url passed as the RGS server.

### Type of change

### Tests

- [x] No test

### Screenshot / Video

Simulator Screenshot - iPhone 14 - 2023-10-19 at 20 22 27
Simulator Screenshot - iPhone 14 - 2023-10-19 at 20 22 36


pwltr (approved):

Jasonvdb (approved):
We should maybe indicate a restart is required to take affect. Also when switching servers incremental syncs won't work so on first startup it'll error and return empty graph but on next startup sync from scratch. So maybe best to stop ldk, delete current graph cache and then restart.

coreyphillips:
Added a notification to alert the user that they may need to update the app once or twice to apply the change. It will currently stop the node and restart LDK with the new RGS info, but I don't believe we have exposed a method to delete the net graph info just yet.

PR #1378, opened 2023-11-14
chore: update lockfile

pwltr:
### Description

I updated to latest node version 20.9.0 LTS and had to reset `yarn.lock` to be able to install dependencies.

@limpbrains @coreyphillips Can you check if this lock file works for your environment?

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -76,9 +76,9 @@ PODS:
   - lottie-react-native (5.1.6):
     - lottie-ios (~> 3.4.0)
     - React-Core
-  - MMKV (1.3.1):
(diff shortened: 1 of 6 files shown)


limpbrains (changes requested):
I'm getting `TypeError: Cannot read property 'get' of undefined` when I start the app
I believe it is `const node = await drive.files.get('/profile.json')` 
Probably beucase slashtags sdk has been upgraded

pwltr:
Yeah I noticed as well this error popped up again too, I think it's one of the hyperswarm dependencies.
image

Probably we should find the package causing this and lock it in `package.json`.

PR #1381, opened 2023-11-16
fix: refactor contact add

limpbrains:
### Description

- upgrade synonymdev/slashtags-url
- refactor contact add form so that user can type slash url + relay

### Type of change

Bug fix

### Tests

Detox test

### DIFF
--- a/src/screens/Contacts/AddContact.tsx
+++ b/src/screens/Contacts/AddContact.tsx
@@ -2,9 +2,8 @@ import React, { ReactElement, useState } from 'react';
 import { View, StyleSheet, TouchableOpacity } from 'react-native';
 import { StackNavigationProp } from '@react-navigation/stack';
 import Clipboard from '@react-native-clipboard/clipboard';
-import { useTranslation } from 'react-i18next';
-import { SlashURL } from '@synonymdev/slashtags-sdk';
 import { parse } from '@synonymdev/slashtags-url';
+import { useTranslation } from 'react-i18next';
 
 import { closeBottomSheet } from '../../store/actions/ui';
 import { handleSlashtagURL } from '../../utils/slashtags';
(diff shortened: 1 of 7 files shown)


pwltr:
Couple of findings:
- In the design the text has the brand color on error state 

image

- The field is overlaying the button when a valid url is passed

Simulator Screenshot - iPhone 15 - 2023-11-28 at 18 19 59

- There is a slight layout shift when the error message appears
- I think the paste button shouldn't require pressing 'Add' again if the url was valid
- The position for the error text is a bit suboptimal when it comes to multi-lang support. Quickly tested german text and it wouldn't fit in one row together with the label on the left.

limpbrains:
> In the design the text has the brand color on error state
>  The field is overlaying the button when a valid url is passed
> There is a slight layout shift when the error message appears
> I think the paste button shouldn't require pressing 'Add' again if the url was valid

fixed

> The position for the error text is a bit suboptimal when it comes to multi-lang support. Quickly tested german text and it wouldn't fit in one row together with the label on the left.

Agree. I've added text wrap. Now if error message doesn't fit, it will appear on the new line

pwltr (approved):

src/screens/Contacts/AddContact.tsx
```diff
@@ -27,57 +28,64 @@ const AddContact = ({
...
-		setError(false);
+	const handleChangeUrl = (contactUrl: string): void => {
+		setUrl(contactUrl);
+		setError(undefined);
+	};
 
-		if (url === '') {
+	const handleAddContact = (contactUrl?: string): void => {
+		contactUrl = contactUrl ?? url;
+		console.info('handleAddContact', contactUrl);
```
pwltr: console.log

limpbrains: oops

PR #1388, opened 2023-11-21
fix: replace hard coded MINIMUM_AMOUNT with blocktank.options.minChannelSizeSat

limpbrains:
fix: replace hard coded MINIMUM_AMOUNT with blocktank.options.minChannelSizeSat

### Description

We should use blocktank.options.minChannelSizeSat instead of hard coded MINIMUM_AMOUNT

### Linked Issues/Tasks

#1323

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -392,7 +392,7 @@ PODS:
     - React-Core
   - react-native-keep-awake (1.2.2):
     - React-Core
-  - react-native-ldk (0.0.122):
+  - react-native-ldk (0.0.123):
     - React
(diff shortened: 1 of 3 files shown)


pwltr:
The `minChannelSizeSat` is not what you want for this use case, probably need an additional field at the BT `/info` endpoint that gives a reasonable value including onchain fees.

cc @SeverinAlexB

SeverinAlexB:
In theory, the minimum amount is equal to the channel purchase fee. Will verify though if I get time.

PR #1389, opened 2023-11-22
fix: improve ln setup logic

### LINKED ISSUE #1253: Fee for Opening a LN channel does not update properly
### Describe the bug

when we try and create a LN channel, I navigate to custom setup > Enter Custom Amount > Continue > When selecting different sizes of channel ($250 / $500 or $990) the Receiving Capacity Fee does not update for ANY of the amounts. If a user clicks continue then they will be presented with a different fee (depending on the size of channel they chose) which will lead to confusion. 

### Reproduce

1. Go to 'Open LN Channel'
2. Click on 'Custom Setup > Enter Amount (make note of the fee displayed) > Click continue (see new fee)

### Screenshots / Recording

_No response_

### Operating system

Graphene OS Pixel 6A

### Bitkit version

v65 

### Log output

_No response_

limpbrains:
### Description

- move ln quick and transder setup to reselect, add tests
- custom setup ln screen, fetch fees everytime something changes

### Linked Issues/Tasks

closes #1253

### Type of change

Bug fix

### Tests

Unit test

### DIFF
--- a/src/screens/Lightning/CustomSetup.tsx
+++ b/src/screens/Lightning/CustomSetup.tsx
@@ -110,7 +110,9 @@ const CustomSetup = ({
 	const blocktankInfo = useSelector(blocktankInfoSelector);
 
 	const [textFieldValue, setTextFieldValue] = useState('');
-	const [channelOpenFee, setChannelOpenFee] = useState('');
+	const [channelOpenFee, setChannelOpenFee] = useState<{
+		[key: string]: string;
+	}>({});
 	const [showNumberPad, setShowNumberPad] = useState(false);
 	const [loading, setLoading] = useState(false);
 	const [spendingPackages, setSpendingPackages] = useState<TPackage[]>([]); // Packages the user can afford.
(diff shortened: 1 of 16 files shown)


src/screens/Lightning/CustomSetup.tsx
```diff
@@ -110,7 +110,7 @@ const CustomSetup = ({
 	const blocktankInfo = useSelector(blocktankInfoSelector);
 
 	const [textFieldValue, setTextFieldValue] = useState('');
-	const [channelOpenFee, setChannelOpenFee] = useState('');
+	const [channelOpenFee, setChannelOpenFee] = useState<{}>({});
```
pwltr: Should avoid using `{}` type

```suggestion
	const [channelOpenFee, setChannelOpenFee] = useState<{
		[key: string]: string;
	}>({});
```

src/utils/wallet/constants.ts
```diff
@@ -14,6 +14,10 @@ export const CHUNK_LIMIT = 15;
 //How much of the users funds we allow to be used for Lightning.
 export const SPENDING_LIMIT_RATIO = 0.8;
 
+export const DIFF = 0.01;
```
pwltr: Since we're export this variable now can we it a better name or add a comment so it's clear what it is for?

src/store/reducers/blocktank.ts
```diff
@@ -9,6 +9,7 @@ const blocktank = (
 ): IBlocktank => {
 	switch (action.type) {
 		case actions.UPDATE_BLOCKTANK_INFO:
+			action.payload.options.max0ConfClientBalanceSat = 0;
```
pwltr: Why is this needed? The value from BT info is wrong?

pwltr (changes requested):

pwltr:
Not sure if it's a regression but when I try to select the largest card I can't continue with the order:

https://github.com/synonymdev/bitkit/assets/8538369/578b661f-7ec0-4d47-958e-dcc11fa4ae70

src/utils/wallet/constants.ts
limpbrains: LIGHTNIG_DIFF, lol, I can't think of anything better

src/store/reducers/blocktank.ts
limpbrains: oops, that was debug code

limpbrains:
> Not sure if it's a regression but when I try to select the largest card I can't continue with the order:

It is not, I believe bug existing for quite a while

src/utils/wallet/constants.ts
pwltr: better :)

pwltr (approved):

PR #1391, opened 2023-11-28
fix: remove detachInactiveScreens={false}

limpbrains:
### Description

We don't need this hack anymore, it is fixed in RN 0.72

### Type of change

Refactoring

### Tests

No test


pwltr:
Nice, was gonna try this today as well 🎉

pwltr (approved):

PR #1392, opened 2023-11-28
refactor: use react-native-quick-crypto, remove rn-nodify

pwltr:
### Description

Replaces `crypto` calls with `react-native-quick-crypto` which should offer way better performance since it's using JSI/C++. Allows us to remove `rn-nodify` and for Node core module polyfilling. Also replaces `text-encoding` with [faster](https://github.com/anonyco/FastestSmallestTextEncoderDecoder#benchmarks) alternative `fastestsmallesttextencoderdecoder`. Renames `.node-version` file so it can be picked up by version managers.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA Notes


package.json
```diff
@@ -30,11 +29,9 @@
     "prepare": "husky install",
     "postinstall": "node postinstall.js"
   },
-  "engines": {
-    "node": ">=16"
-  },
   "dependencies": {
     "@bitcoinerlab/secp256k1": "1.0.5",
+    "@craftzdog/react-native-buffer": "^6.0.5",
```
limpbrains: please exact versions in dependencies

limpbrains (changes requested):

ios/Podfile.lock
```diff
@@ -76,9 +76,9 @@ PODS:
   - lottie-react-native (5.1.6):
     - lottie-ios (~> 3.4.0)
     - React-Core
-  - MMKV (1.3.1):
-    - MMKVCore (~> 1.3.1)
-  - MMKVCore (1.3.1)
+  - MMKV (1.3.0):
+    - MMKVCore (~> 1.3.0)
+  - MMKVCore (1.3.0)
```
limpbrains: I'm qurious why MMKV version changed? I don't have it on my local maching when I do `pod install` maybe you should try `pod install --repo-update` ?

package.json
pwltr: done

ios/Podfile.lock
pwltr: done, but now it updated it to newer versions. Not sure why it always downgrades some packages on my machine.

limpbrains (approved):

coreyphillips:
Seems to break being able to set custom Electrum servers in Settings->Advanced->Electrum Server

coreyphillips (approved):

PR #1396, opened 2023-11-29
chore: replace rimraf with del-cli

pwltr:
### Description

Replaces `rimraf` with [del-cli](https://github.com/sindresorhus/del-cli) which has support for globs on Windows. Also run using `npx` so we can clean even if `node_modules` isn't there.


flaviomoceri (approved):
It works perfectly, before I couldn't run the commands because I was getting errors caused by `rimraf` using Window, like this: 

Screenshot 2023-11-29 141148 (1)


However, if I use the `del-cli` I don't get any errors:

Screenshot 2023-12-01 115117

limpbrains (approved):

Issue #1397, opened 2023-11-30
[Bug]: Profile links don't open if I have the relevant application


flaviomoceri:
### Describe the bug

If you add a valid link to your profile for example https://www.instagram.com/{username}, if the "Instagram" application is not present on your device, if you click on it it opens the link correctly via browser, but if instead I have the application installed on the device, even if I click it nothing happens.

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/63161412/279a87f4-cb73-4039-bae2-f6891b1cf13b

### Operating system

Android 13

PR #1398, opened 2023-11-30
fix(profile): #1397 fix links not opening in app

flaviomoceri:
### Description

If you add a valid link to your profile for example https://www.instagram.com/{username}, if the "Instagram" application is not present on your device, if you click on it it opens the link correctly via browser, but if instead I have the application installed on the device, even if I click it nothing happens.

This PR fixes this bug - https://github.com/synonymdev/bitkit/issues/1397

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/63161412/e39c251f-67ee-407a-9c36-e4f84b0825f1

### DIFF
--- a/src/components/ProfileLinks.tsx
+++ b/src/components/ProfileLinks.tsx
@@ -5,13 +5,13 @@ import {
 	StyleProp,
 	StyleSheet,
 	ViewStyle,
+	Linking,
 } from 'react-native';
 import { useTranslation } from 'react-i18next';
 
 import { Caption13Up, Text02M, Text02S } from '../styles/text';
 import { TrashIcon } from '../styles/icons';
 import { LocalLink } from '../store/types/slashtags';
-import { openURL } from '../utils/helpers';
 import { editLink, removeLink } from '../store/actions/slashtags';
 import LabeledInput from './LabeledInput';
 import Divider from './Divider';
@@ -31,6 +31,14 @@ const trimLink = (link: LocalLink): string => {
 	}
 
 	return trimmedUrl.replace('https://', '').replace('www.', '');
+};
+
+const openURL = async (url: string): Promise<void> => {
+	try {
+		await Linking.openURL(url);
+	} catch (err) {
+		console.log('Cannot open url: ', url);
+	}
 };
 
 const ProfileLinks = ({


pwltr (approved):

src/utils/helpers.ts
```diff
@@ -387,8 +387,8 @@ export const timeAgo = (timestamp: number): string => {
 };
 
 export const openURL = async (url: string): Promise<boolean> => {
-	const supported = await Linking.canOpenURL(url);
 	try {
+		const supported = await Linking.openURL(url);
```
limpbrains: Why are you using .openURL here, should it be .canOpenURL ?

limpbrains (changes requested):

src/utils/helpers.ts
flaviomoceri: Because if the url is valid like `https://www.instagram.com/bitfinex` and I use `.canOpenUrl`, if I don't have the Instagram application and I click on the link it works correctly by opening the browser and the supported variable returns `true`, however if I have the application instagram for some reason even if I click, nothing happens because the supported variable returns `false`.

However, if I use `.openURL` this bug does not occur and returns true if it can display the link via browser or app, or it returns error and prints it.

Instagram is obviously just one example.

Watch the video in this [PR](https://github.com/synonymdev/bitkit/pull/1398#issue-2018370058) and the video in [issue](https://github.com/synonymdev/bitkit/issues/1397#issue-2018358722) and you should see the difference.

limpbrains: Sorry, I wasn't clear.
I'm looking at the code and you are doing same 2 Linking.openURL calls
const supported = await Linking.openURL(url);
and then
await Linking.openURL(url);

Is this correct?

flaviomoceri: Yes, maybe I can write more cleanly so as not to repeat the same code

The first time I need it to check whether the URL is valid or not, if it returns `true` I open it, otherwise it prints what comes back

pwltr: `await Linking.openURL` doesn't return anything so the second call is definitely redundant, see https://reactnative.dev/docs/linking#openurl. That being said I'm not sure it's okay to remove the check with `canOpenURL()` for this helper, it may be needed for non-http(s) URLs. Can you check why `supported` would be `false` for your instagram with `canOpenURL` being there? Doesn't make sense to me maybe there is a bug with that method or we need to do something else?

src/components/ProfileLinks.tsx
```diff
@@ -33,6 +33,17 @@ const trimLink = (link: LocalLink): string => {
...
+const openURL = async (url: string): Promise<boolean> => {
+	try {
+		await Linking.openURL(url);
+		return true;
+	} catch (err) {
+		console.log('Cannot open url: ', url);
+		console.error('Error opening URL: ', err);
+		return false;
+	}
+};
```
pwltr: We don't care about a return value in this case:

```suggestion
const openURL = async (url: string): Promise<boolean> => {
	try {
		await Linking.openURL(url);
	} catch (err) {
		console.log('Cannot open url: ', url);
	}
};
```

pwltr (changes requested):

pwltr (approved):

src/components/ProfileLinks.tsx
```diff
@@ -33,6 +33,15 @@ const trimLink = (link: LocalLink): string => {
 	return trimmedUrl.replace('https://', '').replace('www.', '');
 };
 
+const openURL = async (url: string): Promise<void> => {
+	try {
+		await Linking.openURL(url);
+	} catch (err) {
+		console.log('Cannot open url: ', url);
+		console.error('Error opening URL: ', err);
```
limpbrains: console.error is redundant

limpbrains (approved):

Issue #1399, opened 2023-11-30
[Bug]: Building Bitkit using Window doesn't work


flaviomoceri:
### Describe the bug

I followed the various steps explained in `DEVELOPMENT.md`, but when I try the last command `yarn android` when I get to 99% the build fails, showing me this error

![Screenshot 2023-11-28 155119 (1)](https://github.com/synonymdev/bitkit/assets/63161412/04b5b747-457f-4998-875a-4717ba082cdc)

### Operating system

Android 13

PR #1400, opened 2023-11-30
build: #1399 fix bitkit build if using windows

flaviomoceri:
### Description

Bug: [#1399](https://github.com/synonymdev/bitkit/issues/1399)

By adding packagingOptions in `android/app/build.gradle` using window you should be able to build bitkit correctly


pwltr (approved):
Works on my machine

PR #1401, opened 2023-11-30
feat: show failed BT orders card

limpbrains:
### Description

Fhow failed LN card if there is a failed BT order

### Linked Issues/Tasks

https://app.asana.com/0/1189635884316057/1205966021348821/f

### Type of change

New feature

### Tests

Unit test

### Screenshot / Video

image

### QA Notes

- create new BT order
- wait until it expres

### DIFF
--- a/src/store/reselect/todos.ts
+++ b/src/store/reselect/todos.ts
@@ -16,6 +16,7 @@ import {
 	transferClosingChannel,
 	transferToSavingsTodo,
 	transferToSpendingTodo,
+	btFailedTodo,
 } from '../shapes/todos';
 import {
 	backupVerifiedSelector,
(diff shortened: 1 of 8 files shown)


src/store/reselect/todos.ts
```diff
@@ -81,9 +82,28 @@ export const todosFullSelector = createSelector(
...
+			if (
+				Date.now() - Number(new Date(order.orderExpiresAt)) >
+				7 * 24 * 60 * 60 * 1000
+			) {
+				return false;
+			}
+
+			if (!hide.btFailed) {
+				return true;
+			}
```
pwltr, line 96: Would this mean if I hide this card once it will never show again for any other failed orders?

limpbrains: No. It records timestamp when you hide it. Any new order will be shown

limpbrains: omg, you are right, will fix it

limpbrains: So I've added new test for this case. And everything is correct. This specific check is if hide.btFailed is undefined. And since we are iterating over expired orders, we need to show card anyway

pwltr (approved):

PR #1420, opened 2023-12-11
feat: onboarding e2e test

limpbrains:
### Description

- new onboadrding test
- refactor settings tests: add About, move Security to separate file, add Send with PIN and enable wrong PIN tests

### Type of change

Refactoring

### Tests

Detox test


e2e/settings.e2e.js
```diff
@@ -14,18 +14,18 @@ import {
...
+	'settings-speed',
+	'settings-suggestions',
+	'settings-tags',
+	'settings-backup',
+	'settings-addr-type',
+	'settings-ln-settings',
+	'settings-electrum',
+	'settings-webrelay',
+	'settings-rgs',
+	'settings-dev',
```
pwltr, line 28: There is a way to get the current test name with `const { currentTestName } = jestExpect.getState();`, not sure if it works for our use case though.

pwltr (approved):

Issue #1407, opened 2023-12-04
[Bug]: Edit invoice on "Receive Bitcoin" forces back to "Receive Instantly"


catch-21:
### Describe the bug

If a value is entered on the "Receive Instantly" page it is then not possible to create a "Receive Bitcoin" invoice because the UI is stuck trying to produce a lightning invoice.

### Reproduce

1. Open android app at a new wallet (or probably existing wallet too)
2. Tap "Receive" button in bottom right
3. Tap "Receive Instantly" slider to begin producing lightning invoice
4. Input a valid value (e.g. 55555)
5. Tap "Continue"
6. Tap back arrow **twice** to return to view QR
7. Tap "Receive Instantly" again to toggle back to on-chain Bitcoin
8. Tap "Edit" button to create Bitcoin invoice

**Desired Result:**
"Edit Invoice" page is displayed

**Actual Result:**
"Receive Instantly" page is displayed. If I now tap "Continue" I am forced back into "Receive Instantly" mode.

### Screenshots / Recording

https://drive.google.com/file/d/1nSlxYS2M2TpB924xUSorJsl8RFvxHRTc/view?usp=sharing

### Operating system

Android 13

### Bitkit version

V1.0

JeanlChristophe:
Hi @james-iohk, thanks for sharing this issue.

PR #1429, opened 2023-12-14
fix(receive): #1407 continue to show btc invoice if instant payment is not confirmed

flaviomoceri:
### Description

Issue: https://github.com/synonymdev/bitkit/issues/1407

I moved the logic to generate the order to `ReceiveConnect.tsx` to resolve the error and used `estimateOrderFee` to show the fees without creating an order first.

In `ReceiveAmount.tsx` I had to increase `MINIMUM_AMOUNT` because now the check is done later and if I hadn't done so I could have entered any value and then on the next screen the generated order would have failed.

I believe there are many ways to improve the code, so I would be happy to receive any feedback to improve it.

https://github.com/synonymdev/bitkit/assets/63161412/a34dc173-4f7f-4c0c-afa7-81c49ad3f5c6

### DIFF
--- a/src/screens/Wallets/Receive/ReceiveConnect.tsx
+++ b/src/screens/Wallets/Receive/ReceiveConnect.tsx
@@ -1,4 +1,4 @@
-import React, { memo, ReactElement, useState } from 'react';
+import React, { memo, ReactElement, useEffect, useState } from 'react';
 import { StyleSheet, View } from 'react-native';
 import { useSelector } from 'react-redux';
 import { Trans, useTranslation } from 'react-i18next';
@@ -22,7 +22,6 @@ import { createCJitEntry, estimateOrderFee } from '../../../utils/blocktank';
 import { showToast } from '../../../utils/notifications';
 import { updateInvoice } from '../../../store/actions/receive';
 import { blocktankInfoSelector } from '../../../store/reselect/blocktank';
-import { ICreateOrderRequest } from '../../../store/types/blocktank';
 
 const imageSrc = require('../../../assets/illustrations/lightning.png');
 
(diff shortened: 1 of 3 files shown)


src/screens/Wallets/Receive/ReceiveConnect.tsx
```diff
@@ -26,15 +32,63 @@ const ReceiveConnect = ({
...
+			const estimate = await estimateOrderFee(requestData);
+			if (estimate.isOk()) {
+				setFeeEstimate(estimate.value);
+			} else {
+				console.error('Error in estimate');
+			}
+		} catch (error) {
+			console.error('Error', error);
+		}
+	}
```
pwltr: Should wrap this in `useEffect` so it doesn't run on every render:

```suggestion
	useEffect(() => {
		const getFeeEstimation = async (): Promise<void> => {
			const estimate = await estimateOrderFee({
				lspBalanceSat: amount,
				channelExpiryWeeks: DEFAULT_CHANNEL_DURATION,
			});
			if (estimate.isOk()) {
				setFeeEstimate(estimate.value);
			}
		};

		getFeeEstimation();
	}, [amount]);
```

src/screens/Wallets/Receive/ReceiveConnect.tsx
```diff
@@ -26,15 +32,63 @@ const ReceiveConnect = ({
...
+				title: t('receive_cjit_error'),
+				description: cJitEntryResponse.error.message,
+			});
+			return;
+		}
+		const order = cJitEntryResponse.value;
+		updateInvoice({ jitOrder: order });
 		addCJitEntry(order).then();
 		navigation.navigate('ReceiveQR');
+		setIsLoading(false);
```
pwltr: Only one `setIsLoading(false);` is needed here:

```suggestion
	const onContinue = async (): Promise<void> => {
		setIsLoading(true);

		const cJitEntryResponse = await createCJitEntry({
			channelSizeSat: maxChannelSizeSat,
			invoiceSat: invoice.amount,
			invoiceDescription: invoice.message,
			channelExpiryWeeks: DEFAULT_CHANNEL_DURATION,
			couponCode: 'bitkit',
		});
		if (cJitEntryResponse.isErr()) {
			showToast({
				type: 'error',
				title: t('receive_cjit_error'),
				description: cJitEntryResponse.error.message,
			});
			return;
		}
		const order = cJitEntryResponse.value;
		updateInvoice({ jitOrder: order });
		addCJitEntry(order).then();
		setIsLoading(false);
		navigation.navigate('ReceiveQR');
```

pwltr (changes requested):

src/screens/Wallets/Receive/ReceiveConnect.tsx
pwltr: To avoid the type error update in `types/blocktank.ts`:

```
export interface ICreateOrderRequest {
	lspBalanceSat: number;
	channelExpiryWeeks: number;
	options?: Partial<ICreateOrderOptions>;
}
```

pwltr (approved):

PR #1440, opened 2023-12-19
fix(lightning): lnSetupSelector Fix

coreyphillips:
### Description

- Removes throw from `lnSetupSelector` when onchain balance is zero. This stops the app from crashing when the user has a lightning balance with zero on-chain balance when navigating to the Transfer view.

### Type of change

### Tests

- [x] No test

### DIFF
--- a/__tests__/reselect.ts
+++ b/__tests__/reselect.ts
@@ -99,11 +99,6 @@ describe('Reselect', () => {
 	});
 
 	describe('lnSetupSelector', () => {
-		it('should throw if onchain balance is 0', () => {
-			const state = cloneDeep(s);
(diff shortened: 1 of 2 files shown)


pwltr:
A unit is failing that checks specifically for thos error.

coreyphillips:
Removed the test [here](https://github.com/synonymdev/bitkit/pull/1440/commits/2742fbfa144f9ede989d4a978ddc2d162b04bebc).

I don't see a reason we should throw for that, but we can wait for Ivan to return for his input when setting it up.

pwltr (approved):
I'm not sure either, at the very least the error should be caught and handled if so,

PR #1449, opened 2024-01-01
fix(ui): restore gradients on Android

pwltr:
### Description

Gradients were often not rendered on Android. This updates skia which uses reanimated for animations instead. Seems to fix the issue.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test


src/components/VerticalShadow.tsx
```diff
@@ -23,16 +16,10 @@ const VerticalShadow = (): ReactElement => {
...
-
-	useEffect(() => {
-		runTiming(opacity, 1);
-	}, [opacity]);
-
 	return (
 		<View style={styles.shadowCanvas} onLayout={handleLayout}>
 			<Canvas style={styles.shadowCanvas}>
-				<Rect x={0} y={0} width={width} height={height} opacity={opacity}>
+				<Rect x={0} y={0} width={width} height={height}>
```
pwltr: Not sure why we used a animated value for this @limpbrains? Works the same with a static opacity.

src/components/HorizontalGradient.tsx
```diff
@@ -31,16 +24,10 @@ const HorizontalGradient = ({
...
-
-	useEffect(() => {
-		runTiming(opacity, 0.7);
-	}, [opacity]);
-
 	return (
 		<View style={style} onLayout={handleLayout}>
 			<Canvas style={style}>
-				<Rect x={0} y={0} width={width} height={height} opacity={opacity}>
+				<Rect x={0} y={0} width={width} height={height} opacity={0.7}>
```
pwltr: Same here

src/components/VerticalShadow.tsx
limpbrains: Looks like fade-in amination for gradient to appear.
I've probably added it because gradients wasn't render instantly 
If this patch fixes render delay, we don't need it

limpbrains (approved):

PR #1450, opened 2024-01-02
refactor(wallet): Beignet Migration

coreyphillips:
### Description
This PR migrates a majority of the on-chain wallet logic out of Bitkit.

Implementation details:

Instead of migrating existing user wallet data to match Beignet's data structure I've opted instead to map Bitkit's existing data structure to Beignet's using the [setWalletData](https://github.com/synonymdev/bitkit/blob/0db90a20acb63aca787c3c8522d2ec0a9a2f349b/src/store/actions/wallet.ts#L719) & [getWalletData](https://github.com/synonymdev/bitkit/blob/0db90a20acb63aca787c3c8522d2ec0a9a2f349b/src/store/actions/wallet.ts#L681) methods. The mapping can be seen [here](https://github.com/synonymdev/bitkit/blob/0db90a20acb63aca787c3c8522d2ec0a9a2f349b/src/store/reducers/wallet.ts#L36) in `UPDATE_WALLET_DATA`. This allows all existing calls to Bitkit's store/state to remain intact and reduces the footprint of this PR.
Effectively, everything in the wallet store [here](https://github.com/synonymdev/bitkit/blob/0db90a20acb63aca787c3c8522d2ec0a9a2f349b/src/store/shapes/wallet.ts#L150) is mapped accordingly with the exception of header, exchangeRates & feeEstimates that Beignet uses, but is stored elsewhere in Bitkit.

This PR is meant to act as a 1:1 migration. Updating Bitkit to use the higher level tx and fee API's from Beignet will arrive in a separate PR once this migration is complete.

Regarding the tests. I'm honestly not sure how we should proceed here. I'll need @limpbrains input on the damage I've inflicted once he returns.

### Type of change


yarn.lock
```diff
@@ -1674,7 +1674,7 @@
   resolved "https://registry.yarnpkg.com/@bcoe/v8-coverage/-/v8-coverage-0.2.3.tgz#75a2e8b51cb758a7553d6804a5932d7aace75c39"
   integrity sha512-0hYQ8SB4Db5zvZB4axdMHGwEaQjkZzFjQiN9LVYvIFB2nSUHW9tYpxWriPrWDASIxiaXax83REcLxuSdnGPZtw==
 
-"@bitcoinerlab/secp256k1@1.0.5":
+"@bitcoinerlab/secp256k1@1.0.5", "@bitcoinerlab/secp256k1@^1.0.5":
```
pwltr: We should probably pin dependencies in `beignet` lib as well for better stability.

pwltr:
Will start posting as I go now for the initial click-through test:

- When restoring the onchain wallet I have to pull-to-refresh for the balance to show up, seems to be a regression:

https://github.com/synonymdev/bitkit/assets/8538369/ee0358a9-f09f-4001-bd94-153be5465503

pwltr:
It looks like it's overpaying fees 2x:

- 1 input, 2 outputs, segwit = 141 bytes
- "Normal" setting on regtest should be 2sats/vbyte = 282 sats fee

Simulator Screenshot - iPhone 15 Pro - 2024-01-04 at 16 53 42
Simulator Screenshot - iPhone 15 Pro - 2024-01-04 at 16 53 40

pwltr:
Coming from the main scanner and "Paste QR Code" the "total required" in coin selection is not updated correctly, seems to be a regression:

EDIT: it works with max amount but not when I type in a custom amount
EDIT2: same when coming from "clipboard ease-of-use" feature

image

Also the review screen is broken thereafter, looks like `SendTransaction` is not updated in this case:

image

src/store/actions/wallet.ts
```diff
@@ -141,841 +101,77 @@ export const createWallet = async ({
...
-	selectedNetwork: EAvailableNetwork;
-}): void => {
-	if (!selectedWallet) {
-		selectedWallet = getSelectedWallet();
-	}
-	if (!selectedNetwork) {
-		selectedNetwork = getSelectedNetwork();
+export const updateExchangeRates = async (
+	exchangeRates: IExchangeRates,
+): Promise<Result<string>> => {
```
pwltr: If `exchangeRates` is not required should be marked as such:

```suggestion
export const updateExchangeRates = async (
	exchangeRates: IExchangeRates = {},
): Promise<Result<string>> => {
```

src/store/reducers/wallet.ts
```diff
@@ -12,8 +12,9 @@ const wallet = (
 	state: IWalletStore = defaultWalletStoreShape,
 	action,
 ): IWalletStore => {
-	let selectedWallet = state.selectedWallet;
-	let selectedNetwork = state.selectedNetwork;
+	let selectedWallet = state?.selectedWallet ?? 'wallet0';
+	let selectedNetwork = state?.selectedNetwork ?? 'bitcoin';
```
pwltr: `state` is never undefined here so this can be:

```suggestion
	let { selectedWallet, selectedNetwork } = state;
```

pwltr:
Electrum Server settings falsely say "disconnected", when I switch to my local regtest the balance doesn't update (to 0) no matter what:

EDIT: also when I kill my local regtest network it doesn't give me the toasts for 'disconnected' & 'reconnected', regression (I am cross-testing everything with current master)

https://github.com/synonymdev/bitkit/assets/8538369/3f83a607-8f62-49a1-b484-02598a6032cc

coreyphillips:
[This](https://github.com/synonymdev/bitkit/pull/1450#issuecomment-1877332509) is intentional and due to a check added [here](https://github.com/synonymdev/beignet/blob/2ed07178c479ab9735df853565011aca7962415a/src/utils/transaction.ts#L285). Some Electrum servers will not broadcast such low fee transactions for small transactions at 1 satPerVByte. This can sometimes be seen on our own regtest server as well for small 1 satPerVByte txs. We can tweak the min byte size lower, but need to be careful to avoid failed broadcasts.

pwltr:
> Some Electrum servers will not broadcast such low fee transactions for small transactions at 1 satPerVByte.

That's not the case here, this is a 2 sat/vbyte transaction. If I put a higher fee rate it is still overpaying:

image

Should be 141 * 5 = 705

coreyphillips:
> Electrum Server settings falsely say "disconnected", when I switch to my local regtest the balance doesn't update (to 0) no matter what:
> 
> EDIT: also when I kill my local regtest network it doesn't give me the toasts for 'disconnected' & 'reconnected', regression (I am cross-testing everything with current master)

I'll add an update to the Electrum config component to wipe the transaction history and balance/utxo information when specifying a new server on regtest.

pwltr (changes requested):

pwltr:
If I try to change address types it's crashing with:

image

coreyphillips:
@limpbrains Electrum is now handled through the beignet library. Any thoughts on how the tests should be re-wired to handle this update in Bitkit?

limpbrains:
Looks like electrum instance is already available through getOnChainWalletElectrum. So I believe waiting for connection and closing it after tests would be enought.

limpbrains:
I'm trying to run e2e tests and they are failing because of `Error: Peer's feerate much too low. Actual: 253. Our expected lower limit: 7500`

We need a way to set fees during e2e tests. 
Or please change regtest network fees to
```
	onchain: {
		fast: 4,
		normal: 2,
		slow: 1,
		minimum: 1,
	},
```

Also, now we have 2 places where we fetch fees from mempool.space: in beignet and in bitkit getFeeEstimates function. This is confusing.

We also have fee override in dev settings, would be nice to have to working again

limpbrains:
Send flow doesn't work for me


https://github.com/synonymdev/bitkit/assets/155891/ec3fb6f3-3361-4375-82c9-58bbc9a3f200

coreyphillips:
@limpbrains Fix incoming for the send flow.

yarn.lock
coreyphillips: Resolved as of https://github.com/synonymdev/bitkit/commit/de7b3602e4243de6215b6a15cc4b4bbd4c312c17

src/store/actions/wallet.ts
coreyphillips: Resolved as of [this](https://github.com/synonymdev/bitkit/commit/5ec5503654c09b81cd03c9f063eb5caa5aff4972) commit.

coreyphillips:
> I'm trying to run e2e tests and they are failing because of `Error: Peer's feerate much too low. Actual: 253. Our expected lower limit: 7500`
> 
> We need a way to set fees during e2e tests. Or please change regtest network fees to
> 
> ```
> 	onchain: {
> 		fast: 4,
> 		normal: 2,
> 		slow: 1,
> 		minimum: 1,
> 	},
> ```
> 
> Also, now we have 2 places where we fetch fees from mempool.space: in beignet and in bitkit getFeeEstimates function. This is confusing.

Should be resolved as of this [commit](https://github.com/synonymdev/bitkit/commit/f345df423b370a3c02e11d456376c30037d3db7c). ~~Looking into "fee override in dev settings"~~ Appears to fix fee override as well.

coreyphillips:
> [If I try to change address types it's crashing](https://github.com/synonymdev/bitkit/pull/1450#issuecomment-1877464759)

Resolved as of [this](https://github.com/synonymdev/bitkit/pull/1450/commits/fa165429f454fe22b21dd0b11c30a5e6c6758df2) commit.

limpbrains:
After recovering my onchain wallet, I receive a pop-up of old transaction, but my balance is 0 and I don't see any tx history.
Pull-to-refresh on main screen fixes this and proper balance appear

coreyphillips:
> After recovering my onchain wallet, I receive a pop-up of old transaction, but my balance is 0 and I don't see any tx history. Pull-to-refresh on main screen fixes this and proper balance appear

Should be resolved as of [this commit](https://github.com/synonymdev/bitkit/pull/1450/commits/06e77d8b6d9f62f455107b9451fceaee044fbec4).

limpbrains:
> > After recovering my onchain wallet, I receive a pop-up of old transaction, but my balance is 0 and I don't see any tx history. Pull-to-refresh on main screen fixes this and proper balance appear
> 
> Should be resolved as of [this commit](https://github.com/synonymdev/bitkit/pull/1450/commits/06e77d8b6d9f62f455107b9451fceaee044fbec4).

There is no notification, but balance is still 0 after restoring a wallet

coreyphillips:
Still unsure what's causing the e2e tests to fail. Can't seem to reproduce anything locally.

coreyphillips:
~Receiving issue when trying to pay to a lightning invoice. Ultimately, lightning payments need to be separated from on-chain payments moving forward from this migration. Going to attempt as close to a 1:1 workaround for now.~

Resolved as of [this commit](89521be0691dbe173712061171c98392553002c0).

limpbrains:
I've created a wallet using current master branch.
After I've switched to beignet I can't use Send flow. When I tap Paste Invoice nothing happens and I see the following error in the console:
```
 LOG  [TypeError: Cannot read property 'transaction' of undefined]
 LOG  [Error: Cannot read property 'transaction' of undefined]
 LOG  [Error: Cannot read property 'transaction' of undefined]
 ```

**Comment by coreyphillips** [comment]
> I've created a wallet using current master branch. After I've switched to beignet I can't use Send flow. When I tap Paste Invoice nothing happens and I see the following error in the console:
> 
> ```
>  LOG  [TypeError: Cannot read property 'transaction' of undefined]
>  LOG  [Error: Cannot read property 'transaction' of undefined]
>  LOG  [Error: Cannot read property 'transaction' of undefined]
> ```

Was this a restored wallet from master to beignet or master->beignet upgrade without restore?

**Comment by limpbrains** [comment]
beignet or master->beignet upgrade without restore

**Comment by limpbrains** [comment]
Recieve notification is shown twice for same transaction. First time when wallet sees it, second time when it is confirmed

**Comment by limpbrains** [comment]
I'm playing with Gap limit. I have reached, I have an address with index of 21.
Then I'm starting to send funds to index 0. 
I do not receive notification, which is ok. 
I restart the app.
Balance changing, but activity list still missing tx. Even pull-to-refresh doesn't help

Take a look at the video. Activity list missing tx with amount of 0.00007777


https://github.com/synonymdev/bitkit/assets/155891/e466cea4-c2ac-4439-b792-0aaa4f3a753b

**Comment by coreyphillips** [comment]
Previous issues should be addressed as of [4290b20](https://github.com/synonymdev/bitkit/pull/1450/commits/4290b20d5f747a8e08d215870a8e86da5cbaf1e3) & [48dd353](https://github.com/synonymdev/bitkit/pull/1450/commits/48dd353dd047a622d009caaf563f421430005517)

**Comment by limpbrains** [comment]
All e2e tests are now working on my machine!
Also I've fixed detox not exiting issue by adding forceExit, so I think https://github.com/synonymdev/bitkit/pull/1450/commits/b2c789a573466861f695f11f0569eb0680384940 should be reverted

**Comment by coreyphillips** [comment]
> All e2e tests are now working on my machine! Also I've fixed detox not exiting issue by adding forceExit, so I think [b2c789a](https://github.com/synonymdev/bitkit/commit/b2c789a573466861f695f11f0569eb0680384940) should be reverted

Nice! Reverted [here](https://github.com/synonymdev/bitkit/pull/1450/commits/a135849e6c521c7b68cc8e7190205a47877ec289).

**Comment by limpbrains** [comment]
While using the wallet I often can see 
`TypeError: Cannot read property 'blockchainScripthash_getHistoryBatch' of undefined` looks like a bug

**Comment by limpbrains** [comment]
My wallet only has 2 incoming transactions for 2 first addresses. I've sent some sats to address with index 19. Wallet can't see it. pull-to-refresh doesn't help. If I go to Address Viewer and tap check balances, it works and it finds the output

**Comment by coreyphillips** [comment]
> My wallet only has 2 incoming transactions for 2 first addresses. I've sent some sats to address with index 19. Wallet can't see it. pull-to-refresh doesn't help. If I go to Address Viewer and tap check balances, it works and it finds the output

This is consistent with the master branch of Bitkit, but needs to be addressed/expanded on in the beignet library. Specifically, the gap limit is respected in that it will not let the user create more than 20 empty addresses, but the look-ahead is limited to one in order to reduce the number of queries made to Electrum on startup. However, we should be able to adjust this to suit our needs. For example, we may want to increase the look-ahead to 20 when restoring from a seed and bring it back down to 1 thereafter. We could also bring the look-ahead back to 20 when pulling to refresh, etc. So these params will need to be added in beignet for us to utilize in Bitkit accordingly.

File: src/screens/Wallets/Send/Amount.tsx [hunk 4]
```diff
@@ -96,20 +92,63 @@ const Amount = ({ navigation }: SendScreenProps<'Amount'>): ReactElement => {
...
+			unit,
+			utxos,
+		]),
+	);
+
+	// Set initial text for NumberPadTextField
+	useEffect(() => {
+		const result = getNumberPadText(outputAmount, unit);
+		setText(result);
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
limpbrains, line 129: Can you please add a comment explaining why some dependency is not in the deps. Otherwise it is hard to debug when there is an issue

limpbrains (approved):
Tests are passing, I can't find any major issues. I think we can merge it.
Or maybe @pwltr wants to take another look

__tests__/wallet-restore.ts
```diff
@@ -36,155 +28,129 @@ describe('Wallet - wallet restore and receive', () => {
 			await rpc.generateToAddress(10, address);
 			balance = await rpc.getBalance();
 		}
-
-		waitForElectrum = await initWaitForElectrumToSync(
-			{ host: electrumHost, port: electrumPort },
-			bitcoinURL,
-		);
-	});
```
limpbrains: this logic nees to be restored, otherwise tests will sometimes fail, because electrum is out of sync

coreyphillips: Restored [here](https://github.com/synonymdev/bitkit/pull/1450/commits/75fa00ba5c705d79fc464b5531e7578e644217f0).

src/screens/Wallets/Send/Amount.tsx
coreyphillips: Resolved [here](https://github.com/synonymdev/bitkit/pull/1450/commits/5b64da853fc0fee5752248070a1e30b6024d5107).

pwltr:
I can't switch Electrum server and the UI is behaving buggy still:

https://github.com/synonymdev/bitkit/assets/8538369/72926a44-0ee9-4a18-9c4d-6be0ed4e9469

pwltr:
Fee calculation is still off and sometimes shown incorrectly in the UI, also having issues sending transactions:

https://github.com/synonymdev/bitkit/assets/8538369/dcd40d28-eb91-41f8-ac08-9c84054f5404

Coin selection UI is showing 2 utxos at first and then only one:

https://github.com/synonymdev/bitkit/assets/8538369/fb5a1eda-b022-450e-84e2-0a959c8e5b81

https://github.com/synonymdev/bitkit/assets/8538369/4b1db594-bfcd-4f0f-b30c-5786fd73abe1

Also 'max' button font should be orange

pwltr:
> Coming from the main scanner and "Paste QR Code" the "total required" in coin selection is not updated correctly, seems to be a regression:
> 
> EDIT: it works with max amount but not when I type in a custom amount EDIT2: same when coming from "clipboard ease-of-use" feature

Seems to be still the case

JeanlChristophe:
From a new wallet, using testnet I have received funds with a bech32 address. However, when I change to Native Segwit or legacy and try to send sats, there is $0.00 available.

JeanlChristophe:
@coreyphillips Corey, I don't understand why the issues Phil raised last month still exist. Yet it was merged.

coreyphillips:
> @coreyphillips Corey, I don't understand why the issues Phil raised last month still exist. Yet it was merged.

What issue are you referring to? The custom Electrum server UI issue is the only one I know of as being unresolved, is considered minor and will take some time to resolve once we address the more immediate bugs/issues.

pwltr:
@coreyphillips 

This one is a major issue:

> Coming from the main scanner and "Paste QR Code" the "total required" in coin selection is not updated correctly, seems to be a regression

This one is still there, sending tx seems fine, fees are off in calculation and UI shows something different from calc also:

> Fee calculation is still off and sometimes shown incorrectly in the UI, also having issues sending transactions

and the mentioned Electrum UI bugs:

> ~I can't switch Electrum server and the UI is behaving buggy still~

Comment [here](https://github.com/synonymdev/bitkit/pull/1533) is more up-to-date

PR #1458, opened 2024-01-04
build: regtest app variant

ovitrif:
### Description

This PR adds support to build an app variant having `regtest` as default bitcoin network.

Aside to the default network, the regtest variant has:
- a different name: `Bitkit Regtest`
- a different id: 
  - Android: `to.synonym.bitkit.wallet.regtest`
  - iOS `to.synonym.bitkit.regtest`
- a monochrome icon
  

### Considerations
- It is expected have a `.env.production` file with the corresponding configurations.

---

#### 1️⃣ Regtest Android
1. Run `yarn regtest:set`
2. Run `yarn bundle:regtest`
3. Run `yarn regtest:unset`
4. Install the app on an emulator
5. **Expect** to have the app `Bitkit Regtest` installed with the default Bitcoin Network `Regtest`.

#### 2️⃣ Mainnet Android
1. Run `yarn bundle` then install the app
2. Install the app
2. **Expect** to have the app `Bitkit` installed with the default Bitcoin Network `Mainnet`.

---

#### 3️⃣ Regtest iOS
1. Run `yarn regtest:set`
2. Open the workspace in Xcode, switch to the `bitkit.regtest` scheme
   > **Note** (maybe optional if building via Xcode)
   > Additional 1-time setup is required as described in this [medium post](https://medium.com/@fadilah.arifki29/deployment-ios-in-app-store-react-native-d845e99e8329)
   > - need to register the new bundle id, get certificates, etc
   > - also might need to create a separate app in TestFlight App Store Connect
3. Build the app for release
4. **Expect** to have the app `Bitkit Regtest` installed with the default Bitcoin Network `Regtest`.
5. Run `yarn regtest:unset`

#### 2️⃣ Mainnet iOS
1. Close Xcode, run `yarn clean`, reopen project and switch to `bitkit` scheme
2. Build the app for release
3. **Expect** to have the app `Bitkit` installed with the default Bitcoin Network `Mainnet`.

#### ➕ Other configurations
- Ensure `yarn start` can still run build and run the app on Android / iOS
- Ensure `yarn android` & `yarn ios` still works

---

### Linked Issues/Tasks

https://app.asana.com/0/0/1205821036778041/f

### Type of change

### Tests

- [x] No test

### Screenshot / Video

N/a - build setup.

### QA Notes

N/a - build setup.


pwltr:
Just making sure that normal development workflow works as expected and seems to be the case 👍

package.json
```diff
@@ -24,8 +24,15 @@
     "clean:rn": "watchman watch-del-all && npx del-cli node_modules nodejs-assets/nodejs-project/node_modules && yarn && npx del-cli /tmp/metro-*",
     "clean:android": "cd android && npx del-cli build app/build",
     "clean:ios": "cd ios && npx del-cli Pods Podfile.lock build && pod deintegrate && pod install",
-    "bundle": "yarn clean:android && react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.jsbundle --assets-dest android/app/build/intermediates/res/merged/release/ && cd android && ./gradlew assembleRelease --no-daemon",
+    "bundle": "yarn clean:android && react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.jsbundle --assets-dest android/app/build/intermediates/res/merged/release/ && cd android && ./gradlew assembleMainnetRelease --no-daemon",
+    "bundle:ios": "react-native bundle --reset-cache --platform ios --dev false --entry-file index.js --bundle-output ios/main.jsbundle",
     "prepare": "husky install",
+    "prebundle:regtest": "node ./scripts/pre-regtest-bundle.js",
```
pwltr: These scripts seems to be duplicated with the `:ios` ones and also seem to be unsused? Or undocumented?

scripts/post-regtest-bundle.js
pwltr, line 1: If these are just to backup the `.env` file I wouldn't worry about that, were not doing that either for the other env files

pwltr (dismissed):

scripts/post-regtest-bundle.js
ovitrif: Thanks.
They also create/then-delete a `.production.local.env` which is used to override the config value for the default wallet.

Though the simpler it can get the better IMHO, so I could for now remove the backup part 👍🏻

package.json
ovitrif: Indeed, I went this route of duplicating the pre/post but I ain't a big fan of it either 🙈 .

The other option I see is to have a `pre:regtest` + a `post:regtest` and simply run them before and after the current commands for bundling the regtest version 

That might be cleaner indeed 👍🏻

scripts/post-regtest-bundle.js
pwltr: If that is the goal maybe adding another line in `postinstall.js` does the trick?

```
const createRegtestEnvFile = 'cp .env.production.regtest .env.production.local';
...
exec(createRegtestEnvFile);
```

package.json
pwltr: Any reason for the duplication though? Are these commands picked up in a build step or something? All of these do the same to their non ":ios" counterpart:

"prebundle:regtest:ios", "postbundle:regtest:ios", "bundle:regtest:ios"

scripts/post-regtest-bundle.js
pwltr: This will also be run after `yarn install` so that would remove the need to do it explicitly.

package.json
ovitrif: > Any reason for the duplication though?0

Well there are ^^, they follow the pre/post prefix automatically picked up by node, so basically they remove the need to explicitly run the 2 commands before and after the main script.
I.e. without them, at least of the bundling scripts would need to have:
```
yarn pre… && _theCurrentScript_ && yarn post
```
For example, let's say we remove the ios ones, then the `bundle:regtest:ios` would become:
```sh
yarn prebundle:regtest && react-native bundle --reset-cache --platform ios --dev false --entry-file index.js --bundle-output ios/main.jsbundle && yarn postbundle:regtest
```

Still, as mentioned earlier, not really favoring to keep them 👍🏻

pwltr: Oh I see now, I admit I was not aware you could do custom pre-/post- scripts like that. I think you can see where I was coming from haha. In that case I totally get it 👍

scripts/post-regtest-bundle.js
ovitrif: Coming back to this, the reason I set it up this way was to only **temporarily** set the network to `Regtest`, then revert it back.
It's basically because this is only to be done when prepping an internal release build.

android/app/src/regtest/res/mipmap-anydpi-v26/ic_launcher_round.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<adaptive-icon xmlns:android="http://schemas.android.com/apk/res/android">
+    <background android:drawable="@color/amber_dark" />
```
pwltr:
```suggestion
    <background android:drawable="@color/black" />
```

pwltr (dismissed):

scripts/post-regtest-bundle.js
pwltr: For this did you mean to backup and use `.env.production` (instead of `.env.prodution.local`)? That would make sense to me, but you'd also need to get all the other env values into the bundle if that is not the case yet.

android/app/src/regtest/res/mipmap-anydpi-v26/ic_launcher_round.xml
ovitrif: Thanks for spotting this 👍🏻 , feels that I missed merging something, let me check again and in the process address this as well

scripts/post-regtest-bundle.js
ovitrif: I've went the route to create `.env.production.local` which will partially override env vars declared in the `.env.production`, because IMHO that would most likely be non-existent.*
In this particular case we only care for the wallet default network, so the created `…local` file only declares that. The `react-native-dotenv` dependency we use handles that by merging the values from the two files, giving priority to the `….local` file. This works for different environments as well, in this case `production`. Since there is only one value to "overwrite", all the other configs will have the value from the `.production` file.


\*But if such a file does exist, I didn't want it to be overwritten directly, hence the backup script.

let me know if this is more clear to understand 👍🏻 , I get the feeling I should've been more clear in explaining the mechanism of action involved here

pwltr: Ok that's cool, good solution imo.

android/app/src/regtest/res/mipmap-anydpi-v26/ic_launcher_round.xml
ovitrif: I get it now, seems I forgot to update the `round` one 🤦🏻 
Thanks once again for spotting it 🙇🏻

ovitrif: Resolved (and rebased with master)

pwltr:
Followed instructions and Android is working for me, when building for iOS I get:

`Unable to load contents of file list: '/Users/phil/Dev/synonymdev/bitkit/ios/Pods/Target Support Files/Pods-bitkit/Pods-bitkit-frameworks-Debug.Regtest-input-files.xcfilelist'`

@ovitrif Any ideas? Could have to do with the regtest assets

ovitrif:
> Followed instructions and Android is working for me, when building for iOS I get:
> 
> `Unable to load contents of file list: '/Users/phil/Dev/synonymdev/bitkit/ios/Pods/Target Support Files/Pods-bitkit/Pods-bitkit-frameworks-Debug.Regtest-input-files.xcfilelist'`
> 
> @ovitrif Any ideas? Could have to do with the regtest assets

@pwltr Can you try a `yarn clean:ios` beforehand? 🤔 

Honestly I'm not sure which one of the commands have fixed it in my case, but I do recall having to resolve the same issue.

`clean:ios` seems to tap into the iOS dependencies, but it might have bean `clean` itself.

Xcode build issues feel very unstable to me, for example I couldn't get the app to build via Xcode for 2 days, then randomly it started to work 🤷🏻

pwltr:
> @pwltr Can you try a `yarn clean:ios` beforehand? 🤔

Yeah that works! Should have known 🤷‍♂️. It's stuck on the splash screen now though:

```
No bundle URL present.

Make sure you're running a packager server or have included a .jsbundle file in your application bundle.
*** Terminating app due to uncaught exception 'RCTFatalException: No bundle URL present.

Make sure you're running a packager server or have included a .jsbundle file in your application bundle.', reason: 'No bundle URL present.

Make sure you're running a packager server or have included a .jsbundle file in your application bundle.'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000180491128 __exceptionPreprocess + 172
	1   libobjc.A.dylib                     0x000000018008412c objc_exception_throw + 56
	2   Bitkit Regtest                      0x0000000104b03d10 RCTGetFatalHandler + 0
	3   Bitkit Regtest                      0x0000000104b19ea4 __28-[RCTCxxBridge handleError:]_block_invoke + 516
	4   libdispatch.dylib                   0x0000000107c200f0 _dispatch_call_block_and_release + 24
	5   libdispatch.dylib                   0x0000000107c2193c _dispatch_client_callout + 16
	6   libdispatch.dylib                   0x0000000107c315e4 _dispatch_main_queue_drain + 1228
	7   libdispatch.dylib                   0x0000000107c31108 _dispatch_main_queue_callback_4CF + 40
	8   CoreFoundation                      0x00000001803f1a30 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12
	9   CoreFoundation                      0x00000001803ec148 __CFRunLoopRun + 1936
	10  CoreFoundation                      0x00000001803eb5a4 CFRunLoopRunSpecific + 572
	11  GraphicsServices                    0x000000018e9fbae4 GSEventRunModal + 160
	12  UIKitCore                           0x00000001852f02e4 -[UIApplication _run] + 868
	13  UIKitCore                           0x00000001852f3f5c UIApplicationMain + 124
	14  Bitkit Regtest                      0x000000010494c74c main + 96
	15  dyld                                0x0000000106aa1544 start_sim + 20
	16  ???                                 0x0000000106bb20e0 0x0 + 4407894240
	17  ???                                 0x2962000000000000 0x0 + 2981945903272689664
)
libc++abi: terminating due to uncaught exception of type NSException
``` 

> Xcode build issues feel very unstable to me, for example I couldn't get the app to build via Xcode for 2 days, then randomly it started to work 🤷🏻

I doubt this was the issue in your case but I find myself having to delete `ios/.xcode.env.local` often times.

ovitrif:
> Yeah that works! Should have known 🤷‍♂️. It's stuck on the splash screen now though:
```
No bundle URL present.
```
Thank you @pwltr for testing and sharing this.

I tried again in a fresh clone and I got the same issue when running the `bitkit.regtest` scheme via Xcode.

The thing is, this won't work anyways to set the default network to regtest, because it takes configuration from `.env.development`. But I do agree for testing purposes it makes sense to be able to actually run the `regtest` app, not only install it.

After a lot of trial & error, I found a fix that got it working for me.

---

It's important mention that **the release configuration works differently**, and I can't test it given I don't have the certificates, so I personally can't tell for sure if it will work right.

I'd really appreciate if @Jasonvdb could check if the setup works for iOS `Regtest.Release`, since that's the main problem the PR is trying to solve.

ovitrif:
Note: if we can't get the current setup to work for iOS release builds, this might be an alternative, though it would take some time to set-up fastlane:
- [React Native DevOps Guide](https://medium.com/@tgpski/react-native-devops-guide-2d8e4755ebee)

ovitrif:
Updated with:
- rebase on `master` d62e7a2d
- squash into 2 commits:
  1. Android build
  2. iOS build

ovitrif:
Put on hold until we can build iOS releases using the self-hosted build machine. This way it will enable me to troubleshoot and verify it on iOS as well

ovitrif:
@Jasonvdb I revisited this changeset and made the setup much simpler, also now I'm pretty sure the iOS regtest app should build correctly.

I also updated the testing steps in PR description as well as build instructions for regtest app in `BUILD.md`.

Happy to get your feedback when you can check this again 🙏🏻

ovitrif:
Closing this for now, given it's stale for a long time.
We can always reopen if we reassess its priority.

PR #1459, opened 2024-01-05
feat(lightning): using on chain address for channel close

Jasonvdb:
### Description

- Updates react-native-ldk to include a [change](https://github.com/synonymdev/react-native-ldk/pull/196) where funds from a channel close are sent direct to Bitkit's on chain wallet and no longer require and extra sweeping step.
- Also includes a [fix](https://github.com/synonymdev/react-native-ldk/pull/198) preventing some users from completing the restore flow

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA Notes

- Funds should show up directly on chain after channel close
- Some users stuck on restoring screen should not be able to progress

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -392,7 +392,7 @@ PODS:
     - React-Core
   - react-native-keep-awake (1.2.2):
     - React-Core
-  - react-native-ldk (0.0.124):
+  - react-native-ldk (0.0.125):
     - React
   - react-native-mmkv (2.10.2):
     - MMKV (>= 1.2.13)
@@ -905,7 +905,7 @@ SPEC CHECKSUMS:
   react-native-flipper: 9c1957af24b76493ba74f46d000a5c1d485e7731
   react-native-image-picker: 2e2e82aba9b6a91a7c78f7d9afde341a2659c7b8
   react-native-keep-awake: ad1d67f617756b139536977a0bf06b27cec0714a
-  react-native-ldk: 5ac636bea5e24c687a4a009a339d5ee82e8c8d0f
+  react-native-ldk: c78b18c8c8fe218481721f7083e3e0f825aa957e
   react-native-mmkv: 9ae7ca3977e8ef48dbf7f066974eb844c20b5fd7
   react-native-netinfo: 5ddbf20865bcffab6b43d0e4e1fd8b3896beb898
   react-native-quick-base64: a5dbe4528f1453e662fcf7351029500b8b63e7bb
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@synonymdev/blocktank-client": "0.0.50",
     "@synonymdev/blocktank-lsp-http-client": "0.9.0",
     "@synonymdev/feeds": "2.1.1",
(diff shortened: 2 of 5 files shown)


pwltr (dismissed):
Can you fix the failing onboarding e2e test?

pwltr:
Actually although this works I think we'd want more work on the UI to avoid confusing the user. Right now it shows a pending card (60m) but at the same time also shows an activity item (10-20m) not marked as transfer.

image

pwltr (approved):
Will approve because UI was not in the scope

coreyphillips:
Instead of using `generateNewReceiveAddress` we'll want to grab what the current `addressIndex` is instead. Since `generateNewReceiveAddress` will always create a new address regardless of whether or not the previous index was used. This means opening and closing the app a few times will end up pushing the user's `addressIndex` into the gap limit. I'll make a change suggestion for this shortly.

Jasonvdb:
> Can you fix the failing onboarding e2e test?

I was seeing `webRelayClient not ready yet` in all the PR's? Unless you mean another test?

Jasonvdb:
> Instead of using `generateNewReceiveAddress` we'll want to grab what the current `addressIndex` is instead. Since `generateNewReceiveAddress` will always create a new address regardless of whether or not the previous index was used. This means opening and closing the app a few times will end up pushing the user's `addressIndex` into the gap limit. I'll make a change suggestion for this shortly.

Oh yes right. So originally I had it use `getReceiveAddress` and I changed it to return `IAddress` instead of a string but I thought maybe I was breaking something with the store and didn't want to go down that path right before testing session.

pwltr:
> > Can you fix the failing onboarding e2e test?
> 
> I was seeing `webRelayClient not ready yet` in all the PR's? Unless you mean another test?

Was talking about this:

```
FAIL e2e/onboarding.e2e.js (109.532 s)
  Onboarding
    ✕ Can pass onboarding correctly (101505 ms)

  ● Onboarding › Can pass onboarding correctly

    expect(received).toBe(expected) // Object.is equality

    Expected: "bitcoin:bcrt1q2zkt93dtl8cxqgtvzxurlu0mu6z6ahnh6sy5uf"
    Received: "bitcoin:bcrt1qjz82ap0k2x6r7zg9jketlthx8whm0dne8f0cvx"

      120 | 		const { label: address2 } = await element(by.id('QRCode')).getAttributes();
      121 |
    > 122 | 		jestExpect(address1).toBe(address2);
          | 		                     ^
      123 |
      124 | 		markComplete('onboarding-1');
      125 | 	});
```

Will be fixed by what @coreyphillips described above 👍

coreyphillips:
> Oh yes right. So originally I had it use `getReceiveAddress` and I changed it to return `IAddress` instead of a string but I thought maybe I was breaking something with the store and didn't want to go down that path right before testing session.

PR created [here](https://github.com/synonymdev/bitkit/pull/1465). Since most of this will be changed once the beignet migration is merged anyway I simply created a new method that is effectively the same as `getReceiveAddress` except it returns the entire object. Will make it easier to merge the changes in later on.

coreyphillips (approved):

Issue #1467, opened 2024-01-05
[Bug]: Headlines widget isn't synchronized


JeanlChristophe:
### Describe the bug

Headline from the widget isn't synchronised.

### Reproduce

1. Go to 'add widget'
2. Click on 'bitcoin headlines'
3. Close and reopen the app
4. See error, it's still the same headline

### Operating system

Graphene OS, Iphone, Android

### Bitkit version

v93

flaviomoceri:
If we want to show to the user a different headlines every time, this PR implements this
https://github.com/synonymdev/bitkit/pull/1474

PR #1474, opened 2024-01-05
feat(widgets): Show a different news each time on Headlines widget

flaviomoceri:
### Description

Currently if the user uses the "Headlines" widget the latest article is shown because the `getLatestArticle()` function is used.

In this PR I use the `getAllArticles()` function to show a random article to the user each time.

### Linked Issues/Tasks

https://github.com/synonymdev/bitkit/issues/1467

### Type of change

- [X] Refactoring (improving code without creating new functionality)

### Tests

- [ ] Detox test
- [ ] Unit test
- [X] No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/63161412/e7922010-0b54-4009-92af-1bde43e92513

### DIFF
--- a/src/components/HeadlinesWidget.tsx
+++ b/src/components/HeadlinesWidget.tsx
@@ -66,7 +66,10 @@ const HeadlinesWidget = ({
 					`${url}?relay=${webRelayUrl}`,
 				);
 
-				const _article = await reader.getLatestArtible();
+				const allArticles = await reader.getAllArticles();
+				const _article = allArticles
+					.sort((a, b) => b.published - a.published)
+					.slice(0, 10)[Math.floor(Math.random() * 10)];
 
 				if (!unmounted && _article) {
 					setArticle(_article);


src/components/HeadlinesWidget.tsx
```diff
@@ -66,7 +66,11 @@ const HeadlinesWidget = ({
 					`${url}?relay=${webRelayUrl}`,
 				);
 
-				const _article = await reader.getLatestArtible();
+				const allArticles = await reader.getAllArticles();
+				const randomIndexArticle = Math.floor(
+					Math.random() * allArticles.length,
+				);
+				const _article = allArticles[randomIndexArticle];
```
pwltr: Currently `getAllArticles` returns a lot of articles many of them are very old. So I would suggest filtering for a `publishedDate` that is not too far back so the widget always shows a current article.

See https://webrelay.slashtags.to/tdm8gwcuh6nfw4w8remadpzid9cff4tpoyazeokzbrfnjrxyf9ho/Bitcoin%20Headlines/feed/headlines

pwltr (changes requested):

src/components/HeadlinesWidget.tsx
flaviomoceri: Yes, it makes sense. What do you think is the right number, the latest 10-20…50 most recent articles?

pwltr: Please ask @aldertnl  regarding requirements for this

pwltr (approved):

Issue #1491, opened 2024-01-12
[Bug]: Swipe balance to hide still works in wallet details screen, even if this feature is disabled


limpbrains:
### Describe the bug

Swipe balance to hide still works in wallet details screen, even if this feature is disabled

### Reproduce

1. Disable Swipe balance to hide
2. open wallet details screen
3. swipe the balance

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/155891/0cfddbca-a9e8-45b3-9d98-51b29602d05b

### Operating system

iOS

### Bitkit version

v94

PR #1493, opened 2024-01-12
fix(wallet): Updated SwipeBalance option also in Wallet details screen

flaviomoceri:
### Description

Updated SwipeBalance option also in Wallet details screen
Now when you disable the option from settings you are not able to swipeBalance even from the Wallet details screen

### Linked Issues/Tasks

https://github.com/synonymdev/bitkit/issues/1491

### Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

### Tests

- [ ] Detox test
- [ ] Unit test
- [X] No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/63161412/99f1df02-dfc4-4c13-b3d5-5091a845da75

### DIFF
--- a/src/screens/Wallets/WalletsDetail/index.tsx
+++ b/src/screens/Wallets/WalletsDetail/index.tsx
@@ -44,7 +44,10 @@ import SafeAreaInset from '../../../components/SafeAreaInset';
 import Money from '../../../components/Money';
 import BlurView from '../../../components/BlurView';
 import { updateSettings } from '../../../store/slices/settings';
-import { hideBalanceSelector } from '../../../store/reselect/settings';
+import {
+	hideBalanceSelector,
+	enableSwipeToHideBalanceSelector,
+} from '../../../store/reselect/settings';
 import { capitalize } from '../../../utils/helpers';
 import DetectSwipe from '../../../components/DetectSwipe';
 import type { WalletScreenProps } from '../../../navigation/types';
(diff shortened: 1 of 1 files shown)


src/screens/Wallets/WalletsDetail/index.tsx
```diff
@@ -213,6 +219,7 @@ const WalletsDetail = ({
 									exiting={FadeOut}>
 									<View color="transparent" style={styles.balanceContainer}>
 										<DetectSwipe
+											enabled={enableSwipeToHideBalance}
```
pwltr, line 222: Not sure the user should be able to 'swipe to hide' on this screen (might want to double check with Aldert) but otherwise lgtm

pwltr (approved):

src/screens/Wallets/WalletsDetail/index.tsx
flaviomoceri: Currently on the `WalletDetails` page if I swipe I can hide the various amounts, but if I disabled this function from the settings, in home I am no longer able to do it, but in `WalletDetails` I can still do it, this PR fixes this

I checked the latest Aldert design though and it seems that in `WalletDetails` the user shouldn't be able to do any swiping in general, so I guess this functionality needs to be eliminated? I'll ask Aldert

pwltr: I agree it looks like the design doesn't intend for this functionality to be there. Also because the eye icon is not there (like on the home screen) this could lead to confusion around this feature again which we are currently trying to reduce.

flaviomoceri: Yes, I agree, let's wait for what Aldert says, if so I will apply the changes to remove the functionality from this screen

flaviomoceri: Aldert has said "Sure, I don't see why not", so I didn't make any changes, I just added the eye icon to avoid confusion


https://github.com/synonymdev/bitkit/assets/63161412/dd3856d8-32f1-4cfa-8853-48b1e3890e39

pwltr (approved):

PR #1498, opened 2024-01-15
feat(wallet): Show info toast when hiding balance

ovitrif:
### Description

Added a toast message to explain the hidden wallet balance state for user clarity and resolution.

Works on both:
- wallets overview (home) and
- asset (Bitcoin) overview.

### Linked Issues/Tasks

Design v45 (see slack message or Figma file)

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/4588074/c480a1dc-6cc8-4d60-9bee-b8eab383e4c0

### QA Notes

See above video.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206354352883295

### DIFF
--- a/src/store/slices/user.ts
+++ b/src/store/slices/user.ts
@@ -14,6 +14,7 @@ export type TUser = {
 	requiresRemoteRestore: boolean;
 	startCoopCloseTimestamp: number;
 	ignoresHideBalanceToast: boolean;
+	ignoresSwitchUnitToast: boolean;
 };
 
 export const initialUserState: TUser = {
@@ -28,6 +29,7 @@ export const initialUserState: TUser = {
 	requiresRemoteRestore: false,
 	startCoopCloseTimestamp: 0,
 	ignoresHideBalanceToast: false,
+	ignoresSwitchUnitToast: false,
 };
 
 export const userSlice = createSlice({
@@ -66,6 +68,9 @@ export const userSlice = createSlice({
 		ignoreHideBalanceToast: (state) => {
 			state.ignoresHideBalanceToast = true;
 		},
+		ignoreSwitchUnitToast: (state) => {
+			state.ignoresSwitchUnitToast = true;
+		},
 		resetUserState: () => initialUserState,
 	},
 });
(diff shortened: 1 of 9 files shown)


ovitrif:
Checked with Aldert and confirmed this is needed also in the wallet detail view.

pwltr (approved):

ovitrif:
⚠️ Need to make the toast show only once (the first time the user performs the action),
as suggested by Aldert (thanks @flaviomoceri)

coreyphillips (approved):

ovitrif:
> ⚠️ Need to make the toast show only once (the first time the user performs the action), as suggested by Aldert

Updated the PR with logic to show the toast only once.

src/store/slices/user.ts
```diff
@@ -26,6 +27,7 @@ export const initialUserState: TUser = {
 	lightningSettingUpStep: 0,
 	requiresRemoteRestore: false,
 	startCoopCloseTimestamp: 0,
+	ignoresHideBalanceToast: false,
```
pwltr, line 31: Looks good, but whenever you're adding a field to the redux store you have to add a migration in `store/migrations.ts` to add the new field to persisted store otherwise people updating the app will not have this field initially which will throw an error. Also update the version for `persistConfig` in `store/index.ts`

pwltr (changes requested):

src/store/slices/user.ts
ovitrif: Thanks a lot for this remark, I've added the suggested changes for the store migration
Please have another look when possible 🙏🏻

pwltr (approved):

coreyphillips (approved):

ovitrif:
The PIN e2e test keeps failing repeatedly but I saw it failing on other PRs as well, don't really think the problem comes from the changes.

I will retrigger the e2e tests again.

limpbrains:
And now tests will fail because blocktank staging is down 🫠

ovitrif:
Blocktank staging still down, hence the tests kept failing

PR #1532, opened 2024-02-01
Update env Templates

coreyphillips:
### Description

- Adds empty string to `TREASURE_HUNT_HOST` in env templates to prevent "`TREASURE_HUNT_HOST` is not defined" errors when git cloning and installing.
- Not sure if there's a better way to handle this?

### Type of change

### Tests
- [x] No test

### Screenshot / Video
Simulator Screenshot - iPhone 14 - 2024-02-01 at 14 53 25

### QA Notes
- We should now be able to run the following command without an error resembling the attached screenshot.
```
git clone git@github.com:synonymdev/bitkit.git && cd bitkit && git checkout update-env-templates && yarn install && cd ios && pod install && cd ../ && yarn ios
```


pwltr (approved):
`react-native-dotenv` has an option to `allowUndefined` https://github.com/synonymdev/bitkit/blob/08c7a8c781deb7e70898c826c1458fdd77da9112/babel.config.js#L12-L17

Not sure if that would be the better solution because we actually want the app to throw the above error for most env vars. I think I prefer the empty string.

PR #1537, opened 2024-02-05
feat: run e2e tests on self-hosted

limpbrains:
### Description

- add self-host runner
- fix webrelay test

New runner is much faster than the old one

### Type of change

Refactoring

### Tests

Detox test


limpbrains:
For now I'm adding it as separate check, will see how it works and if it fine, will remove old one text week

pwltr (approved):

PR #1571, opened 2024-02-20
feat(wallet): Add Gap Limit Options

coreyphillips:
### Description
- Upgrades Beignet to `0.0.23`.
- Implements `gapLimitOptions` from upgrade.
    - New wallets default to a `lookBehind` of `5` and a `lookAhead` of `1` to limit queries to Electrum.
    - Restored wallets temporarily use a `lookBehind` of `20` and a `lookAhead` of `20` during the initial scan. Once this scan finishes it reverts back to the default lookBehind-lookAhead of 5-1.
    - If you think the 5-1 defaults are low, we can try something higher like 5-5, but restored wallets may come close to the limits of what a given Electrum server will tolerate for extremely active users or heavy testing sessions.
- Adds two dev-only views.
    - Monitored Address Types: Found in "Settings->Advanced->Bitcoin Address Type" beneath "Monitored Address Types" allowing us to adjust what address types are actively being monitored. An app restart is required for changes to take effect.
    - Gap Limit: Found in "Settings->Advanced->Gap Limit" allowing us to adjust the gap limit in Bitkit as needed for testing/debugging. 
- Updates `broadcastTransaction`, `getReceiveAddress` & `getBalance` methods.

### Type of change

### Tests
- [x] No test

### Screenshot / Video
Simulator Screenshot - iPhone 14 - 2024-02-19 at 20 54 27
Simulator Screenshot - iPhone 14 - 2024-02-19 at 20 54 10

### QA Notes
- Restores should continue to work as expected. Once a restore is successful and the initial scan is complete we should be able to observe that the gap limit lookAhead and lookBehind were reverted to their default of 5-1 in "Settings->Advanced->Gap Limit".

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@synonymdev/web-relay": "1.0.7",
     "backpack-client": "github:synonymdev/bitkit-backup-client#f08fdb28529d8a3f8bfecc789443c43b966a7581",
     "bech32": "2.0.0",
-    "beignet": "0.0.22",
+    "beignet": "0.0.23",
     "bip21": "2.0.3",
     "bip32": "4.0.0",
     "bip39": "3.1.0",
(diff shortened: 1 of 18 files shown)


limpbrains:
How does lookBehind: 20 works during wallet restore, it still can't go behind index 0, right?

limpbrains (approved):

coreyphillips:
> How does lookBehind: 20 works during wallet restore, it still can't go behind index 0, right?

A lookBehind of 20 is not necessarily needed during the initial scan for a restore since we're scanning all addresses as we generate them anyway, it's just there to conform to the standard/default.

PR #1586, opened 2024-02-22
feat(wallet): Detect If Exceeding Electrum Batch Limits

coreyphillips:
### Description
- Bitkit now displays a toast whenever it receives an error when refreshing or rescanning to display to the user.
- Bitkit will also look for batch limit exception warnings when scanning or refreshing and inform the user accordingly.
- Adds `handlerRefreshError` to `refreshBeignet` & `rescanAddresses` methods.
- Upgrades beignet to `0.0.24`.

### Type of change

### Tests
- [x] No test

### QA Notes
I believe Reza bumped the limits so hopefully we'll never see the throttle toast, but it can be triggered by continuously refreshing and rescanning Bitkit using a wallet with an extremely high gap limit and several spread out UTXO's. Beware, spreading out UTXO's, shrinking the gap limit and then rescanning from 0 may make it appear as though funds were lost and not related to being throttled.

### DIFF
--- a/src/store/migrations/index.ts
+++ b/src/store/migrations/index.ts
@@ -519,6 +519,15 @@ const migrations = {
 			},
 		};
 	},
+	39: (state): PersistedState => {
+		return {
+			...state,
+			ui: {
+				...state.ui,
+				isElectrumThrottled: false,
+			},
+		};
+	},
 };
 
 export default migrations;
--- a/src/store/migrations/index.ts
+++ b/src/store/migrations/index.ts
@@ -519,15 +519,6 @@ const migrations = {
 			},
 		};
 	},
-	39: (state): PersistedState => {
-		return {
-			...state,
-			ui: {
-				...state.ui,
-				isElectrumThrottled: false,
-			},
-		};
-	},
 };
 
 export default migrations;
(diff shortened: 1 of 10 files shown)


src/store/migrations/index.ts
```diff
@@ -519,6 +519,15 @@ const migrations = {
 			},
 		};
 	},
+	39: (state): PersistedState => {
+		return {
+			...state,
+			ui: {
+				...state.ui,
+				isElectrumThrottled: false,
+			},
```
pwltr: The migration is not needed in this case since `ui` state is not persisted

pwltr (changes requested):

src/store/migrations/index.ts
coreyphillips: Reverted [here](https://github.com/synonymdev/bitkit/pull/1586/commits/479f61b4bcdb9727eb08ef9598f69c1d29d42fb0).

pwltr (approved):

JeanlChristophe:
@coreyphillips please don't forget to add to each ticket:

- Project
-  Status
- Sprint
- Milestone
Thanks

PR #1611, opened 2024-03-18
chore(wallet): Upgrade react-native-keychain

coreyphillips:
### Description
- Upgrades to the `refactor/android-data-store` branch of `react-native-keychain`.

### Type of change
- [x] Dependency Upgrade

### Tests
- [x] No test

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "react-native-haptic-feedback": "2.2.0",
     "react-native-image-picker": "7.0.2",
     "react-native-keyboard-accessory": "0.1.16",
-    "react-native-keychain": "8.1.2",
+    "react-native-keychain": "git+ssh://git@github.com:synonymdev/react-native-keychain.git#refactor/android-data-store",
     "react-native-localize": "3.0.2",
     "react-native-mmkv": "2.11.0",
     "react-native-modal": "13.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -102,7 +102,7 @@
     "react-native-haptic-feedback": "2.2.0",
     "react-native-image-picker": "7.0.2",
     "react-native-keyboard-accessory": "0.1.16",
-    "react-native-keychain": "8.1.2",
+    "react-native-keychain": "git+ssh://git@github.com:synonymdev/react-native-keychain.git#refactor/android-data-store",
```
pwltr: May want to switch to http for pulling the lib, so tests can run. I see the repo is publicly accessible

```suggestion
    "react-native-keychain": "github:synonymdev/react-native-keychain.git#refactor/android-data-store",
```

pwltr (changes requested):

package.json
coreyphillips: Ah that's right, nice catch. Updated [here](https://github.com/synonymdev/bitkit/pull/1611/commits/c8c4b4740f534b42cbc968382467c862ec78ec89).

ovitrif (approved):
Thank you @coreyphillips for handling this!

Looks good on my end, and still works same as it did when using local dependency pointing to these changes, ie. the data gets migrated on the back, and user-facing everything keeps working, nothing manually needed from the user:

| Android Studio 2024-03-19 001639 |
| - |

pwltr (approved):

PR #1612, opened 2024-03-19
fix: channel UI fixes

pwltr:
### Description

- TotalBalance: don't include channel punishment reserve in total balance
- QuickSetup: fix slider position & disabled button immediately after channel close
- CustomSetup: fix not able to update channel duration
- fix(transfer): fix transfer pending time not updated on first block confirmation
- fix(lightning): avoid flashing different price after channel purchase confirmation
- fix(receive): fix minimum amount for addtitional CJIT
- fix(transfer): only show transfer to savings if tx has output to our wallet

### Linked Issues/Tasks

https://app.asana.com/0/0/1204208728352470/1206703239419962/f

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test


limpbrains (approved):
why are you removing DEFAULT_CHANNEL_DURATION everywhere ?

pwltr:
> why are you removing DEFAULT_CHANNEL_DURATION everywhere ?

Just a small cleanup, I just set it as a default for all the functions that use it. So we don't have to pass it explicitly if it's the same as default, can still pass and override it if needed

PR #1643, opened 2024-04-12
feat: ledger

limpbrains:
### Description

Introducing the Bitkit Ledger Library, designed to meticulously record all transactions in a dedicated database and validate the current wallet balance to ensure accuracy and consistency.

#### Internal Structure

The Bitkit Ledger comprises four wallets:
- **onchain**
- **lightning**
- **onchain_remote**
- **lightning_remote**

Wallets without the `_remote` suffix represent app balances, while those with the suffix correspond to external (outside) wallets.

Each wallet includes two accounts:
- **available**: funds that are available in the wallet and ready to be spent.
- **hold**: funds that are unconfirmed.

A fundamental principle is that money cannot be created from thin air; each transaction always has a source (FROM) and a destination (TO). For instance, if an onchain transaction is received from another wallet, it is recorded as: onchain_remote / available -> onchain / available. After this transaction, onchain_remote / available will show a negative balance, which is expected and helps us verify that the summary of all wallet accounts balances to zero.

#### Example

Here is a simple example demonstrating the flow of receiving onchain funds and then sending them back:

1. Incoming unconfirmed transaction:
   `onchain_remote / available -> onchain / hold`

2. Incoming transaction confirmed:
   `onchain / hold -> onchain / available`

3. Outgoing unconfirmed transaction:
   `onchain / available -> onchain / hold`

4. Outgoing confirmed transaction:
   `onchain / hold -> onchain_remote / available`

#### Screenshots

Column "Actual" represents values retrieved from the LDK/beignet libraries. If these match the "Expected" values, then everything is functioning as intended.

Tapping on a transaction displays detailed information:

Each transaction also records the account balances FROM and TO before execution, aiding in troubleshooting.

#### Emojis
To streamline the transaction history, emojis are used to represent wallets and accounts:

- **onchain**: 🏠🔗
- **lightning**: 🏠⚡
- **onchain_remote**: 🌐🔗
- **lightning_remote**: 🌐⚡

Accounts:
- **available**: 💰
- **hold**: 🔒

#### Implementation Details

The ledger uses BigInt for all calculations, storing values in millisatoshis to align with Lightning Network (LN) specifications. However, because the LDK provides values in satoshis, some rounding discrepancies are anticipated.

The ledger is a standalone library, available at [Bitkit Ledger GitHub repository](https://github.com/synonymdev/ledger).

### Linked Issues/Tasks

[Asana Task](https://app.asana.com/0/1189635884316057/1205833268269694/f)

### Type of Change

New Feature

### Tests

Unit Tests

### QA Notes

1. Conduct several transactions.
2. Open development settings and select 'Ledger'.
3. Ensure that the transaction list and balances match the expected outcomes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "immer": "10.0.3",
     "intl-messageformat": "10.5.4",
     "jdenticon": "3.2.0",
+    "@synonymdev/ledger": "0.0.3",
     "lodash": "4.17.21",
     "lottie-react-native": "5.1.6",
     "mime": "3.0.0",
--- a/src/utils/startup/index.ts
+++ b/src/utils/startup/index.ts
@@ -175,7 +175,7 @@ export const startWalletServices = async ({
 		// Refresh slashpay config
 		updateSlashPayConfig2({ selectedNetwork, forceUpdate: true });
 
-		await syncLedger();
+		syncLedger();
 
 		return ok('Wallet started');
 	} catch (e) {
--- a/src/utils/startup/index.ts
+++ b/src/utils/startup/index.ts
@@ -25,6 +25,7 @@ import { promiseTimeout } from '../helpers';
 import { EAvailableNetwork } from '../networks';
 import { TWalletName } from '../../store/types/wallet';
 import { runChecks } from '../wallet/checks';
+import { setupLedger, syncLedger } from '../ledger';
 
 /**
  * Creates a new wallet from scratch
@@ -109,6 +110,8 @@ export const startWalletServices = async ({
 			refreshBlocktankInfo().then();
 		});
 
+		setupLedger({ selectedWallet, selectedNetwork });
+
 		const mnemonicResponse = await getMnemonicPhrase();
 		if (mnemonicResponse.isErr()) {
 			return err(mnemonicResponse.error.message);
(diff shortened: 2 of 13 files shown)


src/utils/startup/index.ts
```diff
@@ -172,6 +175,8 @@ export const startWalletServices = async ({
 		// Refresh slashpay config
 		updateSlashPayConfig2({ selectedNetwork, forceUpdate: true });
 
+		await syncLedger();
```
coreyphillips: Do we need to await `syncLedger` here? Just want to avoid prolonging startup if we can.

limpbrains: we don't

coreyphillips:
```suggestion
		syncLedger();
```

coreyphillips (commented):
Just one suggested change, but otherwise LGTM!

package.json
```diff
@@ -81,6 +81,7 @@
     "immer": "10.0.3",
     "intl-messageformat": "10.5.4",
     "jdenticon": "3.2.0",
+    "@synonymdev/ledger": "0.0.3",
```
pwltr, line 84: Can we publish the code for this?

limpbrains: Done

coreyphillips (approved):

Issue #1685, opened 2024-04-17
[Bug]: Instant Payments Setup Failed required amount error inconsistency


catch-21:
### Describe the bug

When setting up a LN channel on the Instant Payments Spending Balance screen, the amount required to overcome the following message is inconsistent and sometimes shows when selecting a particular value and other times allows proceeding to next screen.

"Instant Payments Setup Failed"
"Failed to create order. "Channel size too small. Minimum size is 133100sat."

### Reproduce

1. Go to Spending Balance page (tap Pay instantly suggestion on home page -> Quick Setup)
2. Try to invoke the"Channel size too small" error (this can be tricky). Try moving the slider around 20-50%, and tapping "Continue".
3. If it proceeded to "Please Confirm" page then go back and try again with a different approach, e.g. via "Enter amount" page, toggling unit amount and retrying. It can be produced with perseverance but I'm unsure of the exact conditions.
4. Once error has been produced it tends to persist even if spending amount is changed. Try moving the spending amount a bit higher and tap "Continue" to attempt to address the problem.
5. Eventually, with a higher value it will proceed to "Please Confirm" page. Usually this requires a higher-than-expected allocation to spending balance.
6. Go back to "Spending Balance" page
7. Lower the spending balance down to a value that was previously causing the error and press "Continue"
8. There's a good chance it will proceed to "Spending Balance" page even though it previously showed the error and prevented this.

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/74595920/ed7ec0ce-90c0-452f-ab43-dbb63d97904b

### Operating system

Android 13 TKQ1.220829.002

### Bitkit version

v1.0.0-beta.113

limpbrains:
This error comes from BT, @SeverinAlexB can you please take a look?

catch-21:
Tested on 898a58a42979ef8bee906b10182c8a12399e1d16
Unable to reproduce using a similar wallet. Closing.

PR #1727, opened 2024-04-24
fix: add 10% to the BT info minChannelSizeSat field

limpbrains:
### Description

BT updates `minChannelSizeSat` from info every ~20 seconds. So if the user will stay longer than that on Setup screen he can receive an error when. I'm adding 2 things:
- re-fetch /info every 20 seconds while user is on Setup screen
- add 10% reserve to minChannelSizeSat when do calculations.

### Linked Issues/Tasks

#1685

### Type of change

Bug fix

### Tests

No test

### Screenshot / Video

### QA Notes

- open setup screen
- watch for BT /info (https://api.stag.blocktank.to/blocktank/api/v2/info or https://api.blocktank.to/api/v2/info
- when minChannelSizeSat spikes, try to create a channel

### DIFF
--- a/src/screens/Lightning/QuickSetup.tsx
+++ b/src/screens/Lightning/QuickSetup.tsx
@@ -66,9 +66,19 @@ const QuickSetup = ({
 			const setupTransfer = async (): Promise<void> => {
 				await resetSendTransaction();
 				await setupOnChainTransaction();
-				refreshBlocktankInfo().then();
 			};
 			setupTransfer();
+		}, []),
+	);
+
+	// refresh BT info every 20 seconds, because minChannelSizeSat changes
+	useFocusEffect(
+		useCallback(() => {
+			refreshBlocktankInfo().then();
+			const interval = setInterval(() => {
+				refreshBlocktankInfo().then();
+			}, 20000);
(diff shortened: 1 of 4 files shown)


src/screens/Lightning/QuickSetup.tsx
```diff
@@ -66,12 +66,22 @@ const QuickSetup = ({
...
+	// refresh BT info every 20 seconds, because minChannelSizeSat changes
+	useFocusEffect(
+		useCallback(() => {
+			refreshBlocktankInfo().then();
+			const interval = setInterval(() => {
+				refreshBlocktankInfo().then();
+			}, 5000);
+			return (): void => clearInterval(interval);
+		}, []),
+	);
```
coreyphillips: Discrepancy between comment of 20 seconds and interval of 5 seconds.

coreyphillips (changes requested):

src/screens/Lightning/QuickSetup.tsx
limpbrains: Fixed

coreyphillips (approved):

Issue #1619, opened 2024-03-27
[Bug]: Fees indication


JeanlChristophe:
### Describe the bug

I'm trying to make an onchain transaction of 10,000 sats, and it says that the fees will be $0.15, but when I continue, the correct fees are charged.

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/46056211/95ce5e7b-2ce0-44e2-bc92-1392b6ca92ec

### Operating system

GrapheneOS

### Bitkit version

v111

coreyphillips:
To reproduce:

- Create a transaction for 5000 sats.
- Tap "Continue". In a high fee environment the fee should get set to "Custom" at a lower sat per vByte.
- Navigate Back.
- Tap "Continue" again without adjusting the sat amount.
- Observe that "Custom" is now "Normal", but at the same fee rate.

catch-21:
I currently see $0.11, which appears to be the value for 1sat/vB. For some reason it displays the custom value which defaults to 1sat/vB.

catch-21:
Tested on 658b9f2bd22d2ca63893a89dc56f47a054546d79
I'm no longer able to reproduce the original issue 👍 
I am seeing what JC described in their last comment, 'Normal' when it should be 'Custom' 👎  see vid:

https://github.com/synonymdev/bitkit/assets/74595920/a12ac82e-5b23-4551-b253-b92309cdbfe2

Current fees are Slow=5 Normal=18 Fast=21

@coreyphillips Do you consider this related here or a separate UI bug?

coreyphillips:
This would be a new issue.

catch-21:
Raised https://github.com/synonymdev/bitkit/issues/1896

PR #1732, opened 2024-04-25
Fix Max Fee Checks

coreyphillips:
### Description
- Adds max fee checks to the `FeeRate` & `FeeCustom` views to prevent users from selecting or setting a fee-rate above what the max is.
- Adds `getFeeInfo` method.
- Resolves UI bug in #1619

### Linked Issues/Tasks
- #1619

### Type of change

### Tests
- [x] No test

### QA Notes
When setting fees during the send flow, the max fee checks should be in place and prevent you from setting a value above the `maxFee` provided by Beignet's `getFeeInfo` method.

### DIFF
--- a/src/screens/Wallets/Send/FeeCustom.tsx
+++ b/src/screens/Wallets/Send/FeeCustom.tsx
@@ -55,12 +55,12 @@ const FeeCustom = ({
 
 	const onPress = (key: string): void => {
 		const current = feeRate.toString();
-		const newAmount = handleNumberPadPress(key, current, { maxLength: 4 });
+		const newAmount = handleNumberPadPress(key, current, { maxLength: 3 });
 		if (Number(newAmount) > maxFee) {
 			showToast({
 				type: 'info',
-				title: 'Max possible fee rate',
-				description: `${maxFee} sats/vbyte`,
+				title: t('max_possible_fee_rate'),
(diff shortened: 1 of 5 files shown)


src/screens/Wallets/Send/FeeCustom.tsx
```diff
@@ -43,7 +55,15 @@ const FeeCustom = ({
 
 	const onPress = (key: string): void => {
 		const current = feeRate.toString();
-		const newAmount = handleNumberPadPress(key, current, { maxLength: 3 });
+		const newAmount = handleNumberPadPress(key, current, { maxLength: 4 });
+		if (Number(newAmount) > maxFee) {
+			showToast({
+				type: 'info',
+				title: 'Max possible fee rate',
+				description: `${maxFee} sats/vbyte`,
```
limpbrains: These strings should be added to i18n vocabulary

limpbrains (changes requested):

src/screens/Wallets/Send/FeeCustom.tsx
coreyphillips: Updated [here](https://github.com/synonymdev/bitkit/pull/1732/commits/a5b326c9673a28feacc8ad6db39b87831f5da67e).

limpbrains (approved):

catch-21:
Verified in v118

PR #1734, opened 2024-04-25
fix: use BT state2 field, change pending orders logic

limpbrains:
### Description

- use order.state2 and order.payment.state2 fields
- do not watch for EXECUTED or EXPIRED BT orders

### Type of change

Bug fix

### Tests

No test


pwltr (changes requested):
`refreshOrder` is still using old `state` in some places

pwltr (approved):

PR #1748, opened 2024-04-26
chore: remove ST v1

limpbrains:
### Description

- remove v1 -> v2 migration code
- remove all v1 libs 
- remove ldk old restore
- remove ldk recover screen
- remove AuthWidget
- remove old env variables
- add error message if user tries to scan slashauth url

I will squash commits after review

### Type of change

Refactoring

### Tests

No test

### QA Notes

Install prev version, add slashauth, upgrade, try to use slashauth. Widget should disapear


Nuhvi:
@limpbrains I wonder if we should just merge this PR, test the performance of the app, to confirm that the DHT is causing most of the issues, then consider our options to add slashauth again in the light of that expirement.

limpbrains:
Good idea, I will test it tomorrow

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
 	"extends": "@react-native/typescript-config/tsconfig.json",
 	"compilerOptions": {
+		"jsx": "react",
```
pwltr: This option is set by `@react-native/typescript-config/tsconfig.json` with `"jsx": "react-native"

pwltr (changes requested):

tsconfig.json
limpbrains: Removed

src/utils/i18n/locales/en/slashtags.json
```diff
@@ -88,15 +79,15 @@
...
 	"error_saving_contact": "Unable To Save Contact",
 	"error_saving_profile": "Unable To Save Profile",
 	"error_delete_contact": "Unable To Delete Contact",
 	"error_sdk_title": "Please Restart Bitkit",
 	"error_sdk_msg": "An essential component is not working properly. Please restart Bitkit.",
 	"error_pay_title": "Unable To Pay Contact",
-	"error_pay_empty_msg": "There is no payment data available for this contact."
+	"error_pay_empty_msg": "There is no payment data available for this contact.",
+	"auth_depricated_title": "Deprecated",
+	"auth_depricated_msg": "Slashauth is deprecated. Please Bitkit Beta."
```
pwltr:
```suggestion
	"auth_depricated_msg": "Slashauth is deprecated. Please use Bitkit Beta."
```

pwltr (changes requested):

pwltr (approved):
Tested and LGTM

PR #1753, opened 2024-04-29
fix: various UI fixes

### LINKED ISSUE #1709: "Swipe balance to hide" doesn't need to hide "Hide balance on open"
In the Security and Privacy settings, if the "Swipe balance to hide" option is disabled then the "Hide balance on open" option is hidden. Although, if "Hide balance on open" is enabled at the time it is not disabled. This means that the next time the app is loaded the balance is hidden but it is not possible to swipe to reveal it. Fortunately, there is the open eye icon that can be tapped to reveal the balance. Therefore, it is ok to leave the "Hide balance on open" option in the settings for the user to decide how they want it to behave.

There are two ways to address this:
1. Stay with hiding "Hide balance on open" but ensure it is also disabled, because otherwise it could be either enabled or disabled without the user knowing by observing the settings.
2. Don't hide "Hide balance on open" when "Swipe balance to hide" is disabled.

Here's a video to help reinforce my point:

https://github.com/synonymdev/bitkit/assets/74595920/6d3cb66a-9949-408a-971a-3f2c77b19c01

pwltr:
### Description

- fix(settings): #1709 swipe to hide balance
- fix(transfer): fix not navigating to success
- fix(ui): adjust onboarding screen image size

### Linked Issues/Tasks

Closes #1709 

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test


src/components/OnboardingScreen.tsx
```diff
@@ -37,16 +39,18 @@ const OnboardingScreen = ({
 				onClosePress={onClosePress}
 			/>
 			<View style={styles.content} testID={testID}>
-				<View style={styles.imageContainer}>
+				<View
+					style={[
+						styles.imageContainer,
+						{ marginBottom: imagePosition === 'center' ? 'auto' : 48 },
```
limpbrains, line 46: new eslint warning

limpbrains (approved):

PR #1773, opened 2024-05-02
fix: do not run refreshWallet until the wallet is initialized

limpbrains:
### Description

Before initializing `wallet` in the `setupOnChainWallet` function, `wallet.refreshWallet` is called from the `getWalletData` callback. To avoid this, I am introducing a function to ensure that `wallet` is defined before running `refreshWallet`. Similar situations may arise, and we could consider changing the variable type to `TWallet | undefined` in the future. What are your thoughts, @coreyphillips?

### Type of change

Bug fix

### Tests

No test

### QA Notes

Make sure that error `TypeError: Cannot read property 'refreshWallet' of undefined` doesn't appear on app start

### DIFF
--- a/src/utils/wallet/index.ts
+++ b/src/utils/wallet/index.ts
@@ -131,6 +131,20 @@ export const setupAddressGenerator = async ({
 	}
 };
 
+/*
+ * Wait for wallet to be ready
+ */
+const waitForWallet = async (): Promise<void> => {
+	if (wallet) {
+		return;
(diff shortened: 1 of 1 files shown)


coreyphillips (approved):

coreyphillips:
Agreed, setting it as `TWallet | undefined` in the future and applying a wait if found to be undefined whenever calling it may be a good enhancement.

PR #1791, opened 2024-05-06
fix: prevent user from going back during PIN set flow

### LINKED ISSUE #1789: [Bug]: Retype PIN screen shows 'this is not the same PIN' when navigaing back from successfully reentering PIN
### Describe the bug

When setting up your PIN, if the retype is entered incorrectly and then connectly it still displays 'this is not the same PIN' when navigating back. It should ~~clear this error on successful retype~~ prohibit navigating back from the biometrics option screen because at this stage the pin has already been enabled.

### Reproduce

1. Go to enable PIN Code
2. Type a PIN
3. Retype the PIN incorrectly
4. Retype the PIN correctly
5. Navigate back from the Biometric option screen
6. Observe 'Try again, this is no the same PIN' error persists

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/74595920/3c1960e0-9244-4532-98fa-c70326a5a881

### Operating system

Android 13 TKQ1.220829.002

### Bitkit version

2c5b07726fd35cb796a9566c47d0f6c36fda4ef4

### Log output

_No response_

limpbrains:
### Description

By removing screens from navigation we prevent user from going back during PIN set flow

### Linked Issues/Tasks

closes #1789

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/screens/Settings/PIN/ChoosePIN.tsx
+++ b/src/screens/Settings/PIN/ChoosePIN.tsx
@@ -65,7 +65,11 @@ const ChoosePIN = ({
 			if (pinsAreEqual) {
 				addPin(pin);
 				dispatch(hideTodo(pinTodo.id));
-				navigation.navigate('AskForBiometrics');
+				// replace the navigation stack to avoid going back to the PIN screen
+				navigation.reset({
+					index: 0,
+					routes: [{ name: 'AskForBiometrics' }],
+				});
 			} else {
 				vibrate({ type: 'notificationWarning' });
 				setPin('');
--- a/src/screens/Settings/PIN/AskForBiometrics.tsx
+++ b/src/screens/Settings/PIN/AskForBiometrics.tsx
@@ -65,14 +65,21 @@ const AskForBiometrics = ({
 
 	const onSkip = (): void => {
 		const bioType = biometryData?.biometryType ?? 'Biometrics';
-		navigation.navigate('Result', { bio: false, type: bioType });
+		// use navigation.reset to prevent user from going back to previous screen
+		navigation.reset({
+			index: 0,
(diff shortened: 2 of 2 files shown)


src/screens/Settings/PIN/ChoosePIN.tsx
```diff
@@ -65,7 +65,11 @@ const ChoosePIN = ({
 			if (pinsAreEqual) {
 				addPin(pin);
 				dispatch(hideTodo(pinTodo.id));
-				navigation.navigate('AskForBiometrics');
+				// replace the navigation stack to avoid going back to the PIN screen
+				navigation.reset({
+					index: 0,
+					routes: [{ name: 'AskForBiometrics' }],
+				});
```
pwltr, line 72: What's the advantage of not allowing the user to navigate back here vs. just hiding the text with `setTryAgain(false);`?

pwltr (approved):

src/screens/Settings/PIN/ChoosePIN.tsx
limpbrains: there is no easy way to do it. I've tried https://reactnavigation.org/docs/preventing-going-back/ but it is not supportted fro native stack navigator.
So you basically always have to:
- disable android back button
- disable swipe to prevent go back on iOS

this method is just simpler. and works for both platforms

pwltr: I mean what is the problem with letting the user navigate back? It seems to work fine with changing the PIN etc.

limpbrains: Feels wrong. User already has a PIN, why would we allow it to be changed

PR #1793, opened 2024-05-06
fix: activity sorting condition

### LINKED ISSUE #1728: [Bug]: Transaction Activities can restore in wrong order
### Describe the bug

After a wallet restore, the order of the transactions can be different to the order they were originally listed. The expected order is roughly the same they were broadcasted to the network. Whether they were confirmed later then others shouldn't be significant in most cases.

The order can fix itself if further transactions occur. I'm unsure whether this is coincidence or the result of an explicit ordering mechanism. See screenshot 3 for example of it fixing itself.

### Reproduce

TODO: Define reliable reproduction steps

### Screenshots / Recording

1. After wallet restore, showing the Sent transaction occurring before the initial Receive transaction 

2. Same as above but in the Ledger view

3. Showing transaction activities after making a couple more transactions. The order appears correct again now.

4. Same as above but the order hasn't fixed itself in the Ledger view

### Operating system

Android 13 TKQ1.220829.002

### Bitkit version

c55c045f58b7e76ee82acdc1419bccf9db8490ae

### Log output

_No response_

limpbrains:
### Description

Fix condition to sort activity list 

### Linked Issues/Tasks

closes #1728

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/utils/activity/index.ts
+++ b/src/utils/activity/index.ts
@@ -93,6 +93,13 @@ export const mergeActivityItems = (
 			needsSorting = true;
 			break;
 		}
+		if (
+			mergedItems[i].timestamp === mergedItems[i - 1].timestamp &&
+			mergedItems[i].txType === 'sent'
+		) {
+			needsSorting = true;
+			break;
+		}
 	}
 
 	if (!needsSorting) {


src/utils/activity/index.ts
```diff
@@ -93,6 +93,13 @@ export const mergeActivityItems = (
 			needsSorting = true;
 			break;
 		}
+		if (
+			mergedItems[i].timestamp === mergedItems[i - 1].timestamp &&
```
coreyphillips, line 97:
```suggestion
		if (
		        i > 0 &&
			mergedItems[i].timestamp === mergedItems[i - 1].timestamp &&
```

May need to add `i > 0` to the beginning of the condition to prevent an out of bounds error.

coreyphillips (changes requested):

src/utils/activity/index.ts
limpbrains: `i` starts with `1`, so everything is safe here

coreyphillips (approved):

PR #1801, opened 2024-05-08
fix: use blocktank host env var to setup client

Jasonvdb:
### Description

- Updates Blocktank client to use env var in preparation for using the new server

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### DIFF
--- a/src/utils/blocktank/index.ts
+++ b/src/utils/blocktank/index.ts
@@ -19,10 +19,7 @@ import {
 import { sleep } from '../helpers';
 import { DEFAULT_CHANNEL_DURATION } from '../../utils/wallet/constants';
 import { dispatch, getBlocktankStore, getUserStore } from '../../store/helpers';
-import {
-	ICreateOrderRequest,
-	TGeoBlockResponse,
-} from '../../store/types/blocktank';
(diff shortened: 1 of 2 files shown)


src/utils/blocktank/index.ts
```diff
@@ -350,12 +347,22 @@ export const isGeoBlocked = async (fromStorage = false): Promise<boolean> => {
...
-		const response = await fetch(
-			`${__BLOCKTANK_HOST__}/api/v2/channel/geocheck`,
-		);
-		const data: TGeoBlockResponse = await response.json();
-		return !!data?.error;
-	} catch {
+
+		const response = await fetch(`${__BLOCKTANK_HOST__}/geocheck`);
+
+		if (response.status == 403) {
```
limpbrains: eslint errors

limpbrains (changes requested):

src/utils/blocktank/index.ts
Jasonvdb: thank you, switched from swift to JS 😅

limpbrains (approved):

PR #1822, opened 2024-05-14
fix: backport Black screen fix to beta branch

limpbrains:
### Description

Backporting #1820

### Type of change

Bug fix

### Tests

No test


ovitrif:
If I understand correctly this also adds the fix from:
- #1773
Which very nice IMHO, just wanted to have it mentioned for treaceability

Jasonvdb (approved):

PR #1827, opened 2024-05-14
fix: add switch unit logic to CJIT screen

### LINKED ISSUE #1790: [Bug]: Tapping on value at Receive Bitcoin screen for spending doesn't toggle currency
### Describe the bug

On the 'Receive Bitcoin' screen, that's shown when requesting spending balance, tapping on the bitcoin/fiat value does not toggle between the two (need to use the dedicated button). On other screens such as the main wallet view, 'Transfer Funds' and  'Edit Invoice' for onchain, tapping the value does toggle.

### Reproduce

1. Go to input an amount to receive for spending balance
2. Tap on the value
3. Value does not toggle between bitcoin and fiat

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/74595920/9c8c7a77-6aef-4a45-bbc8-4e6a04c7ec3e

### Operating system

Android 13 TKQ1.220829.002

### Bitkit version

2c5b07726fd35cb796a9566c47d0f6c36fda4ef4

### Log output

_No response_

limpbrains:
### Description

Add unit switching for CJIT screen

### Linked Issues/Tasks

closes #1790

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/screens/Wallets/Receive/ReceiveAmount.tsx
+++ b/src/screens/Wallets/Receive/ReceiveAmount.tsx
@@ -32,6 +32,7 @@ import {
 	unitSelector,
 } from '../../../store/reselect/settings';
 import type { ReceiveScreenProps } from '../../../navigation/types';
+import { useSwitchUnit } from '../../../hooks/wallet';
 
 const ReceiveAmount = ({
 	navigation,
@@ -43,6 +44,7 @@ const ReceiveAmount = ({
 	const denomination = useAppSelector(denominationSelector);
 	const invoice = useAppSelector(receiveSelector);
 	const blocktank = useAppSelector(blocktankInfoSelector);
+	const switchUnit = useSwitchUnit();
 	const [minimumAmount, setMinimumAmount] = useState(0);
 
 	const { maxChannelSizeSat } = blocktank.options;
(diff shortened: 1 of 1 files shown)


src/screens/Wallets/Receive/ReceiveAmount.tsx
```diff
@@ -76,12 +78,17 @@ const ReceiveAmount = ({
 	const onChangeUnit = (): void => {
 		const result = getNumberPadText(invoice.amount, denomination, nextUnit);
 		dispatch(updateInvoice({ numberPadText: result }));
+		switchUnit();
```
pwltr, line 81: The unit button component already calls this function

pwltr (changes requested):

src/screens/Wallets/Receive/ReceiveAmount.tsx
limpbrains: I see this problem in other places, like QuickSetup, ReceiveDetails
Maybe we should address it later ?

pwltr: true let's do that

pwltr (approved):

PR #1829, opened 2024-05-14
fix(ui): Replace showBottomSheet with toggleBottomSheet

coreyphillips:
### Description
- Replaces `showBottomSheet` with `toggleBottomSheet` to prevent stuck send/receive bottom-sheets on failed action updates.
- Previously, quickly tapping and closing either the Send or Receive buttons/bottom-sheets would prevent the close action from firing in the bottom-sheet component. This results in a stuck position where the app thinks the Send/Receive view is open, but it's actually closed. So tapping on Send or Receive will not do anything. This PR is a temp fix that re-enables the Send/Receive buttons after two presses instead of remaining in the stuck position, forcing the user to restart in order to re-enable them.
- Ideally in the future, we need to find a way to respond to interrupted open/close animations from the bottom-sheet wrapper.

### Type of change

### Tests
- [x] No test

### QA Notes
- Quickly tap and close either the Send or Receive view. Worst case scenario, you should now be able to re-open the Send/Receive view after no more than two taps instead of needing to close and re-open the app.

### DIFF
--- a/src/components/TabBar.tsx
+++ b/src/components/TabBar.tsx
@@ -12,7 +12,7 @@ import { useTranslation } from 'react-i18next';
 import { FadeIn } from 'react-native-reanimated';
 
 import { receiveIcon, sendIcon } from '../assets/icons/tabs';
-import { showBottomSheet } from '../store/utils/ui';
+import { toggleBottomSheet } from '../store/utils/ui';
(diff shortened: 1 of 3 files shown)


limpbrains (approved):

catch-21:
Verified in v118

PR #1836, opened 2024-05-16
fix: Don't show rocket screen twice during restore

### LINKED ISSUE #1634: [Bug]: Restoring your seed is very slow
### Describe the bug

This error has occurred several times with John, me, Aldert, etc. When I tried to restore my seed, I had a black screen for more than 10 seconds and then the rocket arrived.
Normally I should not have a black screen.

### Screenshots / Recording

_No response_

### Operating system

Graphene OS and iOS

### Bitkit version

v113

### Log output

_No response_

ovitrif:
Fixes #1634 

### Description

Removes code that was displaying the loading screen (aka. rocket screen) from within the `RestoreFromSeed` screen (which is part of onboarding flow).

With this change only `App.tsx` remains to show the rocket screen, which is presented by the `RestoringScreen` subcomponent it loads, while we're waiting on a result from the long-running restore operations.

#### Rationale

> It's a small fix, it's not perfect (I mean we still see a black screen for a very short time after tapping `Restore`, before the rocket screen shows, but the rocket screen is pretty neat now, shows only once, the progress anims on it do feel like making more sense and all in all it should smooth the experience of users while restoring.

### Type of change

### Tests

- [x] No test

### Screenshot / Video

<video src="https://github.com/synonymdev/bitkit/assets/4588074/1a3fc3fc-4080-4384-a7ce-641f9cb39e39" style="max-width: 315px;"></video>

#### To Test

Just restore a wallet, it should be more user-friendly now that we don't show the rocket-screen twice, each time starting from 0%.

#### Context

Check out this comment on what this PR fixes:
- https://github.com/synonymdev/bitkit/issues/1634#issuecomment-2088361555

Compare the video of the BEFORE with the AFTER video/ your experience while validating this 👍🏻

### DIFF
--- a/src/screens/Onboarding/RestoreFromSeed.tsx
+++ b/src/screens/Onboarding/RestoreFromSeed.tsx
@@ -33,7 +33,6 @@ import { useAppDispatch } from '../../hooks/redux';
 import { validateMnemonic } from '../../utils/wallet';
 import { restoreSeed } from '../../utils/startup';
 import { showToast } from '../../utils/notifications';
-import LoadingWalletScreen from './Loading';
 import NavigationHeader from '../../components/NavigationHeader';
 import { updateUser, verifyBackup } from '../../store/slices/user';
 
@@ -51,10 +50,11 @@ const RestoreFromSeed = (): ReactElement => {
 	const { t } = useTranslation('onboarding');
 	const enableButtons = useMemo(
 		() =>
+			!isRestoringWallet &&
 			!seed.includes(undefined) &&
 			!validWords.includes(false) &&
 			validateMnemonic(seed.join(' ')),
-		[seed, validWords],
+		[isRestoringWallet, seed, validWords],
 	);
 	const showRedExplanation = useMemo(
 		() => seed.some((word, index) => word !== undefined && !validWords[index]),
@@ -169,14 +169,6 @@ const RestoreFromSeed = (): ReactElement => {
 			/>
 		);
 	};
-
-	if (isRestoringWallet) {
-		return (
-			<ThemedView style={styles.root}>
-				<LoadingWalletScreen />
-			</ThemedView>
-		);
-	}
 
 	return (
 		<ThemedView style={styles.root}>


src/screens/Onboarding/RestoreFromSeed.tsx
```diff
@@ -171,11 +170,15 @@ const RestoreFromSeed = (): ReactElement => {
 	};
 
 	if (isRestoringWallet) {
-		return (
-			<ThemedView style={styles.root}>
-				<LoadingWalletScreen />
-			</ThemedView>
-		);
+		/**
```
limpbrains: Please remove all unused code
Also need to return something here
`return <></>` would do or full screen black `<ThemedView style={styles.root} />`
Otherwise disable submit button

limpbrains (changes requested):

catch-21:
Tested this on a few android devices (Poco F3, Galaxy A12, Galaxy Note 10+), emulated Pixel 8a and iOS simulator. Much improved. Still some black screen time on older devices such as A12 but newer devices it is reduced. No double rocket observed on any device.

src/screens/Onboarding/RestoreFromSeed.tsx
ovitrif: Thanks!

Didn't want to return because afaiu that means I'm overriding what's already shown, nor use a black screen overlay, since that would just extend the time we see a black screen.

The suggestion to disable the buttons does click with me 👍🏻, thus…
Amended and force-pushed the suggested changes to the same commit.

Updated recordings:

| latest | v1 |
| -  | - |
| <video src="https://github.com/synonymdev/bitkit/assets/4588074/1a3fc3fc-4080-4384-a7ce-641f9cb39e39" style="max-width: 315px;"></video> | <video src="https://github.com/synonymdev/bitkit/assets/4588074/654185a6-330c-4bb9-8eb8-65f98cd1bf9a" style="max-width: 315px;"></video> |

PS.
>Please remove all unused code

💯

limpbrains (approved):

PR #1845, opened 2024-05-17
fix(wallet): Set Fast Fee Rate For CPFP When Able

### LINKED ISSUE #1777: [Bug]: Suggested fee to boost is well bellow market rate
### Describe the bug

When boosting a transaction the suggested fee rate isn't suitable to produce an effective fee rate for current market conditions (not current volatile). Once boosted, a transaction cannot be boosted again so it gets stuck in the mempool and the user has no way to take action in Bitkit.

**Update: Due to the error seen when trying to increase the boost tx fee being "Unable to increase the fee any further. Otherwise, it will exceed half the current sending/output amount." this could be a manifestation of https://github.com/synonymdev/bitkit/issues/1751.**

### Reproduce

See recording section

### Screenshots / Recording

This is a long video so here's the breakdown:
1. Construct and submit an onchain transaction with 10 inputs (I don't know for sure if the tx size causes/exacerbates the issue but this is the condition where I first witnessed it) using a fee rate lower than market rate to ensure the transaction doesn't immediately confirm
2. Check the mempool sees the transaction with 6sat/vB fee rate
3. Boost the transaction using the suggested fee
4. Observe that the effective fee rate is still far too low
5. No option to boost again so now stuck.

https://github.com/synonymdev/bitkit/assets/74595920/0b8499cc-59a6-43c0-a48b-f745b855f9d5

### Operating system

Android 13 TKQ1.220829.002

### Bitkit version

20c194faaf1dc812c806fd7b47730f35a4c94286

### Log output

_No response_

coreyphillips:
### Description
- Updates `setupCpfp` to set the fastest fee-rate when able.
- Fixes `BoostPrompt` `Title` to always display the correct sats per vbyte value.

### Linked Issues/Tasks
- Closes #1777

### Type of change

### Tests
- [x] No test

### QA Notes
- Send some sats to the wallet.
- Tap on the Activity entry for the, unconfirmed, received funds and tap "Boost".
- The default sats per vbyte should either be the "fast" fee preference or the highest satsPerByte value the wallet can achieve depending on the user's balance.

### DIFF
--- a/src/screens/Wallets/BoostPrompt.tsx
+++ b/src/screens/Wallets/BoostPrompt.tsx
@@ -38,6 +38,7 @@ import {
 	transactionSelector,
 } from '../../store/reselect/wallet';
 import { validateTransaction } from 'beignet';
+import { EUnit } from '../../store/types/wallet.ts';
 
 const BoostForm = ({
 	activityItem,
@@ -198,7 +199,12 @@ const BoostForm = ({
 
 	const Title = (
 		<View style={styles.adjustValueRow}>
-			<Money sats={transaction.satsPerByte} size="bodyMSB" symbol={true} />
+			<Money
+				sats={transaction.satsPerByte}
+				unit={EUnit.BTC}
+				size="bodyMSB"
+				symbol={true}
+			/>
 			<BodyMSB> {t('sat_vbyte_compact')}</BodyMSB>
 		</View>
 	);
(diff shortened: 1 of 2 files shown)


limpbrains (approved):

catch-21:
Verified in v118. Boosts suggests to use fast fee rate.

Issue #1800, opened 2024-05-07
[Bug]: Continue button can be pressed at invalid position on the Transfer Funds slider page


catch-21:
### Describe the bug

When moving the slider position, it is possible to tap Continue to continue to the next page with invalid values. It doesn't change the action, just looks wrong and could be confusing.
To resolve, the Continue button could either be disabled until slider gets into valid position or apply the value that the slider is snapping to rather than what it is currently at.

### Reproduce

1. With some spending balance and inbound capacity at its limit (create a LN channel with $999 inbound capacity)
2. Go to Transfer Funds page with the slider
3. Move the slider to an invalid location and quickly tap Continue button before it has time to position itself to the closest valid location
4. The 'Please Confirm' page shows the correct description text (for me it is transferring all funds back to savings) but the percentage values and spending balance amount are wrong and misleading.

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/74595920/158d4317-2040-4f95-a919-6df2541fd827

### Operating system

Android 13 TKQ1.220829.002

### Bitkit version

2c5b07726fd35cb796a9566c47d0f6c36fda4ef4

limpbrains:
fixed

catch-21:
Verified in 63a4dd6db6064916773e1b7983f5192a95b78899
LGTM. Nice.

PR #1887, opened 2024-05-27
fix: ignore Continue on QuickSetup screen when Slider is active

limpbrains:
### Description

ignore Continue on QuckSetup screen when Slider is active

### Linked Issues/Tasks

#1800

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/screens/Lightning/QuickSetup.tsx
+++ b/src/screens/Lightning/QuickSetup.tsx
@@ -58,6 +58,7 @@ const QuickSetup = ({
 	const [loading, setLoading] = useState(false);
 	const [showNumberPad, setShowNumberPad] = useState(false);
 	const [textFieldValue, setTextFieldValue] = useState('');
+	const [sliderActive, setSliderActive] = useState(false);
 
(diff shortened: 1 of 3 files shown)


src/screens/Lightning/QuickSetup.tsx
```diff
@@ -146,40 +151,44 @@ const QuickSetup = ({
 	const onContinue = useCallback(async (): Promise<void> => {
 		const { clientBalance, lspBalance, isTransferringToSavings } = lnSetup;
 
+		if (sliderActive) {
+			return;
+		}
```
pwltr, line 156: disable the continue button instead?

pwltr (changes requested):

src/screens/Lightning/QuickSetup.tsx
limpbrains: This causes the button to flicker every time you touch the slider, which is annoying. 
This issue really only happens if you tap the button very quickly ( < 200ms).

pwltr (approved):

PR #1901, opened 2024-05-28
chore(wallet): Update env templates

coreyphillips:
### Description
- Updates `BLOCKTANK_HOST` to `https://api1.blocktank.to/api` in  `.env` templates.

### Type of change

### Tests
- [x] No test

### QA Notes
- Opening channels with Blocktank and returning exchange rates from `getExchangeRates` should continue working as expected.

### DIFF
--- a/.env.test.template
+++ b/.env.test.template
@@ -21,7 +21,7 @@ BACKUPS_SERVER_PUBKEY=0319c4ff23820afec0c79ce3a42031d7fef1dff78b7bdd69b5560684f3
 WEB_RELAY=https://webrelay.slashtags.to
 
 # Blocktank
-BLOCKTANK_HOST=https://api.stag.blocktank.to
+BLOCKTANK_HOST=https://api1.blocktank.to/api
 
 # Network
 ELECTRUM_BITCOIN_HOST=35.187.18.233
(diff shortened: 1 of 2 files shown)


.env.test.template
```diff
@@ -21,7 +21,7 @@ BACKUPS_SERVER_PUBKEY=0319c4ff23820afec0c79ce3a42031d7fef1dff78b7bdd69b5560684f3
 WEB_RELAY=https://webrelay.slashtags.to
 
 # Blocktank
-BLOCKTANK_HOST=https://api.stag.blocktank.to
+BLOCKTANK_HOST=https://api1.blocktank.to/api
```
pwltr, line 24: I think `.env.test` should stay pointed to staging since tests use it. @limpbrains ?

limpbrains: I think it would be better if they do

coreyphillips: The primary reason for the change was due to the geocheck endpoint no longer existing for [this fetch](https://github.com/synonymdev/bitkit/blob/4192ea75b5a729d258335ed63ad0a71ec4be0132/src/utils/blocktank/index.ts#L351):
- https://api.stag.blocktank.to/geocheck

I can revert this change and only change that fetch request url individually, but then it would start to defeat the purpose of the `BLOCKTANK_HOST` `env` variable.

limpbrains: we only use .env.test.template  during tests, so this is fine. 
Adding `https://api1.blocktank.to/api` here would mean that we spam it with requests.

pwltr: Tests seem to pass so LGTM

limpbrains (approved):

PR #1918, opened 2024-05-30
fix(transfer): prevent NaN percentage

pwltr:
### Description

Fixes a small issue where the `QuickConfirm` screen shows `NaN` as percentage when the user has 0 onchain balance.

Simulator Screenshot - iPhone 15 - 2024-05-29 at 23 42 46


limpbrains (approved):

catch-21:
Verified on v124 when attempting to move spending balance back to savings and I have no savings.

PR #1961, opened 2024-06-06
fix: sync language files

limpbrains:
### Description

sync language files

### Type of change

Refactoring

### Tests

No test


src/utils/i18n/locales/de/lightning.json
pwltr, line 1: these german translations are either nonsensical or very awkward, looks like bad google translate/ChatGPT. I wouldn't merge these and leave english for the missing text, it looks better.

limpbrains: I've removed German for now

pwltr: thanks

pwltr (approved):

PR #1977, opened 2024-06-11
fix: currency conversions with classic denomination

limpbrains:
### Description

- fix convertToSats when convering from fiat with classic denomination
- fix getNumberPadText when convering from fiat with classic denomination
- fix getDisplayValues to show correct bitcoinFormatted when using classic denomination
- add tests for these cases

### Type of change

Bug fix

### Tests

Unit test

### Screenshot / Video

https://github.com/synonymdev/bitkit/assets/155891/36567260-34b2-479b-8e4e-6e389e75200e

### QA Notes

- switch to classic denomination
- test all places where we are using custom number keyboard

### DIFF
--- a/__tests__/numberpad.ts
+++ b/__tests__/numberpad.ts
@@ -1,6 +1,32 @@
-// Fix 'getDispatch is not a function'
 import '../src/store/utils/ui';
-import { handleNumberPadPress } from '../src/utils/numberpad';
+import {
+	resetExchangeRates,
+	updateExchangeRates,
+} from '../src/store/actions/wallet';
+import { EDenomination, EUnit } from '../src/store/types/wallet';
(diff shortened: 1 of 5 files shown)


__tests__/numberpad.ts
```diff
@@ -75,3 +101,65 @@ describe('Receive/Send NumberPad', () => {
...
 });
+
+describe('getNumberPadText', () => {
+	describe('can work without exchange rates', () => {
+		beforeAll(() => resetExchangeRates());
+
+		it('returns the correct fiat fallback', () => {
+			const r = getNumberPadText(100000, EDenomination.classic, EUnit.BTC);
+			expect(r).toEqual('0.001');
+		});
```
pwltr: Description says 'returns the correct fiat fallback' but conversion is sats to BTC?

pwltr (approved):

PR #1984, opened 2024-06-12
v1.0.0 release

Jasonvdb:


limpbrains:
I think we can try disable TriggerRenderError not all dev settings

Issue #1986, opened 2024-06-13
Small V1.0 quirks


JeanlChristophe:
- [x] Standard set of widgets is not installed on fresh empty wallet (see Figma for guidance).
- [ ] Wallet creation animation did not finish, it forwarded at 65% or something. I think we recently decided to let the animation finish, even if the phone / wallet creation is faster.
- [x] ‘Invite - Share Bitkit’ suggestion card triggers outdated share message and incorrect appstore link (and card disappears after one time use). _Download Bitkit, Your Ultimate Bitcoin Toolkit. Handing you the keys to reshape your digital life._ https://apps.apple.com/us/app/bitkit/id6443905609 Change your wallet, change the world. Download Bitkit for iPhone https://apps.apple.com/app/bitkit-wallet/id6502440655 or Android https://play.google.com/store/apps/details

PR #1988, opened 2024-06-13
fix: change share text

limpbrains:
### Description

- update share app text
- add i18n

### Linked Issues/Tasks

#1986

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/components/Suggestions.tsx
+++ b/src/components/Suggestions.tsx
@@ -23,7 +23,7 @@ import {
 import { pinSelector } from '../store/reselect/settings';
 import { useAppDispatch, useAppSelector } from '../hooks/redux';
 import type { RootNavigationProp } from '../navigation/types';
-import { appName, shareText } from '../constants/app';
+import { appName, appStoreUrl, playStoreUrl } from '../constants/app';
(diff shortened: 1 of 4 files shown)


src/components/Suggestions.tsx
```diff
@@ -60,9 +60,12 @@ const Suggestions = (): ReactElement => {
 	useFocusEffect(useCallback(() => setIndex(0), []));
 
 	const onShare = useCallback(async (): Promise<void> => {
-		await Share.share({ title: appName, message: shareText });
+		await Share.share({
+			title: appName,
+			message: t('about.shareText', { playStoreUrl, appStoreUrl }),
+		});
 		dispatch(hideTodo('invite'));
```
pwltr: Looks like card should not be hidden after share

pwltr (changes requested):

src/components/Suggestions.tsx
limpbrains: yeah, forgot this

limpbrains: fixed

pwltr (approved):

PR #1992, opened 2024-06-14
chore: configure multiple APKs for ABIs

pwltr:
### Description

Adds Android split APKs 

https://developer.android.com/build/configure-apk-splits#configure-abi-split


ovitrif:
Think I can review + approve this as well, doing it ✅

ovitrif (approved):

PR #1997, opened 2024-06-18
fix: rescan all addresses pull-to-refresh with on main screen

limpbrains:
### Description

Run `refreshWallet({ scanAllAddresses: true });` when user does pull-to-refresh on main screen 

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "@synonymdev/slashtags-widget-price-feed": "1.1.0",
     "@synonymdev/web-relay": "1.0.7",
     "bech32": "2.0.0",
-    "beignet": "0.0.41",
+    "beignet": "0.0.42",
     "bip21": "2.0.3",
     "bip32": "4.0.0",
     "bip39": "3.1.0",
--- a/src/screens/Settings/GapLimit/index.tsx
+++ b/src/screens/Settings/GapLimit/index.tsx
@@ -1,18 +1,18 @@
 import React, { memo, ReactElement, useMemo, useState } from 'react';
-import { StyleSheet } from 'react-native';
 import { useTranslation } from 'react-i18next';
+import { StyleSheet } from 'react-native';
 
-import { View, TextInput, ScrollView } from '../../../styles/components';
-import { BodyM, Caption13Up } from '../../../styles/text';
-import { useAppSelector } from '../../../hooks/redux';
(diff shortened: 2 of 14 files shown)


coreyphillips:
I would not advise setting `scanAllAddresses` to `true` on pull-to-refresh. This is what "Rescan Addresses" is for in settings.

When set to false as it currently is, Bitkit will scan the currently set address gap limit for all previously used address types. This is sufficient for pull-to-refresh and cannot be as easily or unknowingly abused by the user from constant pull-to-refreshes. If we were to set this to true it could potentially scan hundreds/thousands of addresses per refresh initiated by the user and eventually lead to them getting rate-limited by the server even though we increased the server limits recently.

limpbrains:
How about we only do this once an hour?

coreyphillips:
We need to be cognizant of users that aren't using our Electrum server where this will timeout/return an error on them. If we want to perform this in Bitkit we should place it behind a condition that reads something like this: If the Electrum server is Synonym's, perform a full rescan. Otherwise, refresh like normal.

Ultimately, forcing a full rescan for a well used wallet containing hundreds/thousands of addresses is not a scalable solution for a mobile hot wallet. It's overkill and will hinder the performance of the app as they use it more. Making refresh and load times increase dramatically with usage. If we wanted to implement something scalable here we could bump the gap limit to something like 50, perform the refresh at the higher limit, and then drop it back to what it was set at previously. At least this way we know there is a hard testable limit. And if the user genuinely needs to rescan every address they still have that option in Settings. Even still, we should only do this for wallets connected to our Electrum servers and not assume that their custom server can handle the load.

coreyphillips:
If we wanted the app to be a bit smarter about it, we could only initiate a full-rescan if we think the user needs or is looking for it. For example, we only initiate a full-rescan if the user pulls to refresh for the second time within a minute. Or something like that.

limpbrains:
I believe with this patch merged
https://github.com/synonymdev/beignet/pull/67
It is now safe to merge this one as well. Of course it is a temprorary solution and we need to continue to improve beignet/bitkit scannig logic

limpbrains:
Bumped beignet to 0.0.42 adding lookBehindChange and lookAheadChange gap limit options.

pwltr (approved):

PR #2022, opened 2024-06-28
fix(settings): update channel details

pwltr:
### Description

- Add order information for CJIT channels
- Add fee information
- Add order expiry date for pending channels
- Add closure reason

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

Simulator Screenshot - iPhone 15 - 2024-06-28 at 11 28 26
Simulator Screenshot - iPhone 15 - 2024-06-28 at 11 30 01
Simulator Screenshot - iPhone 15 - 2024-06-28 at 11 30 46


src/screens/Settings/Lightning/ChannelDetails.tsx
```diff
@@ -424,64 +465,41 @@ const ChannelDetails = ({
...
-							/>
-						)}
+					{channel.counterparty_node_id && (
 						<Section
-							name={t('is_usable')}
-							value={<CaptionB>{t(channel.is_usable ? 'yes' : 'no')}</CaptionB>}
-						/>
-						<Section
-							name={t('is_ready')}
-							testID={channel.is_channel_ready ? 'IsReadyYes' : 'IsReadyNo'}
```
limpbrains, line 466: IsReadyYes is checked in e2e tests

limpbrains (changes requested):

src/screens/Settings/Lightning/ChannelDetails.tsx
pwltr: I think it's ok to remove the test since it's only checking if the row exists. For debugging `is_channel_ready` hasn't really been useful so far

limpbrains: If we are removing is_channel_ready from tests, than we need to add is_usable check.

pwltr: The test was only checking if the row exists, now it will exist always (without developer mode on) so I don't see the point of testing this. @limpbrains thoughts?

limpbrains: Not just the row. Test was checking if LDK api return `channel.is_channel_ready`

limpbrains (approved):

PR #2050, opened 2024-07-09
chore: upgrade to react-native 0.74

pwltr:
### Description

- Upgrade react-native to 0.74.3 according to https://react-native-community.github.io/upgrade-helper/
- Update yarn version which is default for new RN projects
- ~Replace `lottie-react-native` with `react-native-skottie` (because of build error on Android)~
- Remove `TextEncoder` polyfill (now supported) and only leave `TextDecoder`
- Replace `stream-browserify` with `readable-stream` which is maintained
- Update some dependencies
- Fix new lint errors

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA Notes

Check for regression


pwltr:
~Currently blocked by https://github.com/react-native-clipboard/clipboard/issues/246~

.github/workflows/e2e-android.yml
```diff
@@ -88,10 +88,10 @@ jobs:
         env:
           HUSKY: 0
 
-      - name: Activate Gradle variables
-        run: |
-          cp .github/workflows/gradle.properties ~/.gradle/gradle.properties
-          patch -p1 -i ./.github/workflows/react-native-quick-crypto.patch
```
limpbrains, line 94: You probably don't need the patch anymore, but you need ~/.gradle/gradle.properties

ovitrif:
FYI would love to see this merged and shipped in the next version :), a user is already waiting for it in hope it fixes their frozen UI

pwltr:
> FYI would love to see this merged and shipped in the next version :), a user is already waiting for it in hope it fixes their frozen UI

I don't expect this to really solve any major issues, just some minor performance and bundle size gains. But will be a stepping stone for more performance improvements.

pwltr:
I think this is good to go now

package.json
```diff
@@ -143,16 +142,16 @@
...
+    "@react-native/eslint-config": "^0.74.85",
+    "@react-native/metro-config": "^0.74.85",
+    "@react-native/typescript-config": "^0.74.85",
     "@types/b4a": "^1.6.4",
     "@types/bip21": "^2.0.3",
     "@types/jest": "^29.5.12",
     "@types/lodash": "^4.17.0",
     "@types/mime": "^3.0.3",
-    "@types/react": "^18.2.78",
+    "@types/react": "^18.3.3",
```
limpbrains, line 154: So we still need this? I remember react now ship types with main package

package.json
```diff
@@ -181,5 +180,6 @@
   "react-native": {
     "net": "react-native-tcp-socket",
     "tls": "react-native-tcp-socket"
-  }
+  },
+  "packageManager": "yarn@3.6.4"
```
limpbrains, line 184: Are we switching to new yarn? Should we update it everywhere?

pwltr: It's default for new RN projects from 0.74 so yeah I think it's a good time to switch. I use it in some other projects and works well imo and a bit faster.

package.json
pwltr: Haven't heard anything like this, is this for react v18.3.0?

limpbrains: I was wrong, RN comes with TS, not react

package.json
```diff
@@ -99,28 +98,29 @@
     "react-native-haptic-feedback": "2.2.0",
     "react-native-image-picker": "7.1.2",
     "react-native-keyboard-accessory": "0.1.16",
-    "react-native-keychain": "github::synonymdev/react-native-keychain.git#refactor/android-data-store",
```
limpbrains, line 102: When I do `yarn` I see this message in yarn logs:
```
react-native-keychain@https://github.com/synonymdev/react-native-keychain.git#commit=7378c469afe81195487a3163295be425af79fa7a can't be foun
```
if i remove `.git` from package.json, seems to work fine

src/polyfills/textdecoder-polyfill.js
```diff
@@ -0,0 +1,113 @@
+(function (window) {
```
limpbrains, line 4: where is it from? Can you please add link to the original source

pwltr: sure, good point.

package.json
pwltr: fixed

limpbrains (approved):

PR #2051, opened 2024-07-09
FIX: test if channel is usable after restoring the wallet

limpbrains:
### Description

Also bump ldk backup server 

### Type of change

Refactoring

### Tests

Detox test

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -122,7 +122,7 @@ services:
 
   ldk-backup-server:
     container_name: ldk-backup-server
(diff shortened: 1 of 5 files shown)


Jasonvdb:
@limpbrains working now. rn-ldk is supposed to auto add persisted peers, no idea but when used from Bitkit it does not. So triggering lm.addPeers() from Bitkit + removing this auto peer removal I think has solved this for good.

Jasonvdb (approved):

catch-21:
No explicit testing performed

PR #2060, opened 2024-07-12
refactor: remove async storage

pwltr:
### Description

This replaces AsyncStorage with MMKV for the web relay cache. No need to have two key-value stores. Also adds Reactotron plugin for react-native-mmkv.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test


limpbrains:
Awesome find

pwltr:
Been testing this for a few days and seems to work. There is a rare issue (that only I seem to have) with the current web relay client cache where it becomes corrupted after a few weeks of use and I have to clear it in dev settings to repair. Not sure if that problem will be fixed but it doesn't seem like it got worse fwiw.

limpbrains (approved):

PR #2063, opened 2024-07-15
fix: Lazy load bottom-sheets

limpbrains:
### Description

- Move all bottom-sheets to dedicated component
- lazy load it
- do not mount bottom-sheets until it is needed

### Linked Issues/Tasks

### Type of change

Refactoring

### Tests

No test

### DIFF
--- a/src/navigation/root/RootNavigator.tsx
+++ b/src/navigation/root/RootNavigator.tsx
@@ -36,9 +36,6 @@ import BuyBitcoin from '../../screens/BuyBitcoin';
 import ScannerScreen from '../../screens/Scanner/MainScanner';
 import SettingsNavigator from '../settings/SettingsNavigator';
 import LightningNavigator from '../lightning/LightningNavigator';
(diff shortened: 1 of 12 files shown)


src/navigation/root/RootNavigator.tsx
```diff
@@ -263,8 +238,6 @@ const RootNavigator = (): ReactElement => {
 					onSuccess={onAuthSuccess}
 				/>
 			)}
-
-			<ForgotPIN />
```
pwltr: This sheet is not showing up, looks like it is because it is hidden below PIN screen.

https://github.com/user-attachments/assets/f42d1d19-1167-4968-a9f6-483b1a43cd4c

pwltr (changes requested):

src/navigation/root/RootNavigator.tsx
limpbrains: Thank you! Fixed it and added test for it

pwltr (approved):

PR #2071, opened 2024-07-17
feat: Add readme and local run bash script for E2E tests

catch-21:
### Description

Add readme to E2E directory to explain the purpose, how to run locally, and how they are run in the CI pipeline.
Add a bash script for running E2E tests locally.

### Linked Issues/Tasks

### Type of change

### Tests

- [x] Detox test
- [ ] Unit test
- [ ] No test

### Screenshot / Video

None

### QA Notes

None

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -1,0 +1,43 @@
+# End-to-End (E2E) Tests
+
+## Purpose
+
+This directory contains end-to-end (E2E) tests for Bitkit. These tests are designed to verify the functionality of the application, ensuring that it behaves as expected from the user's perspective. Using a local regtest environment, greater control and reproducibility is achieved compared to the mainnet for reliable and consistent testing outcomes.
+
+## Triggering Tests on Pull Requests
+
+E2E tests are automatically triggered on pull requests for this repository. When a new pull request is created or updated, the E2E tests run to ensure that the changes do not introduce any regressions. The results of these tests are displayed in the pull request checks, and merging is prevented if any test failures occur. This ensures that only code that passes all tests is integrated into the main codebase.
+
+## Running E2E Tests
+
+You can run the end-to-end (E2E) tests locally if the correct environment is set up:
+
+- For iOS Tests:
+  - The host must be running macOS.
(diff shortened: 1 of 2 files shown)


e2e/README.md
```diff
@@ -0,0 +1,43 @@
...
+- For iOS Tests:
+  - The host must be running macOS.
+  - iOS Simulator must be installed, which comes with Xcode.
+- For Android Tests:
+  - An Android Virtual Device (AVD) must be set up.
+  - The Android Debug Bridge (ADB) must be installed.
+
+Additionally, you may need to update the `devices.emulator.device.avdName` and/or `devices.simulator.device.type` settings in the .detoxrc.js configuration file to match your environment.
+
+### Using `act`
```
limpbrains: Did you try act? Does it work with our e2e tests?
I would rather suggest your bash script by default, looks solid

limpbrains (approved):

e2e/run.sh
```diff
@@ -0,0 +1,75 @@
...
+if [ "$BUILD_OPTION" == "build" ]; then
+  echo "Building the app..."
+  yarn e2e:build:$PLATFORM-$BUILD_TYPE
+else
+  echo "Skipping build..."
+fi
+
+# Start the Docker Compose environment
+docker compose -f docker/docker-compose.yml up -d
+sleep 2 # short wait for Electrum and LND
```
limpbrains: This will fail if the user is running tests for the first time, as the images have not been downloaded yet.

catch-21: If it doesn't have the images then it will pull them before proceeding to start containers. So I think it is ok?

catch-21: I could improve the wait here. I didn't need one at all on my machine as all services are ready before the first test begins.

catch-21: addressed

e2e/README.md
catch-21: Yes, it works flawlessly, for me at least. I've swapped the order of the two so bash script is before act.

JeanlChristophe:
@catch-21 please do no forget to add Projects information and sprint

PR #2073, opened 2024-07-17
feat: randomize 'awesome' i18n text

pwltr:
### Description

Use a variety of words for 'Ok' buttons and choose one at random.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test


src/utils/i18n/locales/en/common.json
```diff
@@ -56,9 +56,11 @@
   "ok": {
     "string": "OK"
   },
-  "awesome": {
-    "string": "Awesome!"
-  },
+  "ok_random": {
+		"string": "Awesome! Nice! Cool! Great! Fantastic! Sweet! Excellent! Terrific!",
```
pwltr: I chose to do it this way so we don't have to have the same number of words for every lang

limpbrains: It is better to use line break as a separator, some languages may have OK phrase in 2 words

src/utils/i18n/locales/en/common.json
```diff
@@ -56,9 +56,11 @@
...
+		"context": "A list of positive words to show the user after they have completed an action",
```
limpbrains, line 61: ❤️

limpbrains (approved):
Awesome! Cool! Заебца!

src/screens/Wallets/NewTxPrompt.tsx
```diff
@@ -85,7 +86,7 @@ const NewTxPrompt = (): ReactElement => {
 					<View style={styles.buttonContainer}>
 						<Button
 							style={styles.button}
-							text={t('awesome')}
+							text={getRandomOkText()}
```
limpbrains, line 89: this will generate new OK on every component re-render. Is it ok?

pwltr: I think it's fine none of the screens where this is used should render more than once.

Issue #2052, opened 2024-07-10
Android UI performance


pwltr:
Try to improve performance on Android:
- setup basic performance profiling (CI)
- improve startup time
- fix lag on initial load and when putting app to foreground
- fix navigation lag

pwltr:
Closing here, I think we're in a good place wrt performance now. Boot time is tracked separately [here](https://github.com/synonymdev/bitkit/issues/2014)

catch-21:
Tested on v130 d8f96e7d6475d2eef04ae08f9d9dd2c8894a8218
The performance of the android app does seem improved. There are still times that UI is unresponsive but that can also be the case on iOS.

PR #2075, opened 2024-07-19
fix: excessive rerendering on home screen

pwltr:
### Description

The Home screen and several components lower in the tree were rerendering way too many times because of bad selectors. Reselect's `createSelector` is not doing its job correctly so I'm adding a custom `createShallowEqualSelector` that does correct shallow equality checking needed for some larger selectors. Also we should not memoize selectors with `createSelector` by default. Should be assessed case by case whether it's needed. For most simple value lookups it is not needed.

### Linked Issues/Tasks

https://github.com/synonymdev/bitkit/issues/2052

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA Notes

Make sure components still rerender when they should.


src/AppOnboarded.tsx
```diff
@@ -111,7 +111,8 @@ const AppOnboarded = (): ReactElement => {
 					});
 				}
 				dispatch(updateUi({ isOnline: true }));
-				updateExchangeRates();
+				// FIXME: this runs too often
+				// updateExchangeRates();
```
pwltr, line 115: This needs to be moved to another place. Not sure why but this fires even when connection status did not change.

limpbrains (approved):
This is strange. Maybe it was because we are mutating state where we should not

pwltr:
> This is strange. Maybe it was because we are mutating state where we should not

I think this is purely about selectors. Docs are not very clear but looks like `createSelector` too only does strict equality `===` for inputs and outputs. For getting non-primitive types from selectors we need to do shallow equality specifically.

catch-21:
Nothing to specifically test. See https://github.com/synonymdev/bitkit/issues/2052#issuecomment-2255707531

Issue #2004, opened 2024-06-20
[Bug]: Refresh Failed toast often seen when reopening app from minimised


catch-21:
### Describe the bug

When I reopen the app from minimised state I often see 'Refresh Failed' 'An error occurred and couldn't refresh the wallet. Please try again.' toast message. A manual refresh will work and there appears to be no connection issue. We should only show this toast when there is a legitimate problem.

### Reproduce

1. Close Bitkit to minimised state
2. Wait a short while (2-5 minutes)
3. Reopen Bitkit from minimsed state
4. Observe 'Refresh Failed' error toast
5. Perform manual refresh and confirm there is no problem with refreshing

### Screenshots / Recording

https://github.com/synonymdev/bitkit/assets/74595920/1ddaad0b-cd00-4a93-a4ba-6ce4a2a319e9

### Operating system

Android 13 TKQ1.221114.001

### Bitkit version

8908053a3198a167c16a3062fb68282160a63112

catch-21:
This appears to be related to when new blocks are minded whilst app is minimised. Reproducible on iOS and android when mining blocks in staging env.

JeanlChristophe:
@pwltr we still have the same issue on latest internal release (v127). It's now happening very often.

catch-21:
Still occurring in v130

catch-21:
Verified in 2e3cf89d424d3f05fe5c2727ab02b3c9f318454a
Not seen it at all. LGTM.

pwltr:
Sadly this is not completely fixed yet. It's way less likely too appear now but still does sometimes. Reopening but can be considered lower prio now imo.

JeanlChristophe:
Let's see if it's happening again.

catch-21:
I have seen it once recently.

catch-21:
This is currently reproducible for me. It will appear roughly 10 seconds after running bitkit (cold start).
I am in a different region atm and I'm suspicious that could be related.

https://github.com/user-attachments/assets/3820d6ae-35ce-4336-a025-ca1e28268b21

JeanlChristophe:
@pwltr Failed toast is happening really frequently now.

PR #2077, opened 2024-07-22
fix: keep electrum connection on background

pwltr:
### Description

Don't intentionally disconnect from Electrum on app to background. This will keep connection up until the OS cuts network access. Reconnection is handled in beignet.

### Linked Issues/Tasks

Potentially improves https://github.com/synonymdev/bitkit/issues/2004

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA Notes

App should reconnect to Electrum

### DIFF
--- a/src/AppOnboarded.tsx
+++ b/src/AppOnboarded.tsx
@@ -5,7 +5,9 @@ import { useTranslation } from 'react-i18next';
 
 import RootNavigator from './navigation/root/RootNavigator';
 import InactivityTracker from './components/InactivityTracker';
+import { showToast } from './utils/notifications';
 import { startWalletServices } from './utils/startup';
+import { getOnChainWalletElectrum } from './utils/wallet';
 import { unsubscribeFromLightningSubscriptions } from './utils/lightning';
 import { useAppSelector } from './hooks/redux';
 import { dispatch } from './store/helpers';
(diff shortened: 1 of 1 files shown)


limpbrains:
I thought iOS will automatically kill an app in this case, not just close the connection

pwltr:
> I thought iOS will automatically kill an app in this case, not just close the connection

Doesn't seem like that is the case. I tested a bunch on Android and on the iOS simulator and it works fine. We also had it like this before https://github.com/synonymdev/bitkit/pull/1450, just the polling has moved to beignet.

Would be nice to get this in for this release because I think it fixes #2004

limpbrains (approved):

pwltr:
Tested this again and it does seem like it helps with issue 2004

catch-21:
Verified in https://github.com/synonymdev/bitkit/commit/2e3cf89d424d3f05fe5c2727ab02b3c9f318454a
I've used a real iPhone 15 (iOS 17) and iPhone 12 (iOS 14) in the farm, closed the app and left it for over 5 minutes, on restore the app was still running, iOS hadn't killed it. No errors shown either.

limpbrains:
Nice!

PR #2085, opened 2024-07-23
fix: add timeout and docker force recreate

limpbrains:
### Description

Added timeout to the job setup action, also run docker compose up with --force-recreate

### Type of change

Refactoring

### Tests

Detox test

### DIFF
--- a/.github/workflows/e2e-ios.yml
+++ b/.github/workflows/e2e-ios.yml
@@ -25,12 +25,13 @@ jobs:
 
       - name: Reset testing enviroment
         continue-on-error: true
+        timeout-minutes: 10
         run: |
           pkill -f react-native || echo 0
           pkill -f Simulator || echo 0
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-ios.yml
```diff
@@ -83,7 +84,7 @@ jobs:
       - name: Restart docker before last attempt
         if: steps.test1.outcome != 'success' && steps.test2.outcome != 'success' && steps.test3.outcome != 'success'
         run: |
-          cd docker && docker compose down -t 60 && docker compose up --quiet-pull -d && cd ..
+          cd docker && docker compose down -t 60 && docker compose up --force-recreate --quiet-pull -d && cd ..
```
catch-21, line 87: You probably don't need `down` if using `--force-recreate`

catch-21 (approved):

PR #2089, opened 2024-07-23
feat: check profile and contact on app restart

catch-21:
### Description

Add check in `Profile and Contacts -> Slashtags -> Can manage Slashtags Profile` for persisted profile and contact after app restart, followed by same checks after wallet restore.

### Linked Issues/Tasks

### Type of change

### Tests

- [x] Detox test
- [ ] Unit test
- [ ] No test

### Screenshot / Video

### QA Notes

### DIFF
--- a/e2e/onchain.e2e.js
+++ b/e2e/onchain.e2e.js
@@ -66,7 +66,7 @@ d('Onchain', () => {
 				let { label: wAddress } = await element(
 					by.id('QRCode'),
 				).getAttributes();
-				wAddress = await wAddress.replace('bitcoin:', '');
+				wAddress = wAddress.replace('bitcoin:', '');
 
 				await element(by.id('SpecifyInvoiceButton')).tap();
 				await element(by.id('TagsAdd')).tap();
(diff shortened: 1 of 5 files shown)


e2e/onchain.e2e.js
```diff
@@ -66,7 +66,7 @@ d('Onchain', () => {
 				let { label: wAddress } = await element(
 					by.id('QRCode'),
 				).getAttributes();
-				wAddress = wAddress.replace('bitcoin:', '');
+				wAddress = await wAddress.replace('bitcoin:', '');
```
limpbrains: replace is not an async function, no need to add await

limpbrains (changes requested):

e2e/onchain.e2e.js
catch-21: For the slashtag test, I found that without this the test would regularly timeout waiting for payment (more often than not locally). My assumption was it would sometimes use an undefined `wAddress` due to timing. Not had any timeouts since making this change.

limpbrains (approved):

e2e/onchain.e2e.js
catch-21: Removed these awaits on sync function call.

PR #2134, opened 2024-08-12
fix: always show selected fee on FeeRate screen

### LINKED ISSUE #1896: [Bug]: Transaction suggested fee rate can display Custom when current fee rate allows for Normal
### Describe the bug

Minor UI bug.

When sending a transaction, it is possible for the fee rate to display as 'Custom' when it is actually using the Normal value. First mentioned here: https://github.com/synonymdev/bitkit/issues/1619#issuecomment-2022813353.

### Reproduce

1. Begin constructing a transaction where the sent amount is high enough that the Normal fee rate can be used (it could matter if it is close to not being allowed, or maybe it was recently not allowed)
2. View the suggested fee rate is Custom
3. Tap on Custom
4. Observe that only Custom is seen even though the Normal fee rate is being applied

### Screenshots / Recording

see https://github.com/synonymdev/bitkit/issues/1619#issuecomment-2122994907.

### Operating system

Android 13 TKQ1.221114.001

### Bitkit version

v118

### Log output

_No response_

limpbrains:
### Description

It is possible that selected fee might be above maximum, we should still show it, because it is already has been selected by the user.

https://github.com/user-attachments/assets/2f3ff41b-3e5e-47eb-a7e4-5c35861f1797

### Linked Issues/Tasks

closes #1896

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/hooks/fees.ts
+++ b/src/hooks/fees.ts
@@ -1,4 +1,5 @@
 import { useAppSelector } from '../hooks/redux';
+import { onChainFeesSelector } from '../store/reselect/fees';
 import { FeeText, TFeeText } from '../utils/fees';
 
 /**
@@ -7,7 +8,7 @@ import { FeeText, TFeeText } from '../utils/fees';
  * @returns {TFeeText}
  */
 export const useFeeText = (satsPerByte: number): TFeeText => {
-	const feeEstimates = useAppSelector((store) => store.fees.onchain);
+	const feeEstimates = useAppSelector(onChainFeesSelector);
 
 	let feeText = FeeText.none;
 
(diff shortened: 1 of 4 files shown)


catch-21:
Is it not more confusing to show one of the three non-custom options that are all invalid for current tx context but still hide the other two invalid options?

Fwiw, I think we should never hide any of them, just grey them out if invalid.

limpbrains:
Agree, let's gray them out

limpbrains:
image

catch-21:
Greying out is definitely better.
You may have observed whilst trying to reproduce https://github.com/synonymdev/bitkit/issues/1896 that, if the user has selected 'Normal' and then changes the value to be sent, it can sometimes change the selected to 'Custom' with a lower fee rate. This does happen if for example using '7000' first and changing it to '700'.


https://github.com/user-attachments/assets/1a0da539-ece0-4fcd-85e4-94db2a499667

This is how I imagine it should always behave when the currently selected rate is invalid because we are prohibiting use of this higher fee rate when the output is too small. Otherwise we are allowing a prohibited rate, or showing the rate that will not actually be used.

pwltr (approved):

PR #2135, opened 2024-08-12
feat(transfer): transfer notifications

### LINKED ISSUE #2094: Local notification to open Bitkit
**Description**

When a user makes a request to open a channel, the user must come back to the app in between the time the channel is ready to be opened and before channel expiry. However, since the transaction can take several hours or even several days some users may forget and miss the time window, ending up with an expired channel order.

The problem is that the BT order is only valid for 2 days. We therefore need to warn the user to open Bitkit.

Ex: https://synonymworkspace.slack.com/archives/C0726U60V6Y/p1721769530317439

**Solution**
Schedule local notifications telling the user that he must open Bitkit in order to continue the channel opening process.
- First notification after 24 hours
- Second notification after 40 hours (this notification will not be shown if the channel was opened before then)

Note:
We don't check if the channel purchase transaction actually got confirmed (will be solved with wakeToOpen), so there can be the case where fees are spiking and the transaction doesn't confirm before order expiry. In this case the notifications will still (falsely) appear.

pwltr:
### Description

Adds local notifications to let the user know to open the app when channel open is ready.

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/2094

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### QA
For easier testing you can change the timestamp in `utils/blocktank.ts` to something earlier like `date1.setSeconds(date1.getSeconds() + 15);`.


src/store/utils/blocktank.ts
```diff
@@ -448,3 +455,48 @@ export const refreshAllBlocktankOrders = async (): Promise<Result<string>> => {
 	);
 	return ok('All orders refreshed.');
 };
+
+export const scheduleNotifications = async (): Promise<void> => {
+	// Request permissions (required for iOS)
+	await notifee.requestPermission();
```
limpbrains, line 461: what if I deny, will it throw and break refreshAllBlocktankOrders function ?

src/store/utils/blocktank.ts
```diff
@@ -448,3 +455,48 @@ export const refreshAllBlocktankOrders = async (): Promise<Result<string>> => {
...
+		const date2 = new Date(Date.now());
+		date2.setHours(date2.getHours() + 40);
+		const trigger2: TimestampTrigger = {
+			type: TriggerType.TIMESTAMP,
+			timestamp: date2.getTime(),
+		};
+
+		const options = {
+			title: 'Important: Bitkit Transfer',
+			body: 'Open Bitkit to complete your transfer',
```
limpbrains: localization needed

package.json
```diff
@@ -38,6 +38,7 @@
     "@formatjs/intl-pluralrules": "5.2.12",
     "@formatjs/intl-relativetimeformat": "11.2.12",
     "@gorhom/bottom-sheet": "4.6.3",
+    "@notifee/react-native": "^7.8.2",
```
limpbrains: better to lock the exact version

limpbrains (changes requested):

src/store/utils/blocktank.ts
pwltr: fixed

package.json
pwltr: fixed

src/store/utils/blocktank.ts
pwltr: The promise will still resolve, but with `AuthorizationStatus.DENIED`. Code will still run but no notifications will show.

limpbrains (approved):

PR #2145, opened 2024-08-19
fix: yarn - do not update checksum by default

limpbrains:
### Description

This PR disables Yarn's automatic checksum updates to enhance security and maintain consistency. It ensures that dependencies remain unchanged unless explicitly intended, preventing potential unauthorized or accidental modifications. This is crucial for a stable and secure build process.

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,5 +1,3 @@
 nodeLinker: node-modules
 
-checksumBehavior: 'update'
-
 yarnPath: .yarn/releases/yarn-3.6.4.cjs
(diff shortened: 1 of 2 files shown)


.yarnrc.yml
```diff
@@ -1,5 +1,5 @@
 nodeLinker: node-modules
 
-checksumBehavior: 'update'
+# checksumBehavior: 'update'
```
pwltr: can remove the comment

pwltr (approved):

pwltr:
@limpbrains Why does the `react-native-keychain` have a different checksum for you than me? I can't install it

limpbrains:
I was afraid of that. Same issue blocked BW migration to yarn v3.
We might want to revert it

pwltr:
meh, looks like packages pulled from git can have slightly different content depending on OS when building with yarn v3. Yeah let's revert.

PR #2199, opened 2024-09-05
fix: stop auto removing peers

Jasonvdb:
### Description

removeUnusedPeers was added a few years back to avoid connecting to peers that wallet did not have a channel with or was not a Blocktank node. On restore this was being triggered before syncing channels with redux state so it was removing peers before realising there was actually an active channel. 

I think auto dropping peers might be a bit premature of an optimisation so I moved it to a manual button for now. A user might have the intention of adding a peer and opening a channel with that peer at a later stage and with this method the peer would be removed on each start up.

### Linked Issues/Tasks

https://github.com/orgs/synonymdev/projects/12/views/1?pane=issue&itemId=69388158

https://github.com/synonymdev/react-native-ldk/issues/259

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

### QA Notes

After adding an external peer, restore from remote backup and peer should still be there. Or if there was a channel from another LSP it should still be usable.

### DIFF
--- a/src/screens/Settings/Lightning/Channels.tsx
+++ b/src/screens/Settings/Lightning/Channels.tsx
@@ -45,6 +45,7 @@ import {
 	recoverOutputsFromForceClose,
 	refreshLdk,
 	setupLdk,
+	removeUnusedPeers,
 } from '../../../utils/lightning';
 import { showToast } from '../../../utils/notifications';
 import {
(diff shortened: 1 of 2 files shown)


limpbrains:
I've merged this PR into my #2051 PR with tests. They are still failing

limpbrains (approved):

PR #2201, opened 2024-09-05
fix: disable RBF

limpbrains:
### Description

We still have issues with RBF, disable it by default

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -32,7 +32,7 @@ const persistConfig = {
 	key: 'root',
 	storage: reduxStorage,
 	// increase version after store shape changes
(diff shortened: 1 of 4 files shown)


src/store/index.ts
```diff
@@ -32,7 +32,7 @@ const persistConfig = {
 	key: 'root',
 	storage: reduxStorage,
 	// increase version after store shape changes
-	version: 45,
+	version: 43,
```
limpbrains, line 35: I guess 44 was stealth mode, which is removed by now

pwltr (approved):

Issue #1934, opened 2024-06-03
[Bug]: Balances displayed before all values have been determined on wallet restore


catch-21:
Raising off the back of https://github.com/synonymdev/bitkit/issues/1841.

On wallet restore, the main screen is displayed before all balances have been determined. This can mean showing 0 for some balance (or flicking between some value and 0) for a short while before displaying the correct value.

We could either wait at the rocket or give some visual indication on the main screen to minimise confusion, such as some loading icon instead of 0.

Couple of examples of how it behaves atm on restore for two of my wallets:

https://github.com/synonymdev/bitkit/assets/74595920/dbca6291-c876-4731-a9b7-63914ad4a9e3

https://github.com/synonymdev/bitkit/assets/74595920/49673845-6e68-4788-8169-e48c31e360f1

JeanlChristophe:
Last month we spoke about this [here](https://synonymworkspace.slack.com/archives/C01J0QPPV08/p1714646054340819).

aldertnl:
I think we decided on using a spinner instead of displaying 0 in case the balance is still loading.

JeanlChristophe:
That's what I thought too but I preferred to check first with you since i am not seing it on Figma. Which spinner do you would like to use?

aldertnl:
Our regular, default spinner (see mockup below). 

https://github.com/synonymdev/bitkit/assets/81586998/ff756416-c081-4b62-9a6f-09bf8a74157f

pwltr:
This looks like a regression to me, the wallet didn't need loading time before. We should look into why this is happening more before adding a loading UI.

catch-21:
@pwltr It is a regression. One of the previous solutions caused the long black screen bug on startup. Check the issue referenced in the description for history. If we're waiting on data retrieval its probably best that the UI isn't complete blocked.

pwltr:
I think this issue might've been resolved by https://github.com/synonymdev/beignet/pull/75. With LN channels it can take a little while for all funds to appear but I think this is acceptable.

https://github.com/user-attachments/assets/634c4e5e-b6da-4e99-824e-e4d81ac0a552

catch-21:
@pwltr I don't see any improvement with my wallet. I started this recording 5 seconds after restoring and seeing 0 for all balances. Takes a little while to see spending balance and a manual refresh to get my savings.

https://github.com/user-attachments/assets/ea09d7da-f9c5-454d-a231-4a3473b213ad

[...]

PR #2210, opened 2024-09-06
fix: prevent wallet from beeing initalized twice

### LINKED ISSUE #2177: [Bug]: Transfer RBFs previous sent transaction
### Describe the bug

See screen recording

### Screenshots / Recording

https://github.com/user-attachments/assets/502a0d39-241b-4866-a64b-0c0f5905f1c1

### Operating system

iOS, Android

### Bitkit version

f3b10fce

### Log output

_No response_

limpbrains:
### Description

After new wallet creation `setupOnChainWallet` was called twice resulting in 2 instances of Beignet competing against each other. 
I'm fixing it by `walletExists: true` setter to be executed later, after the `createDefaultWallet` function is completed

### Linked Issues/Tasks

closes #2177

### Type of change

Bug fix

### Tests

No test

### QA Notes

- make sure RBF is enabled by default
- create new wallet, do not restart the app
- create 2-3 onchain transactions
- make sure you don't have any error messages

### DIFF
--- a/src/store/slices/wallet.ts
+++ b/src/store/slices/wallet.ts
@@ -32,7 +32,6 @@ export const walletSlice = createSlice({
 	initialState: defaultWalletStoreShape,
 	reducers: {
 		createWallet: (state, action: PayloadAction<IWallets>) => {
-			state.walletExists = true;
 			state.wallets = {
 				...state.wallets,
 				...action.payload,
@@ -186,6 +185,9 @@ export const walletSlice = createSlice({
 		resetExchangeRates: (state) => {
 			state.exchangeRates = defaultWalletStoreShape.exchangeRates;
 		},
+		setWalletExits: (state) => {
+			state.walletExists = true;
+		},
 	},
 });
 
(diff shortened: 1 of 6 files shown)


src/store/slices/wallet.ts
pwltr, line 1: Do we need to split `createWallet` into 2 reducers? They are dispatched at the same time anyway:

```
dispatch(createWallet(response.value));
dispatch(setWalletExits());
```

pwltr (changes requested):

src/store/slices/wallet.ts
pwltr: In `createDefaultWallet`, `createWallet()` action is also dispatched but not `setWalletExits()`. Doesn't seem right.

pwltr: Ok I see what's going on, it's confusing that it has to dispatch `createWallet()` twice.

pwltr:
Can confirm this seems to fix https://github.com/synonymdev/bitkit/issues/2177 and https://github.com/synonymdev/bitkit/issues/1934

src/store/slices/wallet.ts
limpbrains: The whole thing is confusig and compicated

pwltr (approved):

Issue #1780, opened 2024-05-03
[Bug]: No confetti or transaction activity on requesting spending balance via CJIT


catch-21:
### Describe the bug

Although the channel setup completed (see Created cjitEntry ef5d90a2-7d12-417b-8163-30db71dc814e) I did not see any confetti or transaction activity. After manually refreshing the wallet the balance updated to reflect my new spending balance. After a short wait and restarting the app I still do not see the transaction activity. Other new wallet have not seen this issue. This is occurring on a wallet that recently had a force closure (could be relevant).

It is possible to produce a CJIT invoice, but if Bitkit isn't in the foreground when the invoice is paid then it is possible for it to not detect the 'Received' event. This leads to correct spending balance but no activity to assure the user the payment was received.

### Reproduce

1. Start with a Bitkit wallet with no LN channels
2. Tap Receive and produce CJIT invoice
3. Screenshot the invoice
4. Minimise Bitkit
5. Scan to pay the invoice from a different wallet. The LN payment will succeed provided Bitkit is still running in the background. If Bitkit has been killed then LN payment fails and the invoice goes from Pending to Holding to Cancelled state.
6. Now you can kill Bitkit (or leave it for the OS to dispose of after some time)
7. Start Bitkit again
8. Observe new spending balance, but no confetti and no 'Received' activity item. Also observe open Lightning connection in Advanced settings.

### Screenshots / Recording

This shows my correct balance and LN channel is open but no transaction activity

https://github.com/synonymdev/bitkit/assets/74595920/34274d3d-c245-40ab-965b-608d607527ba

### Operating system

Android/iOS

### Bitkit version

1.0.4 (135)

### Log output

[bitkit_ldk_logs_1714731123960.zip](https://github.com/synonymdev/bitkit/files/15198712/bitkit_ldk_logs_1714731123960.zip)

limpbrains:
I can't reproduce this issue

limpbrains:
I conducted extensive testing. Opened 5 channels with CJIT. Received several payments. I always get confetti.
If this issue does not recur in the next testing session, we can consider it resolved. It could be linked to the wallet's malfunction or another unknown cause.

JeanlChristophe:
Same. Closing the ticket.

JeanlChristophe:
Joe has a similar issue, except that its balance is not updated.

pwltr:
[...]

PR #2225, opened 2024-09-11
fix(receive): force refresh LDK on channel open

pwltr:
### Description

Updating to force `refreshLdk()` when a channel is opened. Also makes the code around CJIT more readable and adds logs. Removing `InteractionManager.runAfterInteractions` lines since it doesn't do anything.

### Linked Issues/Tasks

https://github.com/synonymdev/bitkit/issues/1780

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### DIFF
--- a/src/store/utils/activity.ts
+++ b/src/store/utils/activity.ts
@@ -6,14 +6,14 @@ import { EPaymentType } from 'beignet';
 import { onChainTransactionToActivityItem } from '../../utils/activity';
 import { formatBoostedActivityItems } from '../../utils/boost';
 import { sleep, vibrate } from '../../utils/helpers';
-import { getChannels } from '../../utils/lightning';
+import { getOpenChannels } from '../../utils/lightning';
 import { getCurrentWallet } from '../../utils/wallet';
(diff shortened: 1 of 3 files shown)


pwltr:
On Staging it seems to hit the fallback code with log line: `LOG  No CJIT entry found for funding_txid. Trying channel size.`

It looks like the following code block is rarely hit, probably because BT takes longer to update the db entry than we are waiting for. I don't see this as being an issue, do you @limpbrains ?

```
// Try to find the CJIT entry for this channel by funding_txid.
	let cJitEntry: ICJitEntry | undefined;
	let i = 0;
	while (!cJitEntry && i < 5) {
		await sleep(1000); // wait until Blocktank has updated its database.
		await updatePendingCJitEntries();
		cJitEntry = cJitEntries.find((entry) => {
			return entry.channel?.fundingTx.id === channel.funding_txid;
		});
		i++;
	}
```

limpbrains:
I agree it is not a big issue

pwltr:
Let's try and see if this improves it.

limpbrains (approved):

PR #2227, opened 2024-09-11
fix: force wallet refresh after onchain tx broadcasted

limpbrains:
### Description

Normally beignet listens to the wallet addresses and automatically starts the update. But just in case subscription failed, run it manually. It will not hurt, since the refreshWallet method is protected by a Promise and can't be run twice at the same time

This should prevent errors when wallet tries to use UTXOs that is already spent

### Type of change

Bug fix

### Tests
No test

### QA Notes

Try to make 2 transactions quickly

### DIFF
--- a/src/screens/Wallets/Send/ReviewAndSend.tsx
+++ b/src/screens/Wallets/Send/ReviewAndSend.tsx
@@ -1,18 +1,56 @@
+import { TInvoice } from '@synonymdev/react-native-ldk';
+import { validateTransaction } from 'beignet';
 import React, {
 	ReactElement,
+	ReactNode,
 	memo,
 	useCallback,
+	useEffect,
 	useMemo,
 	useState,
-	useEffect,
-	ReactNode,
 } from 'react';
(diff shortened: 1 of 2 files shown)


pwltr:
If I understand correctly this issue is still there:

https://github.com/user-attachments/assets/5fb4ee45-34a4-4f74-8ea7-6d64003b59c1

limpbrains:
New version with refresh after boost broadcast, can you please try it

pwltr:
Still running into to this:

https://github.com/user-attachments/assets/a0925200-fab4-4f0a-8b2c-157096da74b2

Also after boosting with RBF the balance is briefly set to 0:

https://github.com/user-attachments/assets/20bed3bf-ad1c-44ba-b10d-3d1eb732ded8

pwltr:
Codewise this feels like this logic needs to be in a central place (probably beignet) instead of being spread out in multiple places. UTXO state should be always 'fresh', maybe after `broadcastTransaction`?

EDIT: I think if the automatic update fails it needs to try again

limpbrains:
I agree there should be a better way

PR #2231, opened 2024-09-12
fix: don't use StrongBox on Android

### LINKED ISSUE #2014: Android startup time is slow
This task is to coordinate the work to improve the startup time on Android, ie. opening the app when it's not in background.

In my case, with Pixel 7 Pro it's a random time that always exceeds 5 sec, and there are times when it hangs for as long as ± 16 sec.

## Refs
See …`.slack.com/archives/C01H106JRQB/p1718628911888799`

pwltr:
### Description

Adds a patch to remove usage of StrongBox secure element on Android.

> Although StrongBox is a little slower and resource constrained (meaning that it supports fewer concurrent operations) compared to TEE, StrongBox provides better security guarantees against physical and side-channel attacks. If you want to prioritize higher security guarantees over app resource efficiency, we recommend using StrongBox on the devices where it is available. Wherever StrongBox isn't available, your app can always fall back to TEE to store key materials.

See https://synonymworkspace.slack.com/archives/C01H106JRQB/p1725562106199429

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/2014


ovitrif (approved):
LGTM (looks good to me) 👏

PR #2242, opened 2024-09-18
fix(receive): CJIT activity not added when app in background

### LINKED ISSUE #1780: [Bug]: No confetti or transaction activity on requesting spending balance via CJIT

pwltr:
### Description

In the common scenario where a user created a CJIT invoice and then paid from another LN wallet on the same phone (putting Bitkit in the background) a couple of issues arise.

- `refreshLdk()` is calling `lm.syncLdk()` which hangs when the app is in the background blocking further execution
- `addCJitActivityItem` only checked for open channels, it should also consider pending channels (currently opening)
- `redux-persist` doesn't flush the state to the database when the app is in the background because it uses `setInterval()`, quickfix by manually calling `flush()` for now

### Note

I wasn't able to reproduce this issue on iOS since network connection was always cut immediately for me when putting the app to the background causing the hold invoice to never resolve. Hopefully these fixes work on iOS as well.

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/1780

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Screenshot / Video

https://github.com/user-attachments/assets/34f3e203-577c-44c7-8a38-9dc0810221e0

### QA Notes

Use reproduction steps from https://github.com/synonymdev/bitkit/issues/1780

### DIFF
--- a/src/store/utils/activity.ts
+++ b/src/store/utils/activity.ts
@@ -1,20 +1,19 @@
-import { ICJitEntry } from '@synonymdev/blocktank-lsp-http-client';
 import { TChannel } from '@synonymdev/react-native-ldk';
 import { ok, Result } from '@synonymdev/result';
 import { EPaymentType } from 'beignet';
 
 import { onChainTransactionToActivityItem } from '../../utils/activity';
 import { formatBoostedActivityItems } from '../../utils/boost';
-import { sleep, vibrate } from '../../utils/helpers';
-import { getOpenChannels } from '../../utils/lightning';
+import { vibrate } from '../../utils/helpers';
+import { getOpenChannels, getPendingChannels } from '../../utils/lightning';
 import { getCurrentWallet } from '../../utils/wallet';
 import { dispatch, getBlocktankStore } from '../helpers';
 import { addActivityItem, updateActivityItems } from '../slices/activity';
 import { updateSettings } from '../slices/settings';
 import { closeSheet } from '../slices/ui';
 import { EActivityType, TLightningActivityItem } from '../types/activity';
-import { updatePendingCJitEntries } from './blocktank';
 import { showBottomSheet } from './ui';
+import { persistor } from '..';
 
 /**
  * Attempts to determine if a given channel open was in response to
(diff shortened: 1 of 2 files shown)


src/store/utils/activity.ts
```diff
@@ -33,31 +34,15 @@ export const addCJitActivityItem = async (channelId: string): Promise<void> => {
...
-	let cJitEntry: ICJitEntry | undefined;
-	let i = 0;
-	while (!cJitEntry && i < 5) {
-		await sleep(1000); // wait until Blocktank has updated its database.
-		await updatePendingCJitEntries();
-		cJitEntry = cJitEntries.find((entry) => {
-			return entry.channel?.fundingTx.id === channel.funding_txid;
-		});
-		i++;
-	}
```
pwltr, line 48: I'm removing this code block since this check never succeeds, probably because Blocktank takes too long to update.

coreyphillips (approved):

PR #2271, opened 2024-10-01
fix(transfer): fix force transfer bottom sheet

### LINKED ISSUE #2190: "Keep app open" card
The toast is still there even atfer the channel has been created.
Toast still display after a force close app.

Image

pwltr:
### Description

Keep `<ForceTransfer />` mounted so the interval can run.

This card is for when a channel couldn't be cooperatively closed after a transfer to savings. It will keep trying for 30m and then offer to force close. For example if the channel peer is offline. If the user wants Bitkit to keep retrying the coop close they need to keep Bitkit open, otherwise on next start it should offer to force close.

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/2190

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### Recording

https://github.com/user-attachments/assets/15a11248-dc1d-47c0-90a5-2fb43d88fe2a

### DIFF
--- a/e2e/channels.e2e.js
+++ b/e2e/channels.e2e.js
@@ -275,6 +275,9 @@ d('Transfer', () => {
 
 		// TODO: mine single blocks and check updated transfer time
 
+		// Sometimes the channel is only opened after restart 
+		await device.launchApp();
+
 		// wait for channel to be opened
 		await waitForActiveChannel(lnd, ldkNodeId);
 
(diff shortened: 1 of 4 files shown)


limpbrains:
We need to improve e2e tests. they are failing a lot

limpbrains (approved):

PR #2311, opened 2024-10-14
fix: reduce PIN attempts in Send and ChangePIN flows

limpbrains:
### Description

 - extract all complex PIN logic to the hook
 - use this hook in all 3 flows: General auth, Change PIN, Send flows
 - disable annoying eslint rule, we don't need it, we have prettier

### Type of change

Bug fix

### Tests

Detox test


pwltr (approved):
tested, LGTM

src/components/PinPad.tsx
```diff
@@ -302,16 +178,7 @@ const styles = StyleSheet.create({
...
-		justifyContent: 'center',
-		marginBottom: 48,
-	},
-	dot: {
-		width: 20,
-		height: 20,
-		borderRadius: 10,
-		marginHorizontal: 12,
-		borderWidth: 1,
+		marginBottom: 'auto',
```
pwltr: Small spacing issue here between the biometry button and the attempts text 
image

pwltr: Also here (horizontal padding)
image

limpbrains: fixed this

PR #2315, opened 2024-10-21
fix(transfer): refresh LDK after CJIT

pwltr:
### Description

Need to refresh LDK after a CJIT channel has been opened for the channel to become usable.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test


limpbrains (approved):
This is a type of thing that should be fixed in RN-LDK
cc @Jasonvdb

PR #2324, opened 2024-10-31
feat(send): add unified onchain/lightning switch

### LINKED ISSUE #2003: [Feature]: Add ability to switch between sending from savings & spending (unified QR)
### Describe the problem

No way to switch between “Savings” and “Spending” after scanning a unified QR code

### Describe the solution

- [x] Balance Toggle Functionality: It would be nice to allow the sender to switch between “Savings” and “Spending” after scanning a unified QR code (see mockup video)
- [x] Fixed States for Toggle:

	•	For On-chain-only QR codes, the toggle will be set to a fixed “Savings” state (see mockup video).
	•	For Lightning-only invoices, the toggle will be set to a fixed “Spending” state.
	
- [x] Max amount: Added max amount modal with short explanation for end user for when ‘max’ amount is tapped in spending balance send flow. https://synonymworkspace.slack.com/archives/C01H106JRQB/p1712855096363569?thread_ts=1712850323.807499&cid=C01H106JRQB

### Additional context

_No response_

pwltr:
### Description

- Add ability to switch between sending from savings & spending for 0 amount unified QRs. For this I added a new field in the `ui` reducer `paymentMethod` that is set in `handleData` (last step of `processUri`) and read in the send flow. There is also a new QrType "Unified"
- Refactor scanning logic to avoid duplication and improve readability. The goal is to have clear steps for processing any URI we support: 1. parse 2. validate 3. process/transform for handling 4. actually handle (UI actions)
- Add e2e tests in new file `send.e2e.js`. I would like to have this be the main file for testing send flow since other tests are concerned with only onchain or lightning at the moment, but edge cases need to be handled for all variants.

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/2003

### Type of change

### Tests

- [x] Detox test
- [ ] Unit test
- [ ] No test

### Screenshot / Video

https://github.com/user-attachments/assets/350d06ea-f895-4519-b75f-9913a037e231

### QA Notes

I tested basic functionality as well as max amount, coin selection and slashpay.


e2e/helpers.js
```diff
@@ -47,22 +47,9 @@ export const sleep = (ms) => {
...
-	}
-};
-
 export const isButtonEnabled = async (element) => {
-	try {
-		await expect(element).tap();
-		return true;
-	} catch (e) {
-		return false;
-	}
```
pwltr, line 65: These didn't work or broke a while back. New one is reliable.

src/utils/lnurl.ts
```diff
@@ -221,15 +220,15 @@ export const handleLnurlChannel = async ({
 			? i18n.t('other:lnurl_channel_success_msg_peer', { peer })
 			: i18n.t('other:lnurl_channel_success_msg_no_peer'),
 	});
-	return ok({ type: EQRDataType.lnurlChannel });
+	return ok('');
```
limpbrains, line 223: OK :)

pwltr: Just to clarify the `ok` result was not used anywhere, this makes it clear we don't have to care for this return value.

limpbrains: agree

limpbrains:
Unit tests are failing

limpbrains:
When my wallet has no LN channels and I scan LN+onchain qrcode, I get this error:


Should it be so?

limpbrains:
I aslo have similar error if I have 0 Onchain balance and scanning QRcode with Onchain address only

pwltr:
> When my wallet has no LN channels and I scan LN+onchain qrcode, I get this error: 
> 
> Should it be so?

Looks like this one is a regression.

> I aslo have similar error if I have 0 Onchain balance and scanning QRcode with Onchain address only
> 
> 

This one is expected behavior.

limpbrains:
Tapping on Available balance when trying to send from Spending account works only once
And stops working after switching to Savings and back

https://github.com/user-attachments/assets/f5ee62f9-4a30-4cc0-94c9-12f3cfbd58e6

limpbrains:
When scan Unified QR code with onchain and LN amount exceeding your balance, bitkit opens Spending and allows me to specify the amount. And then fails.

https://github.com/user-attachments/assets/9f7bfe9d-ae6a-4005-91c1-3895b427f248

I actually can switch to Savings in this case. 
IDK what should happend in this case, maybe just an error and nothing else?

limpbrains:
Sometimes Bitkit doesn't allow me to choose how to send. 
LN is forced in this example, while I have funds to pay via LN or on-chain.

https://github.com/user-attachments/assets/42d832c4-bd21-41fe-9884-c9accbcc8594

limpbrains:
> This one is expected behavior.

Error message doesn't make sense

limpbrains:
After scanning the Unified QR without an amount and switching to Savings, I get the Spending MAX amount when I tap on Available balance.

https://github.com/user-attachments/assets/19c5c0d3-f192-474a-b59c-533d5fa4b26d

pwltr:
> Sometimes Bitkit doesn't allow me to choose how to send. LN is forced in this example, while I have funds to pay via LN or on-chain.
> 
>  kek2.mov

This feature is only for 0 amount unified invoices. See description at the top.

pwltr:
> Tapping on Available balance when trying to send from Spending account works only once And stops working after switching to Savings and back

Fixed and added tests

pwltr:
> > This one is expected behavior.
> 
> Error message doesn't make sense

Fixed. Toast says "More ₿ needed to pay this Bitcoin invoice." now like before.

pwltr:
> When scan Unified QR code with onchain and LN amount exceeding your balance, bitkit opens Spending and allows me to specify the amount. And then fails.
> I actually can switch to Savings in this case. IDK what should happend in this case, maybe just an error and nothing else?

Fixed and added tests. In this case the UI should only allow spending from savings and set the amount to 0 (and showing a error toast for the amount)

pwltr:
Think I fixed everything, hopefully it's good now.

limpbrains (approved):

PR #2329, opened 2024-11-04
fix: Show Info notification for LNURL-withdraw success

### LINKED ISSUE #2326: [Bug]: Do not show green notification during LNURL withdraw
### Describe the bug

When trying to withdraw from LNmarkets you have green notification which is confusing. 
We need to change it to yellow and change the text, mention that payment is still in process and we are waiting for it.

### Reproduce

1. Go to lnmarkets withdraw
2. withdraw using LNURL-withdraw
3. See Success notification but not the confetty 

### Screenshots / Recording

_No response_

### Operating system

iOS, Android

### Bitkit version

_No response_

### Log output

_No response_

limpbrains:
### Description

Instead of Success show Info, because payment is still not received yet.
Also alter text

### Linked Issues/Tasks

closes #2326

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/utils/lnurl.ts
+++ b/src/utils/lnurl.ts
@@ -334,13 +334,6 @@ export const handleLnurlWithdraw = async ({
 			});
 			return err(jsonRes.reason);
 		}
-
-		showToast({
-			type: 'success',
-			title: i18n.t('other:lnurl_withdr_success_title'),
-			description: i18n.t('other:lnurl_withdr_success_msg'),
-		});
-
 		return ok('');
 	} catch (e) {
 		console.log(e.message);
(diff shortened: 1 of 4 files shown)


src/utils/lnurl.ts
```diff
@@ -337,7 +337,7 @@ export const handleLnurlWithdraw = async ({
 		}
 
 		showToast({
-			type: 'success',
+			type: 'info',
 			title: i18n.t('other:lnurl_withdr_success_title'),
 			description: i18n.t('other:lnurl_withdr_success_msg'),
 		});
```
pwltr: This toast shows at the same time as the one in the component no? Can remove one of them I think

pwltr (changes requested):

src/utils/lnurl.ts
limpbrains: You are right, removed

pwltr (approved):

Issue #2319, opened 2024-10-22
[Feature]: v52 MINOR Changes


JeanlChristophe:
### Describe the problem

- These changes are ready to be implemented in frontend and backend.
- Please check the interactive prototype for user experience and interactions
- Please check the static frames in Figma for design nuances and edge cases
- This version excludes ongoing experimental design work.	
New user: https://www.figma.com/proto/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?page-id=26999%3A127282&node-id=26999-132385&node-type=canvas&viewport=97%2C-174%2C0.06&t=Mezz22lLvIviLzJ6-8&scaling=scale-down&content-scaling=fixed&starting-point-node-id=26999%3A132385&show-proto-sidebar=1&hide-ui=1
Active user: https://www.figma.com/proto/ltqvnKiejWj0JQiqtDf2JJ/Bitkit-Wallet?page-id=26999%3A127282&node-id=26999-132389&node-type=canvas&viewport=97%2C-174%2C0.06&t=Mezz22lLvIviLzJ6-8&scaling=scale-down&content-scaling=fixed&starting-point-node-id=26999%3A132389&show-proto-sidebar=1&hide-ui=1

### Describe the solution

**MINOR CHANGES:**

- [x] Receive > Switch apps edge case: Added toast warning (triggered when switching between apps while trying to receive).
- [x] Receive > NFC Card Campaigns: Extended UI views for Bitkit NFC Card campaigns to cover edge cases and errors.
- [x] Receive > CJIT > Geoblock: Introduced a geoblock view in the CJIT receive flow with a link for manual setup.
- [x] ~~Send > Contact/Lightning > Geoblock: Added a geoblock view in the Send flow for transaction failures due to geoblock policies.~~
- [x] Advanced Lightning Onboarding > Manual Setup > Confirm: Enabled custom fee setting for manual Lightning setup (accessible via pencil icon on network fee overview).
- [x] Profile > Edit: Added a button to delete profile information.
- [x] :Changed payment animation, reused part of the LN sync animation. (Quickpay is accessible in the prototype by using the Active user link, by tapping the big central [ ] scan QR button in the footer, right from the wallet overview)

pwltr:
[...]

PR #2346, opened 2024-11-11
feat: show warning if user tries to switch apps during LN receive

limpbrains:
### Description

It works better on iOS because it has `inactive` state, where but is not yet fully background.

### Linked Issues/Tasks

#2319

### Type of change

New feature

### Tests

No test

### Screenshot / Video

https://github.com/user-attachments/assets/a305f632-60b5-4df7-b93b-030bb7c62acb

### QA Notes

- open receive with LN invoice shown
- try to put app into background

### DIFF
--- a/src/screens/Wallets/Receive/ReceiveQR.tsx
+++ b/src/screens/Wallets/Receive/ReceiveQR.tsx
@@ -41,6 +41,7 @@ import { createLightningInvoice } from '../../../store/utils/lightning';
 import { updatePendingInvoice } from '../../../store/slices/metadata';
 import { generateNewReceiveAddress } from '../../../store/actions/wallet';
 import {
+	appStateSelector,
 	isLDKReadySelector,
 	viewControllerIsOpenSelector,
 } from '../../../store/reselect/ui';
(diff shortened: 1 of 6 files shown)


src/screens/Wallets/Receive/ReceiveQR.tsx
```diff
@@ -244,6 +246,16 @@ const ReceiveQR = ({
 		dispatch,
 	]);
 
+	useEffect(() => {
```
pwltr, line 249: The toast also shows in the send flow:

https://github.com/user-attachments/assets/390c501d-17ab-47c3-85fd-911b7f6c48cb

Maybe this needs to be `useFocusEffect` ?

pwltr (changes requested):

src/screens/Wallets/Receive/ReceiveQR.tsx
limpbrains: fixed by checking `receiveNavigationIsOpen`

pwltr (approved):

catch-21:
Tested on [release-v1.0.7](https://github.com/synonymdev/bitkit/tree/release-v1.0.7) [3b9a833]
Works well on iPhone.
Using android, it shows the message once you return to the app. This is still useful even if it is noticed after an attempted spend.


https://github.com/user-attachments/assets/74e342e4-aebc-4a61-b091-45f1e315ed86

PR #2347, opened 2024-11-11
fix: compress all illustrations

limpbrains:
### Description

using https://tinypng.com/

### Type of change

Refactoring

### Tests

No test


pwltr:
Need to be careful most of these are already compressed [here](https://github.com/synonymdev/bitkit/pull/667), at some point you are going to have visible quality loss, for example in the `card.png` you can see banding. I think compressing once is enough.

Issue #1856, opened 2024-05-22
[Bug]: Use new LSP policy limits


pwltr:
### Describe the bug

When trying to open an imbalanced channel where the LSP balance is significantly low LDK throws an error when the LSP tries to open the channel.

Low priority because the UI currently doesn't allow for such channels but would be good to catch that at the channel order purchase request.

### Reproduce

1. Order a channel with Blocktank with `clientBalance: 100000, lspBalance: 10000`
2. Channel won't open, see LDK error

### Operating system

iOS

### Bitkit version

510f311d

### Log output

```shell
2024-05-22 13:20:37 INFO (JS): channel_manager_open_channel_request: {"counterparty_node_id":"03b9a456fb45d5ac98c02040d39aec77fa3eeb41fd22cf40b862b393bcfc43473a","requires_zero_conf":true,"supports_zero_conf":true,"requires_anchors_zero_fee_htlc_tx":true,"push_sat":100000,"temp_channel_id":"689dceda78d082ff02efa73870610ee669f2a489d22ace48a3c0952a839f2ddc","funding_satoshis":110000}
2024-05-22 13:20:37 INFO (JS): Accepting zero conf channel from peer 03b9a456fb45d5ac98c02040d39aec77fa3eeb41fd22cf40b862b393bcfc43473a
2024-05-22 13:20:37 ERROR (LDK): Funding amount (9340 sats) can't even pay fee for initial commitment transaction fee of 13590 sats. (lightning::ln::channelmanager 6036)
2024-05-22 13:20:37 DEBUG (SWIFT): Adding channel manager backup to queue
2024-05-22 13:20:37 DEBUG (SWIFT): Persisted channel manager to disk
2024-05-22 13:20:37 DEBUG (SWIFT): Error: channel_accept_fail. Message: 'Funding amount (9340 sats) can't even pay fee for initial commitment transaction fee of 13590 sats.'
2024-05-22 13:20:37 ERROR (JS): acceptChannel() failed: Error: Funding amount (9340 sats) can't even pay fee for initial commitment transaction fee of 13590 sats.
2024-05-22 13:20:37 ERROR (JS): Accept channel (trusted zero conf) error: "Funding amount (9340 sats) can't even pay fee for initial commitment transaction fee of 13590 sats."
```

catch-21:
This might be related: https://github.com/synonymdev/bitkit/issues/2379

catch-21:
Verified on v1.0.8
[...]

PR #2350, opened 2024-11-14
fix: migrate to Blocktank client v2

limpbrains:
### Description

Migrate to the new blocktank http client v2

### Type of change

Refactoring

### Tests

No test

### DIFF
--- a/src/screens/Wallets/Receive/ReceiveConnect.tsx
+++ b/src/screens/Wallets/Receive/ReceiveConnect.tsx
@@ -41,7 +41,7 @@ const ReceiveConnect = ({
 	const { maxChannelSizeSat } = blocktank.options;
 	const minChannelSize = Math.round(amount * 2.5);
 	const maxChannelSize = Math.round(maxChannelSizeSat / 2);
-	const channelSize = Math.max(minChannelSize, maxChannelSize);
+	const channelSize = Math.min(minChannelSize, maxChannelSize);
 	const lspBalance = channelSize - amount;
 	const payAmount = amount - feeEstimate;
 	const displayFee = useDisplayValues(feeEstimate);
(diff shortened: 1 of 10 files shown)


src/screens/Wallets/Receive/ReceiveConnect.tsx
```diff
@@ -41,7 +41,7 @@ const ReceiveConnect = ({
 	const { maxChannelSizeSat } = blocktank.options;
 	const minChannelSize = Math.round(amount * 2.5);
 	const maxChannelSize = Math.round(maxChannelSizeSat / 2);
-	const channelSize = Math.max(minChannelSize, maxChannelSize);
+	const channelSize = Math.min(minChannelSize, maxChannelSize);
```
limpbrains, line 44: It is a mistake here. `minChannelSize = amount * 2.5` and it can be bigger than `maxChannelSize`
Looks like it was changed at https://github.com/synonymdev/bitkit/commit/2a6b1fbd5c97eaaae93867e5093ecebf7b90732d by mistake

limpbrains: Maybe it should be ?
```js
const channelSize = min(max(amount * 2.5, maxChannelSizeSat / 2), maxChannelSizeSat);
```

pwltr: yeah looks wrong, but I think in that case BT will return an error. I'd say leave it for now. I have a task to update these values to the new policy [here](https://github.com/synonymdev/bitkit/issues/1856)

limpbrains: I think my fix is still better than what we have right now in master. I think we can merge this. New policy might not be ready before next release

pwltr (approved):

catch-21:
Sanity check looks good. Will be testing more through regular use and testing sessions. Moving to Done.

src/screens/Wallets/Receive/ReceiveConnect.tsx
pwltr: This is causing a bug with minimum channel size. I think we need to revert this.

PR #2356, opened 2024-11-21
fix: do not start LDK unless we have fresh fee and block data

### LINKED ISSUE #2352: [Bug]: Using cached/stale fee rate and block header data causes lightning issues
### Describe the bug

To startup LDK we need latest fee rates and block headers.

Beignet can fail and display no error, just return last cached data: https://github.com/synonymdev/beignet/blob/84911083a01013aa769ba91f89ff326f1f5576e8/src/wallet/index.ts#L479C7-L482C49

This can cause LDK to assume it has the latest data when it does not so payments fail and possibly channels force close.

We should not attempt to start LDK with stale data, rather handle errors/retries in Beignet or show network connection issue notification in Bitkit.

limpbrains:
### Description

- as discussed with @Jasonvdb we are now always trying to get fresh fees for LDK. Never use cached value. We try mempool and blocktank. Both sources two times before we fail. Also with 10 seconds timeout. If function fails, it throws an error, LDK needs to process it.
- same for blockdata. We check that electrum is connected and block data is not the default one.
- we do not start LDK unless we make sure both getFees and getBestBlock works
- getFees is covered with unit tests
- simplified onchain fee fetching logic for Bitkit, as it is covered by beignet
- show error if LDK start / refresh fails
- react-native-ldk probably needs to be updated to handle possible exceptions @Jasonvdb 

### Linked Issues/Tasks

closes #2352

### Type of change

Refactoring

### Tests

Unit test

### QA Notes

Everything should work as before. Just LDK should not start if network is down and should show an error

### DIFF
--- a/src/utils/lightning/index.ts
+++ b/src/utils/lightning/index.ts
@@ -253,6 +253,7 @@ export const getFees: TGetFees = async () => {
 		});
 	};
 
+	// fetch, validate and map fees from mempool.space to IOnchainFees
 	const fetchMp = async (): Promise<IOnchainFees> => {
 		const f1 = await fetch('https://mempool.space/api/v1/fees/recommended');
 		const j: IGetFeeEstimatesResponse = await f1.json();
@@ -276,6 +277,7 @@ export const getFees: TGetFees = async () => {
 		};
 	};
 
+	// fetch, validate and map fees from Blocktank to IOnchainFees
 	const fetchBt = async (): Promise<IOnchainFees> => {
 		const f2 = await fetch('https://api1.blocktank.to/api/info');
 		const j: IBtInfo = await f2.json();
--- a/src/utils/lightning/index.ts
+++ b/src/utils/lightning/index.ts
@@ -4,7 +4,13 @@ import * as bitcoin from 'bitcoinjs-lib';
 import ecc from '@bitcoinerlab/secp256k1';
 import RNFS from 'react-native-fs';
 import { err, ok, Result } from '@synonymdev/result';
-import { EPaymentType, TGetAddressHistory } from 'beignet';
+import {
+	EPaymentType,
+	IBtInfo,
+	IGetFeeEstimatesResponse,
+	IOnchainFees,
+	TGetAddressHistory,
+} from 'beignet';
 import lm, {
 	ldk,
 	defaultUserConfig,
@@ -25,11 +31,11 @@ import lm, {
 	TCloseChannelReq,
 	TCreatePaymentReq,
 	TGetFees,
-	THeader,
 	TInvoice,
 	TPaymentReq,
 	TTransactionData,
 	TTransactionPosition,
+	TGetBestBlock,
 } from '@synonymdev/react-native-ldk';
 
 import {
@@ -42,7 +48,6 @@ import {
 	getBip39Passphrase,
 	getCurrentAddressIndex,
 	getMnemonicPhrase,
-	getOnChainWalletDataAsync,
 	getOnChainWalletElectrumAsync,
 	getSelectedNetwork,
 	getSelectedWallet,
@@ -55,9 +60,7 @@ import {
 	getFeesStore,
 	getLightningStore,
 	getStore,
-	getWalletStore,
 } from '../../store/helpers';
-import { defaultHeader } from '../../store/shapes/wallet';
 import { updateBackupState } from '../../store/slices/lightning';
 import {
 	moveMetaIncPaymentTags,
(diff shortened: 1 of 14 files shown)


src/utils/wallet/transactions.ts
```diff
@@ -904,11 +905,11 @@ export const getFeeEstimates = async (
...
 		const wallet = await getOnChainWalletAsync();
-		const feeRes = await wallet.getFeeEstimates();
-		if (!feeRes) {
-			return err('Unable to get fee estimates.');
+		const feeRes = await wallet.updateFeeEstimates(forceUpdate);
+		if (feeRes.isErr()) {
+			return err(feeRes.error);
 		}
-		return ok(feeRes);
+		return ok(feeRes.value);
```
pwltr: Can simply return the result with `return wallet.updateFeeEstimates(forceUpdate);`

src/utils/lightning/index.ts
```diff
@@ -234,8 +241,98 @@ const getScriptPubKeyHistory = async (
...
 };
 
-const getFees: TGetFees = async () => {
-	const fees = getFeesStore().onchain;
+/**
+ * Fetch fees from mempool.space and blocktank.to, prioritizing mempool.space.
+ * Multiple attempts are made to fetch the fees from each provider
+ * Timeout after 10 seconds
+ */
+export const getFees: TGetFees = async () => {
```
pwltr, line 249: This could be more readable, asked AI:

```ts
/**
 * Fetch fees from mempool.space and blocktank.to, prioritizing mempool.space.
 * Multiple attempts are made to fetch the fees from each provider
 * Timeout after 10 seconds
 */
export const getFees: TGetFees = async () => {
	const throwTimeout = (t: number): Promise<never> => {
		return new Promise((_, rej) => {
			setTimeout(() => rej(new Error('timeout')), t);
		});
	};

	const fetchMp = async (): Promise<IOnchainFees> => {
		const response = await fetch('https://mempool.space/api/v1/fees/recommended');
		const data: IGetFeeEstimatesResponse = await response.json();
		if (!response.ok || !isValidMempoolData(data)) {
			throw new Error('Failed to fetch mempool.space fees');
		}
		return mapMempoolDataToFees(data);
	};

	const fetchBt = async (): Promise<IOnchainFees> => {
		const response = await fetch('https://api1.blocktank.to/api/info');
		const data: IBtInfo = await response.json();
		if (!response.ok || !isValidBlocktankData(data)) {
			throw new Error('Failed to fetch blocktank fees');
		}
		return mapBlocktankDataToFees(data);
	};

	let fees: IOnchainFees;
	if (getFeesStore().override) {
		fees = getFeesStore().onchain;
	} else if (getSelectedNetwork() !== 'bitcoin') {
		fees = initialFeesState.onchain;
	} else {
		fees = await fetchFeesWithPriority();
		updateOnchainFeeEstimates({ feeEstimates: fees });
	}

	return mapFeesToOutput(fees);
};

// Helper functions
const isValidMempoolData = (data: IGetFeeEstimatesResponse): boolean => {
	return (
		data.fastestFee > 0 &&
		data.halfHourFee > 0 &&
		data.hourFee > 0 &&
		data.minimumFee > 0
	);
};

const mapMempoolDataToFees = (data: IGetFeeEstimatesResponse): IOnchainFees => {
	return {
		fast: data.fastestFee,
		normal: data.halfHourFee,
		slow: data.hourFee,
		minimum: data.minimumFee,
		timestamp: Date.now(),
	};
};

const isValidBlocktankData = (data: IBtInfo): boolean => {
	return (
		data?.onchain?.feeRates?.fast > 0 &&
		data?.onchain?.feeRates?.mid > 0 &&
		data?.onchain?.feeRates?.slow > 0
	);
};

const mapBlocktankDataToFees = (data: IBtInfo): IOnchainFees => {
	const { fast, mid, slow } = data.onchain.feeRates;
	return {
		fast,
		normal: mid,
		slow,
		minimum: slow,
		timestamp: Date.now(),
	};
};

const fetchFeesWithPriority = async (): Promise<IOnchainFees> => {
	return new Promise<IOnchainFees>((resolve, reject) => {
		const mpPromise = Promise.race([
			fetchMp().catch(fetchMp),
			throwTimeout(10000),
		]);
		const btPromise = Promise.race([
			fetchBt().catch(fetchBt),
			throwTimeout(10000),
		]).catch(() => null); // Prevent unhandled rejection

		mpPromise.then(resolve).catch(() => {
			btPromise
				.then((btFees) => {
					if (btFees !== null) {
						resolve(btFees);
					} else {
						reject(new Error('Failed to fetch fees'));
					}
				})
				.catch(reject);
		});
	});
};

const mapFeesToOutput = (fees: IOnchainFees) => {
	return {
		onChainSweep: fees.fast,
		maxAllowedNonAnchorChannelRemoteFee: Math.max(25, fees.fast * 10),
		minAllowedAnchorChannelRemoteFee: fees.minimum,
		minAllowedNonAnchorChannelRemoteFee: Math.max(fees.minimum - 1, 0),
		anchorChannelFee: fees.slow,
		nonAnchorChannelFee: fees.normal,
		channelCloseMinimum: fees.minimum,
	};
};
```

limpbrains: Problem is that `src/utils/lightning/index.ts` itself is about 1700 lines, extracting some logic from getFees to a separate functions, suggests that they are will be reused somewhere else, but they are not.
Using `isValidMempoolData` and `mapMempoolDataToFees` really is an overhead, it's like just a couple of calls

pwltr: it's mostly to name functionality and avoid indentation, but ok with me as is.

pwltr (approved):

PR #2358, opened 2024-11-22
chore: upgrade detox

pwltr:
### Description

Upgrade detox dependency to fix e2e tests for Xcode 16 / iOS 18


limpbrains:
Android doesn't work, unfortunately

pwltr:
I think we're affected by a bug that was fixed lately on Android side that gives false positives for some tests: https://github.com/wix/Detox/pull/4519

I couldn't get the tests to run locally on Android, can revisit later. Bumped detox to `20.23.1` for now to only include the fix for iOS 18.

limpbrains (approved):

PR #2359, opened 2024-11-22
fix: always use iphone 15 for detox iOS tests

limpbrains:
### Description

- always use iPhone 15 for e2e tests

### Type of change

Bug fix

### Tests

Detox test


limpbrains:
Strange that tests are working fine again.
I'm changing this PR to just always use iPhone 15 for e2e tests

pwltr (approved):

PR #2363, opened 2024-11-26
fix: sync language files

limpbrains:
### Description

Sync language files

### Type of change

Refactoring

### Tests

No test

### DIFF
--- a/src/utils/i18n/locales/ca/cards.json
+++ b/src/utils/i18n/locales/ca/cards.json
@@ -13,9 +13,6 @@
     }
   },
   "quickpay": {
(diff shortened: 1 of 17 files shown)


pwltr (approved):
I don't know who does these but seeing how fast translations are being submitted for new strings makes me assume these are generated/auto-translated. Can only validate German translation but some of these are pretty bad. IMO it's better to leave text untranslated (English) than be incomprehensible. I'm happy to review for German before we open these PRs in the future.

limpbrains:
I've excluded German

PR #2385, opened 2024-12-04
fix: stop beignet wallet before replace/wipe it

limpbrains:
### Description

This should prevent wallet instance to write any data to the redux or call any onMessage callback after it has been stoped.
Maybe we can stop restarting RN after this

### Type of change

Bug fix

### Tests

No test

### DIFF
--- a/src/store/utils/settings.ts
+++ b/src/store/utils/settings.ts
@@ -9,7 +9,7 @@ import { removePin } from '../../utils/settings';
 import { wipeKeychain } from '../../utils/keychain';
 import { showToast } from '../../utils/notifications';
 import { wipeLdkStorage } from '../../utils/lightning';
-import { getSelectedWallet } from '../../utils/wallet';
+import { getOnChainWallet, getSelectedWallet } from '../../utils/wallet';
 import { TWalletName } from '../types/wallet';
 import { __E2E__ } from '../../constants/env';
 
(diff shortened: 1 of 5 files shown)


src/store/utils/settings.ts
```diff
@@ -31,6 +31,12 @@ export const wipeApp = async ({
 	restartApp?: boolean;
 } = {}): Promise<Result<string>> => {
 	try {
+		// stop onchain wallet if it exists
+		try {
+			const wallet = getOnChainWallet();
+			await wallet.stop();
+		} catch (e) {}
+
```
pwltr, line 39: We can expect the wallet to exist at this point, so it should be able to be stopped. If it does fail we should probably propagate and return the error. Especially if we're going to remove the restart in the future.

```suggestion
		// stop onchain wallet if it exists
		const wallet = getOnChainWallet();
		await wallet.stop();

```

pwltr (changes requested):

src/store/utils/settings.ts
limpbrains: We run wipeApp during Onboard/TermsOfUse, at this point wallet does not exists

pwltr: Right ok let's leave it for now. If we're going to remove the restart later we should probably return the error outside the onboarding though.

pwltr (approved):

PR #2388, opened 2024-12-06
fix: load intl polyfill locale data

limpbrains:
### Description

Because we did not load the polyfill language files, some translations were missing

### Type of change

Bug fix

### Tests

No test

### Screenshot / Video

German before and after
 

### QA Notes

Make sure other languages that we support are showing correct dates on android


src/utils/i18n/index.ts
```diff
@@ -58,7 +62,27 @@ i18nICU
...
+	.then(async () => {
+		// we need to load language related polyfill data
+		const lang = i18nICU.language;
+		try {
+			// @ts-ignore
+			if (NumberFormat.polyfilled) {
+				await numberFormatPolyfills[lang]?.();
+			}
+			// @ts-ignore
+			if (Intl.PluralRules.polyfilled) {
```
pwltr, line 76: Did you mean to set this somewhere?

src/utils/i18n/index.ts
```diff
@@ -58,7 +62,27 @@ i18nICU
...
+				await pluralRulesPolyfills[lang]?.();
+			}
+			// @ts-ignore
+			if (Intl.RelativeTimeFormat.polyfilled) {
```
pwltr, line 80: Same here?

pwltr (changes requested):

src/utils/i18n/index.ts
limpbrains: This is set by polyfill library itself

pwltr: I see

pwltr (approved):

PR #2406, opened 2024-12-17
fix(transfer): remove 80% onchain cap, misc fixes

### LINKED ISSUE #2249: [Feature]: Improve error when unable to top up spending balance due to max liquidity limit
### Describe the problem

When a user wants to transfer from savings to spending but are already very near to or have even exceeded the €950 limit they will get a message like:

Screenshot 2024-09-26 at 19 19 39

It isn't clear what they can do to resolve the issue.

### Describe the solution

We could suggest that they close one of their LN channels, perhaps the one with most inbound liquidity.

### Additional context

_No response_

### LINKED ISSUE #2393: [Feature]: Add maximum bound validation to CJIT flow
### Describe the problem

Since increasing the maximum amount that can added to the spending balance via CJIT to 950euro (as part of the new Liquidity Management Policy for Lightning Channels) it is possible to exceed the boundary so that blocktank rejects it, showing an unpleasing toast error and a loading spinner that displays indefinitely.

https://github.com/user-attachments/assets/07f5ef4b-a9ad-4ffe-b309-8515982c1529

### Describe the solution

We should add some validation to prevent the user from proceeding when the maximum is exceeded.
Or at least show a nice error and allow the user to retry instead of showing the spinner for eternity.

### Additional context

_No response_

pwltr:
### Description

- remove 80% onchain cap for channel orders, now uses fee estimation from LSP and adds that to the spendable onchain balance
- add CJIT max amount client validation

### Linked Issues/Tasks

Closes https://github.com/synonymdev/bitkit/issues/2393 https://github.com/synonymdev/bitkit/issues/2249

### Type of change

### Tests

- [x] Detox test
- [ ] Unit test
- [ ] No test

### Screenshot / Video

https://github.com/user-attachments/assets/3acea7d9-11cf-493d-a720-2c07f0d9ef4e

### QA Notes

Because the LSP service fee estimation when calculating the max amount is slightly lower than the actual there will be a dust output leftover which is added to the onchain fee.

### DIFF
--- a/src/store/utils/blocktank.ts
+++ b/src/store/utils/blocktank.ts
@@ -14,19 +14,17 @@ import { __E2E__ } from '../../constants/env';
 import { addTransfer, removeTransfer } from '../slices/wallet';
 import { updateBeignetSendTransaction } from '../actions/wallet';
 import { setLightningSetupStep } from '../slices/user';
-import { getBlocktankStore, dispatch, getFeesStore } from '../helpers';
+import { getBlocktankStore, dispatch } from '../helpers';
 import * as blocktank from '../../utils/blocktank';
 import {
 	createOrder,
 	getBlocktankInfo,
 	getCJitEntry,
-	getOrder,
 	isGeoBlocked,
 	openChannel,
 	watchOrder,
 } from '../../utils/blocktank';
 import {
-	getBalance,
 	getSelectedNetwork,
 	getSelectedWallet,
 	refreshWallet,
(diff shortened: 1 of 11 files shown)


src/store/utils/blocktank.ts
```diff
@@ -211,100 +207,38 @@ export const refreshBlocktankInfo = async (): Promise<Result<string>> => {
...
-		fee += 5;
-	}
-
-	const feeRes = updateFee({ satsPerByte: fee });
-	if (feeRes.isErr()) {
-		return err(feeRes.error.message);
-	}
-
-	return ok(buyChannelData);
+	return ok(order);
```
pwltr, line 241: The transaction is already setup correctly and the amount validated in the beginning of the transfer flow so this can be simplified to only updating the recipient address and amount.

coreyphillips (approved):

PR #2492, opened 2025-02-25
fix(ui): fix home screen rerender causing navigation lag

pwltr:
### Description

Fixes a rerender on the home screen caused by smelly `onFocus` code. This moves the timed/conditional bottom sheets into the home screen and extends the check in the `<TabBar />` so it hides properly. This significantly improves navigation responsiveness especially on Android. Also renames `Wallets/index.ts` to `Home.tsx`.

### Type of change

### Tests

- [ ] Detox test
- [ ] Unit test
- [x] No test

### DIFF
--- a/e2e/backup.e2e.js
+++ b/e2e/backup.e2e.js
@@ -91,7 +91,7 @@ d('Backup', () => {
 		await element(by.id('NavigationClose')).atIndex(0).tap();
 
 		// remove 2 default widgets, leave PriceWidget
-		await element(by.id('WalletsScrollView')).scroll(200, 'down', 0);
+		await element(by.id('HomeScrollView')).scroll(200, 'down', 0);
 		await element(by.id('WidgetsEdit')).tap();
 		for (const w of ['NewsWidget', 'BlocksWidget']) {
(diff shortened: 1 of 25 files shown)


coreyphillips (changes requested):
Tapping "Receive" and quickly closing the modal results in the TabBar disappearing.


https://github.com/user-attachments/assets/61362f17-4508-4ee0-a9ef-f7a5c3e6443b

pwltr:
> Tapping "Receive" and quickly closing the modal results in the TabBar disappearing.
> 
>  Simulator.Screen.Recording.-.iPhone.16.-.2025-02-27.at.18.49.45.mp4

This is an issue with the bottom-sheet and not caused by the changes here, although it does make it more likely to occur. I have a refactor coming for the sheets that will make it more reliable and will also solve this issue by not having to hide the TabBar at all. There are several long-standing issues with the bottom-sheet so it's a bit difficult to not break something by fixing another.

I updated the code to change when it hides the TabBar to at least fix this for send/receive sheets but it will still occur when you do the same with any one of the sheets in `shouldHide`. Intention is to get these fixes in for the upcoming release.

coreyphillips (approved):

PR #2568, opened 2025-08-11
chore: sync translations

jvsena42:
### Description

This PR sync transifex translations

### Linked Issues/Tasks
`n/a`

### Type of change

- [X] Refactoring (improving code without creating new functionality)

### Tests

- [ ] Detox test
- [ ] Unit test
- [X] No test

### Screenshot / Video
`n/a`

### QA Notes

`n/a`


src/utils/i18n/locales/en/other.json
```diff
@@ -307,17 +307,6 @@
...
-		"used_up": {
-			"title": {
-				"string": "Out of Gifts"
-			},
-			"text": {
-				"string": "Sorry, you're too late! All gifts for this code have already been claimed."
-			},
-			"button": {
-				"string": "OK"
-			}
```
pwltr: This needs to be kept, in general when syncing from Transifex english translation files should not be touched by the script

pwltr (changes requested):

src/utils/i18n/locales/en/other.json
jvsena42: Should I restore it manually?

It was autogenerated after run  `tx pull --mode translator -f -a` in the repository root and run `node ./clean_empty_strings.js` in bitkit-loc root

pwltr: As far as I know if you don't pass the `--source/-s` flag it should not pull the english files. But yeah just discard this change, it looks good otherwise

pwltr (approved):

ovitrif (approved):
LGMT, except

Not sure why we have some new strings added (e.g. in German) that are just the English translation, but I checked and there were old strings the same way, so it should be fine 👍🏻

